PR #10, opened 2020-09-02
version info (probably broken)

pgierz:
A small utility to print the version number into the main output

@koldunovn: My fortran is (very) rusty, but maybe this is already a good start..


hegish:
Hi guys, I think the build info should be compile time generated. Otherwise you could update the repo and have a newer commit sha being displayed without a rebuild. Also, some systems do not have a git installed on the compute nodes and this approach would fail. I can provide a compile time solution next week, if you like.

pgierz:
Closed in favor of #13

PR #23, opened 2020-10-23
Workbench fixpartcell

patrickscholz:
- change range limiter in cvmix_tke.F90 for KappaM_max
- fix small issue in oce_ale_pressure_bv.F90
- change error output


koldunovn:
@patrickscholz We identified the problem with `w_split`, but do you think it still makes sence to merge this one? Will not hurt, I guess?

patrickscholz:
let me recommit some actualisation, i added some comments for myself and we i will also add the new default parameter for w_cfl_max

PR #27, opened 2020-10-29
Workbench fixpartcell

patrickscholz:
Things to commit as result from the debugging of the AO4 setup issues


koldunovn:
@dsidoren Dima, I think it's better if you can have a look :)

PR #31, opened 2020-11-05
add mesh checksum to output files

hegish:


koldunovn:
@hegish Do you plan to add `fesom` to each of the global attributes within this PR, or you plan it for the different one and we can merge this?

hegish:
I thought I would add the prefix from a separate PR, as the topic of this Pr is about the checksum. So you can merge this one first.

PR #33, opened 2020-11-13
prefix the global NetCDF attributes so they do not conflict with attr…

hegish:
…ibutes inserted from other tools (e.g. during CMORization)


koldunovn:
Works fine on ollie and mistral.

PR #34, opened 2020-11-13
Add parameters to the namelist.ice

koldunovn:
Put additional parameters to the `namelis.ice` to make it possible to use ice tuning from @lzampier and change some EVP related parameters form the namelist.


koldunovn:
@dsidoren or @patrickscholz can you, please have a look and merge if OK?

patrickscholz (approved):

PR #40, opened 2020-11-24
enable omitted unit test

hegish:
we had a unit test which was never used during testing...


koldunovn:
Sorry it took so long to merge :(

PR #53, opened 2021-01-05
Workbench partcell

patrickscholz:
@dsidoren & @koldunovn 
**This branch should be merged after the Cavity stuff is merged!!!**

- This merge should improve the capabilities of partial cells, i took Qiangs advice into account and try to avoid extrapolation for the computation of the pressure gradient force. Therefor I redefine the vertices partial cell bottom depth, which corresponds now to the deepest partial bottom depth of the surrounding elements. This way there is never a vertices partial bottom depth that is shallower than a connected elemental partial bottom depth --> no extrapolation anymore. This seemed to have solved time step issue of partial cells, now with or without partial cells for CORE2 mesh the time step remained the same. 

- Further i introduce partial bottom layer thickness threshold where partial cells should not be applied (see namelist.config: partial_cell_tresh=0.0). The same is also possible in case of cavity surface partial cells (cavity_partial_cell_tresh=0.0) . Not sure if this has an advantage yet. For CORE2 mesh it didn't make a difference, but maybe when going to higher  horizontal and vertical resolutions,  still need to test this out. This could help to avoid to make already thin layers in shelf areas even thinner, due to partial cells. I hope it could help to improve time step and CFL_z issues. 

- Further introduce another way to compute PGF (easypgf), simply 2nd order  interpolate vertices temperature and salinity  values to elemental depth and than compute densities on element depth using state equation and  than compute pressure gradient on element. I tested easypgf together with shepetkin method, both produce very similiar results on real ocean (CORE2 mesh+forcing). However on simplified testcase, of constant temp and salinity ocean, without boundary condition using partial bottom cells, easypgf was the only method which reduced there the bottom pressure gradient force to zero (machine precision), like it should be!


src/gen_modules_config.F90
```diff
@@ -9,86 +9,116 @@ module g_config
...
+  logical                :: use_partial_cell=.false.  
+  
+  ! if a thicker partial bottom layer thickness is more realistic than always 
+  ! apply it, BUT when a thinner partial bottom layer thickness is more realistic than 
+  ! only apply it when the initial full cell bottom layer thickness is above the 
+  ! treshhold partial_cell_tresh to not allow already thin layers to become even 
+  ! thinner. e.g 
+  ! partial_cell_tresh=10 --> thinner partial bottom cells will be only applied for initial 
+  ! bottom layer thicknesses larger than 10m 
+  real(kind=WP)          :: partial_cell_tresh=0.0_WP
```
koldunovn, line 56: Maybe rename to `partial_cell_thresh` :)

src/gen_modules_config.F90
```diff
@@ -9,86 +9,116 @@ module g_config
...
+  
   ! for zlevel: in case min_hnode criteria is reached over how many level should 
   ! ssh change be distributed
   integer                :: lzstar_lev=4
+  
   ! maximal pressure from ice felt by the ocean
   real(kind=WP)          :: max_ice_loading=5.0
 
-  namelist /ale_def/ which_ALE, use_partial_cell, min_hnode, lzstar_lev, max_ice_loading
+  namelist /ale_def/ which_ALE, use_partial_cell, partial_cell_tresh, min_hnode, lzstar_lev, max_ice_loading
```
koldunovn, line 69: Same here, rename to `partial_cell_thresh` :)

koldunovn:
@patrickscholz Sounds great! :) I will now try it with CORE2, viscosity 7, that with the master crashes in the 1990s and with Claudia's 1km Fram Strait.

koldunovn:
@patrickscholz The Fram test case seem to work now (I was able to compute 15 days instead of 2). Now on to the AO1 case, fingers crossed :)

src/gen_modules_config.F90
patrickscholz: How about partial_cell_trash ?! ;-D

koldunovn: That's how I read it initially :))

koldunovn:
The AO1 run did 15 days without problems. I will test it with a bit more extreme wsplit and viscosity settings, but I think you solved the problem, Patrick :) 🥇 Thanks a lot!

PR #57, opened 2021-01-18
integrate master with openacc

dguibert:
This PR integrates the merge of master into pgi-fixes and adds some commits to get FESOM compile with the PGI compiler.

I propose this to the `pgi-fixes` branch but it could be merged into `master` as without any support of OpenACC directives it behaves as usual.

PR #58, opened 2021-01-18
CMake: modernize and avoids some pitfulls

dguibert:
The last 8 commits are really belonging to this PR (the others are already in #57).

This PR aims to modernize the CMakeLists and its modules to avoid some errors due to some assumptions not necessarely met.

This has been tested on 2 of our clusters with PGI or Intel compilers and required librairies.


koldunovn:
@hegish Jan, can you, please have a look at it?

hegish:
Yes, I will have a look. Give me some time...

koldunovn:
@dguibert Thanks a lot for the PR! Can you, please, update to the latest https://github.com/FESOM/fesom2/blob/master/.github/workflows/fesom2_urot.yml so that (hopefully) the tests are running on the PR?

dguibert:
> @dguibert Thanks a lot for the PR! Can you, please, update to the latest https://github.com/FESOM/fesom2/blob/master/.github/workflows/fesom2_urot.yml so that (hopefully) the tests are running on the PR?

done.

koldunovn:
> done.

Great! Tests are passing, already a good sign :))

hegish:
Hi all, sorry it took so long, I am not sure how to proceed here as it is a bit difficult to test all the cmake related changes, as there are also changes to e.g. Fortran files.
- on f750c13, we mimic the default cmake behaviour here, which does not present a box. Has it changed in cmake?
- 2669fe6 might be a good catch
- Has it been checked if the newer cmake version required by 42063e9 is available on all systems we use?
- 236e077: globbing has been a design decision as it e.g. allows non-cmake savvy people to easily add source files. and we know if we have stray source files in the directory which are not compiled (which has been given us headaches in the past).
- 41b4ec9 is a different approach, but in which way do we benefit from it? Will it not break with older cmake versions?
- We can not use eae8ed6, as recompiling the info each time is the exact purpose of this file.
- I would opt against 7cbabc1, as we leave all the MPI stuff to the compiler wrappers on purpose.
- I do not quite understand why eb9246e is required.

dguibert:
>     * We can not use [eae8ed6](https://github.com/FESOM/fesom2/commit/eae8ed66b6ba481f5762e72f37520e95b9864a0b), as recompiling the info each time is the exact purpose of this file.

The patch does not change this behaviour as it generates the file each time a new hash is given. Without this patch the file is generated each time we call `make` recompiling each time `fesom`.

>     * I would opt against [7cbabc1](https://github.com/FESOM/fesom2/commit/7cbabc15c548102e290fd195b9d9c37d3edd6e0d), as we leave all the MPI stuff to the compiler wrappers on purpose.

Can you elaborate on that?

hegish:
> > ```
> > * We can not use [eae8ed6](https://github.com/FESOM/fesom2/commit/eae8ed66b6ba481f5762e72f37520e95b9864a0b), as recompiling the info each time is the exact purpose of this file.
> > ```
> 
> 
> The patch does not change this behaviour as it generates the file each time a new hash is given. Without this patch the file is generated each time we call `make` recompiling each time `fesom`.

Hey @dguibert, sorry I was not clear about this. Recompiling each time you issue make is the sole purpose of this info file. It should reflect the git status at the time of building Fesom.
> 
> 
> > ```
> > * I would opt against [7cbabc1](https://github.com/FESOM/fesom2/commit/7cbabc15c548102e290fd195b9d9c37d3edd6e0d), as we leave all the MPI stuff to the compiler wrappers on purpose.
> > ```
> 
> 
> Can you elaborate on that?
To my experience it is very error prone to teach a given compiler all MPI related dependencies like libraries and header files from within cmake. Using the wrappers OTOH always seems to work and is not so fragile. We also need fine grained control over which MPI implementation is being used which is quite error prone to achieve from within cmake. So our approach is to complete rely on the compiler wrappers, which also get full support from the HPC centers and the vendors of the MPI implementation at hand.

dguibert:
> Hey @dguibert, sorry I was not clear about this. Recompiling each time you issue make is the sole purpose of this info file. It should reflect the git status at the time of building Fesom.

The patch generates the `fesom_version_info-generated.F90` only when the `FESOM_GIT_SHA` value changes to reflect the current git status at the time of building Fesom.
With this patch it allows to not recompile Fesom if not necessary. For example calling `make; make` (in a row), without any new commits between the two `make` will compile fesom only once. Currently, with this example, `fesom_version_info-generated.F90` is recompiled because of a new timestamp, thus recompiling fesom.

hegish:
> The patch generates the `fesom_version_info-generated.F90` only when the `FESOM_GIT_SHA` value changes to reflect the current git status at the time of building Fesom.
> With this patch it allows to not recompile Fesom if not necessary. For example calling `make; make` (in a row), without any new commits between the two `make` will compile fesom only once. Currently, with this example, `fesom_version_info-generated.F90` is recompiled because of a new timestamp, thus recompiling fesom.

This sounds great. Can you provide a separate PR for this feature?

dguibert:
> This sounds great. Can you provide a separate PR for this feature?

sure see #60.

src/CMakeLists.txt
```diff
@@ -35,18 +18,148 @@ add_subdirectory(../lib/parms ${PROJECT_BINARY_DIR}/parms)
...
+      -Minfo=accel
+     >
+   >
+  )
+
+target_link_options(fesom PRIVATE
+  $<$<Fortran_COMPILER_ID:PGI>:
+    -acc -ta=tesla:cc35
+  >
+)
```
dguibert, line 162: maybe #515 could integrate this to simplify the option definitions: we see them all at first glance!

dguibert (commented):
#515 is a reimplementation of this patch, it could use the option parametrisation defined here in the CMakeLists (see the comment).

Otherwise this could be closed.

src/CMakeLists.txt
suvarchal: @dguibert apologies again for not being proactive. it  would be great if you can please check if what you suggest is relevant in refactoring and production_DE branches if you find anything can you please make a PR to those branches respectively. I wonder merging to current master which will soon be replaced by refactoring is not worth your effort and time.

JanStreffing:
As mentioned this is reimplemented in https://github.com/FESOM/fesom2/pull/515

PR #60, opened 2021-01-26
cmake: don't recompile at each build

dguibert:
Without this patch, the file `fesom_version_info-generated.F90` is generated
at each build, rebuilding all its dependencies up to the fesom executable.

The patch generates the fesom_version_info-generated.F90 only when the
FESOM_GIT_SHA value changes to reflect the current git status at the
time of building Fesom.
With this patch it allows to not recompile Fesom if not necessary. For
example calling make; make (in a row), without any new commits between
the two make will compile fesom only once. Currently, with this example,
fesom_version_info-generated.F90 is recompiled because of a new
timestamp, thus recompiling fesom.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -16,10 +16,7 @@ set(FESOM_ORIGINAL_VERSION_FILE ${src_home}/fesom_version_info.F90)
 set(FESOM_GENERATED_VERSION_FILE ${CMAKE_CURRENT_BINARY_DIR}/fesom_version_info-generated.F90)
 list(REMOVE_ITEM sources_Fortran ${FESOM_ORIGINAL_VERSION_FILE}) # we want to compile the generated file instead
 list(APPEND sources_Fortran ${FESOM_GENERATED_VERSION_FILE})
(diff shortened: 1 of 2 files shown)


hegish:
not working for me, I get
`fatal: not a git repository (or any of the parent directories): .git`
when running cmake and the SHA is empty

dguibert:
> not working for me, I get
> `fatal: not a git repository (or any of the parent directories): .git`
> when running cmake and the SHA is empty

On the same setup but without this patch, does it fill the FESOM_GIT_SHA variable?
I assume that your source directory has been copied to the cluster without the .git directory.

koldunovn:
@hegish Any news on this one? :)

hegish:
O sorry, I did not notice the PR has changed. Will try again ASAP.

dguibert:
Hi, @hegish I've pushed an update, worth to try.
It is simplified and it works when the build directory is outside the source directory.

hegish:
Ah, did the previous PR not work with out of source build? This maybe was the issue I had with it.

dguibert:
The issue was that the working directory of the git describe command was wrong and we get the error:
````
fatal: not a git repository (or any of the parent directories): .git
````
It's now fixed in the latest update.

dguibert:
Any issue on this one?

suvarchal:
> This has been integrated in #515

Sorry and apologies for being tardy  and not merge changes from this PR before.

PR #64, opened 2021-01-29
Backscatter

StJuricke:
Pull request for Backscatter branch


config/namelist.config
```diff
@@ -50,8 +50,8 @@ include_fleapyear=.true.
 
 &run_config
 use_ice=.true.                  ! ocean+ice
-use_cavity=.false.              !
-use_cavity_partial_cell=.false. 
+!use_cavity=.false.              !
+!use_cavity_partial_cell=.false.
```
koldunovn, line 54: @StJuricke Can you, please, also uncomment those two lines?

src/oce_dyn.F90
```diff
@@ -319,27 +355,32 @@ SUBROUTINE visc_filt_biharm(option, mesh)
...
+     nzmax = minval(nlevels(el))
+     !!PS DO  nz=1,minval(nlevels(el))-1
+     DO  nz=nzmin,nzmax-1
+     u1=(U_c(nz,el(1))-U_c(nz,el(2)))
+     v1=(V_c(nz,el(1))-V_c(nz,el(2)))
+     UV_rhs(1,nz,el(1))=UV_rhs(1,nz,el(1))-u1/elem_area(el(1))
+     UV_rhs(1,nz,el(2))=UV_rhs(1,nz,el(2))+u1/elem_area(el(2))
+     UV_rhs(2,nz,el(1))=UV_rhs(2,nz,el(1))-v1/elem_area(el(1))
+     UV_rhs(2,nz,el(2))=UV_rhs(2,nz,el(2))+v1/elem_area(el(2))
+     UV_total_tend(1,nz,el(1))=UV_total_tend(1,nz,el(1))-u1/elem_area(el(1))
```
dsidoren, line 375: @StJuricke if backscatter is off I would not do the lines with UV_*_tend
why to spend resources then? the same for allocation (not to allocate at all)

StJuricke: @dsidoren I also used the output of UV_*_tend for some of the classical viscosity options to see how the dissipation tendencies look there. It's just a general diagnostic so it could probably be switched on and off by one of the diagnostics switches we already have. Or you can leave it out for now and I just use it when I do intercomparison studies. Not sure if anyone else is generally interested in dissipation tendencies right now.

dsidoren: @StJuricke having it under IF (BACKSTATTER) would be nice otherwise we compute and allocate something which is not used and is costly (3D elemental fields).

StJuricke: @dsidoren  I think BACKSCATTER as a switch would be missleading, as it is more a tendency diagnostic. So we could introduce a switch like momentum_diag or so. How should we proceed here? Should I for now just delete those lines and resubmit to github so that we can merge, and I will add this switch option later? Or can you delete those lines from your side?

koldunovn: @StJuricke It would be best if you could delete it on your side and introduce a diagnostic switch later on. It's good to have the opportunity of turning on additional diagnostic when they are needed. My understanding is @dsidoren is fine with the rest of the PR, so when you delete those variables, I will test and merge. It would be great to have another PR from you with additional diagnostics!

StJuricke: Sorry, it seems like my post below was not published. So I will remove the diagnostics in my side and resubmit, right

Old reply to Dima's comment:
@dsidoren  I think BACKSCATTER as a switch would be missleading, as it is more a tendency diagnostic. So we could introduce a switch like momentum_diag or so. How should we proceed here? Should I for now just delete those lines and resubmit to github so that we can merge, and I will add this switch option later? Or can you delete those lines from your side?

StJuricke: Oh, it was published... weird. Anyway, I will change it on my side

StJuricke: @dsidoren @koldunovn I removed the extra diagnostics for all schemes except the dynamic backscatter. I can add them again at a later point with a specific momentum_diagnostics switch, so that they can be used by other (which I doubt many people, if any, will ever use) but that can wait for later

koldunovn: @StJuricke Thnaks a lot! I will try to run it with the new option and everything is fine merge. Sorry for slow reaction, but things here are just crazy as usual, as you well know :)

koldunovn (approved):

PR #67, opened 2021-02-05
AWI-CM3: enthalpy of fusion for solid runoff

JanStreffing:
This branch keeps track of which fraction of the precipitation runs off the ice sheets as liquid and which part gets turned into land-ice. In the AOGCM we assume that the ice sheet is in balance and all the excess snowfall in the land surface scheme enters the ocean in solid form (rather than building up the glacier). The energy required to melt this ice in the iceberg melting zone is calculated in the runoff-mapper. Here is it applied as an additional heat flux in ice_thermo_cpl.F90

The mass flow inside FESOM2 is unaffected as the solid excess is added to the runoff in the runoff-mapper as before.


src/ice_setup_step.F90
```diff
@@ -132,7 +132,7 @@ subroutine ice_array_setup(mesh)
  allocate(stress_iceoce_x(n_size), stress_iceoce_y(n_size))    
  allocate(U_w(n_size), V_w(n_size))   ! =uf and vf of ocean at surface nodes
 #if defined (__oasis)
-  allocate(oce_heat_flux(n_size), ice_heat_flux(n_size))
+  allocate(oce_heat_flux(n_size), ice_heat_flux(n_size), enthalpyoffuse(n_size))
```
koldunovn, line 135: Does it make sense to only to wrap it in the `if defined (__oifs)` statement to avoid creation of an empty variable when not in use? Similar to how it's done in https://github.com/FESOM/fesom2/pull/70/files

koldunovn (commented):
I think @dsidoren  should decide if he wants this new array to be only defined when IFS is used. Otherwise fine with me :)

koldunovn (approved):

JanStreffing:
I've moved everything behind ifdefs. a bit of code readability traded for a bit of memory for awicm2 ;-)

koldunovn:
> I've moved everything behind ifdefs. a bit of code readability traded for a bit of memory for awicm2 ;-)

Thanks! In your future configurations every 2d field will matter I guess :))

JanStreffing:
Can we merge this until this afternoon? I would like to have Lorenzo start the AWC-CM3 Icepack coupling by branching off of the master branch.
Cheers, Jan

PR #70, opened 2021-02-09
Icepack merge

lzampier:
Addition of the Icepack subroutines to the FESOM2 model. The old sea-ice structure is maintained as default.


koldunovn:
@lzampier Thanks a lot! 
Just first simple comment - you have changed permissions for come of the files in config directory to 755 from 644. Any specific reason to do this?

env.sh
```diff
@@ -43,6 +43,8 @@ elif [[  $LOGINHOST = bsc ]]; then
    STRATEGY="bsc"
 elif [[  $LOGINHOST =~ ^juwels[0-9][0-9].ib.juwels.fzj.de$ ]]; then
    STRATEGY="juwels"
+elif [[  $LOGINHOST =~ ^jwlogin[0-9][0-9].juwels$ ]]; then
+   STRATEGY="juwels"
```
koldunovn, line 47: Better to keep one for `juwels`, the shorter the better :)

lzampier:
> @lzampier Thanks a lot!
> Just first simple comment - you have changed permissions for come of the files in config directory to 755 from 644. Any specific reason to do this?

No, no specific reason for that. You can retain the old permission level.

env.sh
lzampier: I added the second juwels case to allow an easy compilation on the login nodes. But I agree, it is not necessary.

env/juwels/shell
```diff
@@ -4,12 +4,12 @@ module use /gpfs/software/juwels/otherstages
...
-module load Intel/2019.3.199-GCC-8.3.0  IntelMPI/2018.5.288 imkl/2019.3.199
-export FC=mpiifort CC=mpiicc CXX=mpiicpc
+#module load Intel/2019.3.199-GCC-8.3.0  IntelMPI/2018.5.288 imkl/2019.3.199
+#export FC=mpiifort CC=mpiicc CXX=mpiicpc
 
 # For ParaStation MPI 
-#module load  Intel/2019.3.199-GCC-8.3.0 ParaStationMPI/5.4 imkl/2019.5.281
-#export FC=mpifort CC=mpicc CXX=mpicxx
+module load  Intel/2019.3.199-GCC-8.3.0 ParaStationMPI/5.4 imkl/2019.5.281
+export FC=mpifort CC=mpicc CXX=mpicxx
```
koldunovn, line 12: Are we more happy with this settings?

lzampier: Lately yes, for large mesh partitions (approximately >2000 CPUs) I have issues with IntelMPI when the communication structure is initialized by FESOM.

koldunovn:
> > @lzampier Thanks a lot!
> > Just first simple comment - you have changed permissions for come of the files in config directory to 755 from 644. Any specific reason to do this?
> 
> No, no specific reason for that. You can retain the old permission level.

Would be easier if you just put the permissions back in the PR :)

env/juwels/shell
koldunovn: Ok, let's keep your version then as a default.

src/ice_maEVP.F90
```diff
@@ -379,7 +409,7 @@ subroutine EVPdynamics_m(mesh)
      
      if (a_ice(i) >= 0.01_WP) then
         inv_thickness(i) = (rhoice*m_ice(i)+rhosno*m_snow(i))/a_ice(i)
-        inv_thickness(i) = 1.0_WP/max(inv_thickness(i), 9.0_WP)  ! Limit the mass
+        inv_thickness(i) = 1.0_WP/inv_thickness(i) !max(inv_thickness(i), 9.0_WP)  ! Limit the mass
```
koldunovn, line 412: Are you sure we don't want to limit it?

src/ice_maEVP.F90
```diff
@@ -407,9 +437,12 @@ subroutine EVPdynamics_m(mesh)
      msum=sum(m_ice(elnodes))*val3
      if(msum > 0.01) then
         ice_el(el) = .true.
-        asum=sum(a_ice(elnodes))*val3     
-     
+        asum=sum(a_ice(elnodes))*val3          
+!#if defined (__icepack)
+!        pressure_fac(el) = det2*sum(strength(elnodes))*val3
```
koldunovn, line 442: Should be commented out?

src/ice_modules.F90
```diff
@@ -32,7 +32,7 @@ MODULE i_PARAM
   real(kind=WP)             :: theta_io=0.0_WP           ! rotation angle
                                                          ! (ice-ocean), available
 						         ! in EVP
-  real(kind=WP)             :: alpha_evp=250, beta_evp=250
+  real(kind=WP)             :: alpha_evp=500, beta_evp=500
```
koldunovn, line 35: Please return to old default values for consistency. Now one can change this parameters in the `namelist.io`.

PR #72, opened 2021-02-16
Landseamask

rakowsk:
Two helpful changes for developers who prefer Makefiles - the usual cmake workflow is not affected!
1. src/Makefile: up to date w/o C++ routines
2. src/async_threads_module.F90: preprocessor flag DISABLE_MULTITHREADING encapsulated also the calls to C++ code

This helps a lot to test new hardware and new compilers with a focus on compute intensive parts, not the I/O bottleneck.

A new branch just for this change would have been better... please excuse the misleading name, I will open a new branch now for the original purpose to play around with changing sea level - lot's of changes nobody will want to merge...

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -20,7 +20,7 @@ CPP_SOL   = -DPARMS
 
 ###### Objects for Mesh Partitioning ################################################
 # modules
-MOD_INI =  fort_part.o  oce_modules.o  gen_modules_config.o gen_modules_partitioning.o gen_modules_rotate_grid.o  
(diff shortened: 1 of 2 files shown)


koldunovn:
OK with me :) 

@hegish Jan, Can you have a look, since it affects the  `async_threads_module.F90`?

koldunovn (approved):

hegish:
Looks fine to me!
This will introduce possible problems once we try to enable the threads at runtime though. Currently this is not an issue, as we only disable threads dynamically at runtime at the moment.
I will adjust the cmake files to exclude the build of the async stuff altogether once this is merged.

Issue #69, opened 2021-02-09
Latest FESOM2 not working with esm-tools


PengyangSong:
Dear developers,
I am running some cases using esm-tools but it is not working with the latest version.
The log says:
 4:  error in line          549 io_meandata.F90 NetCDF: File exists && NC_NOCLOBBER
 4:  error in line           97 async_threads_module.F90
Maybe the esm-tools is not working with async_threads_module?
which package does this require? or can I disable this approach in FESOM2?
Many thanks.

denizural:
Hi @PengyangSong,
could you please provide the path of your runscript then we can have a look.
Best,
Deniz

koldunovn:
Hi @PengyangSong 

Can you paste here more details on the error here?

PengyangSong:
Thanks @denizural  and  @koldunovn .
I checked again the settings and found that with all the original settings, the model is running good.
But when I set 'ldiag_energy =.true.' (in namelist.io), the error occurs.

The location of logs: /work/ollie/psong/FESOM2/fesom2_NOTIDE5/scripts
The location of config: /home/ollie/psong/esm-tools/esm-master/fesom-2.0/config_notide5
The location of src: /home/ollie/psong/esm-tools/esm-master/fesom-2.0/src

Note that I changed all variables in ldiag_energy to annual mean output (src/io_meandata.F90).

koldunovn:
Maybe the reason is that you try to use `Av` output from the standard output and from energy diagnostics? try to remove `Av` from `namelist.io`. We should avoid to have the same names in `namelist.io`, I will try to take are of it.

PengyangSong:
@koldunovn You are right. However, it is kind of weird because the same setting without using esm-tools will not leading to this error.
Thanks a lot.

denizural:
@PengyangSong
you can also view the namelist that esm-tools uses by giving the `-v `(verbose) option to the `esm_runscripts`. Or of course from your work directory.

PR #91, opened 2021-04-06
do not allow to add the same variable name multiple times to the outp…

hegish:
…ut variables (io streams), e.g. via duplicates in namelist.io

### DIFF
--- a/src/io_meandata.F90
+++ b/src/io_meandata.F90
@@ -948,10 +948,7 @@ subroutine def_stream3D(glsize, lcsize, name, description, units, data, freq, fr
      write(*,*) 'adding I/O stream 3D for ', trim(name)
   end if
 
-  ! add this instance to io_stream array
-  io_NSTREAMS = io_NSTREAMS +1
-  call assert(size(io_stream) >= io_NSTREAMS, __LINE__)
-  entry=>io_stream(io_NSTREAMS)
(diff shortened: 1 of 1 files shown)


koldunovn:
Related to https://github.com/FESOM/fesom2/projects/1#card-54565263 and #69

koldunovn:
Jan, thanks a lot! I will do the testing this week (need some time after holidays :))

koldunovn:
@hegish I expand a bit the error message and make it print only once. If it's fine with you I will merge, everything works as expected! 👍

hegish:
Hi @koldunovn, regarding the error only from process 0:
I would not do it since we do not know if all processes participate in I/O activity. We also should not optimize for a failure case with additional code. I would not recommend it.
Regarding the extended error message: This is good, but we might not always add I/O streams from namelist only, but also just via the add stream API. I think this might help to understand the cause of the error. I would change "define it two times" to "define it multiple times", though.

koldunovn:
@hegish Point taken :) I remove the output only for `mype==0` and modify the message. It will clatter stdout with a bunch of messages, but I guess it's a rare case anyway :) Is it OK to merge form your perspective now?

hegish:
Looks good! I can not hit the approve button of the PR, maybe because I filed it...

PR #93, opened 2021-04-10
Awicm3 ice thermo cpl fixes

JanStreffing:
This branch is the result of extensive check of the sea ice coupling by Tido, Qiang, Dima Sidorenko, Wolfgang Dorn and others. The changes are confined to routines that are called when FESOM2 is coupled to OpenIFS.

The discussion behind this branch took place in https://github.com/AWI-CM3/project_digital_twin/issues/14

### DIFF
--- a/src/ice_thermo_cpl.F90
+++ b/src/ice_thermo_cpl.F90
@@ -132,17 +132,17 @@ subroutine thermodynamics(mesh)
         rsf     = 0._WP
      end if
 
-!#if defined (__oifs)
-!     !---- different lead closing parameter for NH and SH
-!     call r2g(geolon, geolat, coord_nod2d(1,inod), coord_nod2d(2,inod))
-!     if (geolat.lt.0.) then
-!        h0min = 0.75
-!        h0max = 1.0
-!     else
-!        h0min = 0.5
-!        h0max = 0.75
-!     endif
-!#endif /* (__oifs) */
+#if defined (__oifs)
+     !---- different lead closing parameter for NH and SH
+     call r2g(geolon, geolat, coord_nod2d(1,inod), coord_nod2d(2,inod))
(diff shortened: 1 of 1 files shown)


src/ice_thermo_cpl.F90
```diff
@@ -132,17 +132,17 @@ subroutine thermodynamics(mesh)
...
-!     call r2g(geolon, geolat, coord_nod2d(1,inod), coord_nod2d(2,inod))
-!     if (geolat.lt.0.) then
-!        h0min = 0.75
-!        h0max = 1.0
-!     else
-!        h0min = 0.5
-!        h0max = 0.75
-!     endif
-!#endif /* (__oifs) */
+#if defined (__oifs)
```
koldunovn, line 135: We had before the situation, when people where trying to optimize for `h0`, available in the namelist, but then hardcoded downstream. Will we get into similar trap here? Should we maybe expose those parameters explicitly through `namelist.ice` and comment that they only used in a coupled setting?

koldunovn (approved):
I am fine with it, have only one comment, that just a suggestion, feel free to consider or ignore it :) I would wait for at least another review before merge.

src/ice_thermo_cpl.F90
dsidoren: great job, great bug fixes! technically it can be merged as is. @koldunovn shall we introduce namelist.cpl or a section &cpl in one of the namelists for FESOM?

koldunovn: Maybe not for this PR, but later I think it's not a bad idea. If you OK with this PR I would merge as is.

JanStreffing: I would be ok with that. Any to have the namelist.cpl be a separate development.

PR #98, opened 2021-04-16
Update oce_ale.F90

patrickscholz:
Fix bug when make a restart from a partial cell:off simulation towards a partial cell:on simulation!


koldunovn:
Great! Should I test, or it's fine to merge?

koldunovn (approved):

patrickscholz:
it should be fine to merge, it should not influence the solution et all since it just affects this transition from pc:off to pc:on,

PR #100, opened 2021-04-16
Making sure that fesom2.1 can run coupled to OpenIFS43 on Aleph

JanStreffing:
Hello everyone,

during our meeting regarding AWICM3 physical biases in polar regions the point was raised that we should do all of our tests with the same versions for FESOM2 and OpenIFS. It also became clear that this version should be the 2.1 release version. I would like to make sure that all the physics fixes for the coupling of FESOM2 to OpenIFS43 are contained within the FESOM2.1 release. Furthermore there are some infrastructure changes that are also required to run FESOM2.1 on our production machine aleph.

I believe all of these merge can happen with bit identical results for standalone FESOM2.1.


koldunovn:
@JanStreffing I have updated fesom2.1 to master, since some of the changes are already there, and I want to update from  master anyway. Unfortunately the in this case GitHub does not recompute the diff (https://github.com/isaacs/github/issues/750) Can you, please, follow the solution described in this issue, and change base branch to something different and then change to fesom2.1 again?

JanStreffing:
it updates the commits and files but does not run checks again. I will close here and open a new one

Issue #99, opened 2021-04-16
MPI routine used before intialization of MPI, for coupled setups in ALEPH


mandresm:
Hi FESOM developers,

@JanStreffing and I are trying to run AWICM3 in ALEPH computer (a Cray system) and we are experiencing some issues involving MPI during run time.

**Branches used**

- FESOM: fesom-2.0-frontiers
- OASIS3MCT: awicm-3-frontiers

**Machine details**

Type: Cray
Batch system: PBS
Scheduler: ALPS (aprun)

**Compilation details**

We are using `ftn` compiler as it is the one recommended by the system admins for Cray. We are also using `mpich`.

**Cases tested**

- FESOM standalone (passes MPI initialization)
- **FESOM coupled** for AWICM3 has the error described below (setting the coupled flags on `FESOM_COUPLED ON` and `OIFS_COUPLED ON`. OASIS libraries are compiled before FESOM and linked).

**Detailed description of the problem**

We are experiencing the following issue during runtime of FESOM coupled in AWICM3:
```
Attempting to use an MPI routine before initializing MPICH
Attempting to use an MPI routine before initializing MPICH
Attempting to use an MPI routine before initializing MPICH
...
Attempting to use an MPI routine before initializing MPICH
[NID 00150] 2021-04-16 20:00:33 Apid 423459: initiated application termination
Fri Apr 16 20:00:33 2021: [PE_120]:inet_recv:inet_recv: recv error (fd=7) Connection reset by peer
Fri Apr 16 20:00:33 2021: [PE_120]:_pmi_network_barrier:_pmi_inet_recv from target 0 failed pmi errno -1
Fri Apr 16 20:00:33 2021: [PE_120]:_pmi_barrier:network_barrier failed
[Fri Apr 16 20:00:33 2021] [c0-0c2s6n1] Fatal error in MPI_Init: Other MPI error, error stack:
MPIR_Init_thread(537): 
MPID_Init(246).......: channel initialization failed
MPID_Init(638).......:  PMI2 init failed: 1 
Application 423459 exit codes: 1
Application 423459 exit signals: Killed
Application 423459 resources: utime ~3s, stime ~13s, Rss ~16864, inblocks ~51250, outblocks ~0
```

However, when we run the exact same version of FESOM-2.0 but compiled for standalone we seem to get further, as MPI reports initialization (we run standalone just to narrow down the problem, the crash that happens after MPI intialization is probably related to something we've done wrong in the configuration of the standalone test):
```
[...]

PR #104, opened 2021-04-20
hotfix to run awicm3 with mpi that is not intelmpi

JanStreffing:
This solves #99 and should be part of the release still.


JanStreffing:
Sorry should be part of https://github.com/FESOM/fesom2/pull/102/

PR #106, opened 2021-04-20
Rename some variables in mesh.diag

### LINKED ISSUE #87: Inconsistent dimension names in mesh diag and model output
Variable (dimension) names that represent same fields in model output and mesh diagnostics are named differently. Apart from consistency reason, having a same naming convention, will simplify post-processing and probably also make it easier for a new FESOM data user to understand.

I don't know of any plans to standardize names but I think at minimal, using same dimension names in mesh diag as in model data is useful.

for instance associations of names in mesh diagnostics to model data:

	nod_n  -> nod2
	elem_n -> elem
	nl -> nz
	nl1 -> nz1 

and I think, these level variables may be named same as dimension names harmlessly.

```
Zbar -> nz
Z -> nz1
```
I don't know if mesh diag file is used for restarts? and renaming may pose other challenges?

koldunovn:
Make naming of dimensions in `fesom.mesh.diag.nc` consistent with naming in the output.

Closes #87

### DIFF
--- a/src/io_mesh_info.F90
+++ b/src/io_mesh_info.F90
@@ -67,11 +67,11 @@ implicit none
   call my_create(filename, IOR(NF_CLOBBER,IOR(NF_NETCDF4,NF_CLASSIC_MODEL)), ncid)
 
   !Define the dimensions
-  call my_def_dim(ncid, 'nod_n',   nod2D,  nod_n_id)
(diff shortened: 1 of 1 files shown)


helgegoessling (commented):
spheRlab only uses nlevels and nlevels_nod2D, so these changes are no problem (and if they would have been affected, I could have adapted things in spheRlab).

Issue #85, opened 2021-03-25
Missing nod_in_elem2D data in mesh diag


suvarchal:
The variable `nod_in_elem2D` in  `fesom.mesh.diag.nc` which , I guess, is intended to provide indices of elements that contain a given node is filled with just zeros. 

I am coming from recently updated `pi-grid` test data in `pyfesom2`, so I wonder if this might be applicable to other FESOM2 simulations a well? 

To reproduce:
* [Download fesom.mesh.diag.nc](https://github.com/FESOM/pyfesom2/raw/master/tests/data/pi-grid/fesom.mesh.diag.nc) 
* `ncdump -v nod_in_elem2D fesom.mesh.diag.nc`

suvarchal:
If it helps: now I am certain that it might be a bug in fesom2 as the issue exists not just in pi grid example above but also in later versions of a core2 [data](https://swiftbrowser.dkrz.de/public/dkrz_c719fbc3-98ea-446c-8e01-356dac22ed90/PYFESOM2/) that was used in [overview example](https://nbviewer.jupyter.org/github/FESOM/pyfesom2/blob/master/notebooks/pyfesom2_overview.ipynb) of pyfesom2.

koldunovn:
This one is closed by #107

PR #107, opened 2021-04-21
a bug in nod_in_elem2D computation for mesh diagnostic file. could le…

dsidoren:
…ad to a segmentation fault.


koldunovn:
Closes #85

PR #120, opened 2021-05-03
Workbench fixsrfstress

patrickscholz:
exchange oce-atmos (stress_atmoce_x(n)) surface stress when computing KPP and MOMIX with the total surface stress (stress_iceoce_x(n)*a_ice(n) + stress_atmoce_x(n)*(1.0_WP-a_ice(n)))

It has surprisingly less effect after one full JRA55 cycle, but nevertheless the github testcase will of course fail. 

image

image


koldunovn:
@patrickscholz The question is why ICEPACK tests did not fail? :)

lzampier:
Sorry I cannot follow the discussion at the moment. I need a bit more context. I am having a look at the code right now.

lzampier:
It's strange because `call oce_fluxes_mom` is done in the same way for both versions... Is it possible to see the namelists used for the test?

patrickscholz:
its rather the actual default namelists you can check in /home/ollie/pscholz/fesom2.0_github/work_test_fixsrfstress_ctrl

JanStreffing:
Here is the tripyview evaluation of a 1958 to 2017 standalone core2 run with these changes vs the same period with the current main branch. 
[fesom26rc_fixsrfstress.pdf](https://github.com/user-attachments/files/16661031/fesom26rc_fixsrfstress.pdf)

@patrickscholz, I see some changes, especially in the Tropical upper ocean. Does this look ok to you?

patrickscholz:
@JanStreffing : Feels a bit strange since the main effect of this bug fix should not be in the equatorial area. There, not much should have changed, if there are changes than they should rather happen in the icecovered areas? When i made the testcase a while ago (see header figures) the effect looked way much smaller. Although i plotted the mean over the upper 250m

JanStreffing:
I might have made a mistake in my evaluation run. I made the reference run, and all other experiments for branches that change the results at the same time. I made this test simulation here more recently. I think that while updating this branch in preparation to merge, I may have merged some other the other result changing merges into here. In that case the difference, we see, would be the sum of multiple branches. 

I will make a new set of experiments with current main and this branch, to be 100% sure.

JanStreffing:
All clear. The new test that compares this branch to main has an order of magnitude smaller differences, and they are where we expect them, in high latitude regions:
[fesom26rc_fixsrfstress2.pdf](https://github.com/user-attachments/files/16677102/fesom26rc_fixsrfstress2.pdf)

JanStreffing (approved):

PR #122, opened 2021-05-05
print information about the currently used MPI library at FESOM startup

hegish:


koldunovn:
Checked on ollie, mistral, juwels.

PR #136, opened 2021-06-09
Update ice_thermo_cpl.F90 to fix the ice/snow surface temperature (273.15K)

trackow:
By checking the subroutine "ice_surftemp" I found that t, the ice surface temperature, is limited to the freezing temperature of sea water (TFrezs, l.491).
If I understand the code correctly, this can be artificially wrong by up to 2 Kelvin, because ice is in principle allowed to reach 0°C at the surface (admittedly there is a small salinity content in the ice, but none in the snow). 

While this should help with getting reduced ice thickness due to the smaller atm-ice/snow heat flux (smaller temp gradient), it could also help with fixing the snow (which melts first).

### DIFF
--- a/src/ice_thermo_cpl.F90
+++ b/src/ice_thermo_cpl.F90
@@ -488,7 +488,7 @@ contains
   zcprosn=rhosno*cpsno/dt               ! Specific Energy required to change temperature of 1m snow on ice [J/(sm³K)]
   zcpdte=zcpdt+zcprosn*hsn              ! Combined Energy required to change temperature of snow + 0.05m of upper ice
   t=(zcpdte*t+a2ihf+zicefl)/(zcpdte+con/zsniced) ! New sea ice surf temp [K]
-  t=min(TFrezs,t)                       ! Not warmer than freezing please!
+  t=min(273.15_WP,t)                       ! Not warmer than freezing please!
  end subroutine ice_surftemp
 
  subroutine ice_albedo(h,hsn,t,alb)
--- a/src/ice_thermo_cpl.F90
+++ b/src/ice_thermo_cpl.F90
@@ -474,6 +474,11 @@ contains
   real(kind=WP)  zcprosn
   !---- local parameters
   real(kind=WP), parameter :: dice  = 0.05_WP                       ! ECHAM6's thickness for top ice "layer"
+    !---- freezing temperature of sea-water [K]
+  real(kind=WP)  :: TFrezs
+
+  !---- compute freezing temperature of sea-water from salinity
+  TFrezs = -0.0575_WP*S_oc + 1.7105e-3_WP*sqrt(S_oc**3) - 2.155e-4_WP*(S_oc**2)+273.15
 
   snicecond = con/consn                 ! equivalence fraction thickness of ice/snow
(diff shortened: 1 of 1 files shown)


JanStreffing:
Makes sense. I had applied the sea water condition to the sea ice here.

JanStreffing:
I just went back through the commits. This used to be:
`ctfreez = 271.38_WP                   ! ECHAM6's temperature at which sea starts freezing/melting`
This has always been in the scheme. I think even on ECHAM6 side.

JanStreffing:
Conceptually, this would limit the maximum temperature in Summer, while not affecting winter. Would this not lead to less ice melt in summer because the temperature gradient is reduced? If so this might be good news for the SO. And maybe the Arctic has a different problem with the missing snow cover leading to more ice instead.

JanStreffing (approved):

trackow:
> I just went back through the commits. This used to be:
> `ctfreez = 271.38_WP ! ECHAM6's temperature at which sea starts freezing/melting`
> This has always been in the scheme. I think even on ECHAM6 side.

I see! In https://github.com/trackow/fesom2/blame/f28d26f415bf56318d931e667e33948b431fe9e6/src/ice_thermo_cpl.F90 it had still been 0°C (l.503, this was only later changed to units of Kelvin).

src/ice_thermo_cpl.F90
```diff
@@ -474,11 +474,6 @@ subroutine ice_surftemp(h,hsn,a2ihf,t)
   real(kind=WP)  zcprosn
   !---- local parameters
   real(kind=WP), parameter :: dice  = 0.05_WP                       ! ECHAM6's thickness for top ice "layer"
-  !---- freezing temperature of sea-water [K]
-  real(kind=WP)  :: TFrezs
-
```
trackow, line 479: TFrezs is still needed in line 480 for computation of

  zicefl=con*TFrezs/zsniced             ! Conductive heat flux through sea ice [W/m²]

so we should keep it?

JanStreffing: I'm not fully there today it seems. Yes.

JanStreffing (approved):

JanStreffing:
> > I just went back through the commits. This used to be:
> > `ctfreez = 271.38_WP ! ECHAM6's temperature at which sea starts freezing/melting`
> > This has always been in the scheme. I think even on ECHAM6 side.
> 
> I see! In https://github.com/trackow/fesom2/blame/f28d26f415bf56318d931e667e33948b431fe9e6/src/ice_thermo_cpl.F90 it had still been 0°C (l.503, this was only later changed to units of Kelvin).

IIRC that commit never ran. We explored two ideas initially, adapting the standalone FESOM2 scheme or adapting the AWICM2 scheme for sea ice surface temperature for AWICM3. The commit you liked was when we were still working on a coupled versions of the FESOM2 standalone scheme. 

I double checked and the ECHAM6 version is correct, they use tmelt = 0°C not ctfreez.

dsidoren:
@trackow @JanStreffing do we merge this?

JanStreffing:
Not necessary, it was already part of https://github.com/FESOM/fesom2/pull/156

PR #139, opened 2021-06-18
fix build error on macOS

hegish:


hegish:
Hi @koldunovn, there are errors in the icepack checks which are probably not a result from my changes here. What is going on?

koldunovn:
Hi @lzampier . Looks like you changed things in your `icepack_fesom2` branch of Icepack and now it breaks the tests. This also breaks stable 2.1 version, which is not that stable anymore :)

The ideal solution would be if you could revert the `icepack_fesom2` branch to the state it was during 2.1 release and continue to work in some development branch. If you think it's too hard or not possible, we can think of another solution. 

We also would need more safe solution for the Icepack integration in the future, at least to use the hash in `download_icepack.sh` instead of the branch name :) But we can discuss it in another tread.

lzampier:
Hi Nikolay, sorry for this issue. I have only slightly changed the value of two constants to increase the consistency between the standard sea ice thermodynamics and Icepack, but I did not think about the tests... I have reverted the commit so now the problem should disappear. I agree that using a hash when downloading Icepack would be better! We can do this soon, as we will anyway need to merge some changes coming from the work Patrick is currently doing.

koldunovn:
@lzampier Lorenzo, no problems at all, we still learning on how to do this things right :) I see that the new version of FESOM/Icepack is coming, so we will do everything right next time :) Thanks for reverting the changes, it makes things much easier on my side!

PR #142, opened 2021-06-22
Fix bug in freshwater balancing and drifting ocean volume when using icepack 

patrickscholz:
- Fix a variety of bugs in the FESOM2 - ICEPACK interface but also in ICEPACK itself (which is outsourced into branch icepack_fesom2_fixpscholz within the ICEPACK repository), so that we are able to fully balance the freshwater fluxes in FESOM2 zstar case and the ocean volume remains constant and does not drift away. 

- Of course with these changes the actual icepack github testcase will fail but the changes have no influence on the actual default case !


src/ice_setup_step.F90
```diff
@@ -31,6 +31,7 @@ subroutine ice_setup(mesh)
     ice_dt=real(ice_ave_steps,WP)*dt
     ! ice_dt=dt
     Tevp_inv=3.0_WP/ice_dt 
+    Tevp_inv=1.0_WP/1800.0_WP
```
koldunovn, line 34: That's a bit too radical :) We will have to make it a switch to be able to keep old behaviour.

lzampier: Hey @koldunovn! Agreed... I was just testing something on the branch and I forgot there was a pull request open associated with it! I will switch to a different branch and revert this change.

koldunovn: Nice to see you back in the FESOM2 repo! :)

lzampier: Yes, I gave myself a couple of days to clean a bit of the mess I left behind ;)

JanStreffing:
Master is no longer default. Can not be merged for 2.6. Shall be redone for 2.7.

JanStreffing:
Is this issue still relevant @fkauker, @lettie-roach, or can this be closed?

PR #151, opened 2021-07-08
give default values to uninitialized variables

hegish:
Using this would otherwise lead to undefined state at runtime and compiling these files lets the cray ftn compiler choke.

### DIFF
--- a/src/oce_ale_tracer.F90
+++ b/src/oce_ale_tracer.F90
@@ -342,6 +342,9 @@ subroutine diff_ver_part_expl_ale(tr_num, mesh)
     real(kind=WP)            :: zinv1,Ty
 
 #include "associate_mesh.h"
+
+    Ty = 0.0_WP
+    
     !___________________________________________________________________________    
     do n=1, myDim_nod2D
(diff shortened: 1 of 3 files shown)


JanStreffing:
Do we need this in the awicm-3-frontiers branch too?

hegish:
I am not sure if these parts are actually used. Otherwise the outcome would be pure luck. Can you merge it to the awicm-3-frontiers branch?

JanStreffing:
I'll do that.

JanStreffing:
Looking at it, it would merge all the master changes as well. I think we should wait till this is merged into master, then we can have a look if we want to merge the master changes of the last ~3 months into awicm-3-frontiers. I would want to discuss with Patrick and Dima ahead of time if:
a) if this is needed
b) if anything would break
Before I do that.

hegish:
Without it you should see alt least a bunch of warnings from the cray compiler at aleph. I can also cherry-pick this commit to awicm-3-frontiers if you like.

JanStreffing:
That would be a good quick fix for now, yes.

PR #159, opened 2021-07-28
 ignore the error message from reading the restart if the variable is…

dsidoren:
… not in the file (i.e. the model should not stop but use the values with which the variable was initialised initially)

### DIFF
--- a/src/io_restart.F90
+++ b/src/io_restart.F90
@@ -29,6 +29,7 @@ MODULE io_RESTART
     integer        :: ndim
     integer        :: dims(2) !<=2; assume there are no variables with dimension more than 2xNLxT
     real(kind=WP), pointer :: pt1(:), pt2(:,:)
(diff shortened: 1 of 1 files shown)


hegish:
There might be an issue when first skipping some variables (e.g. ice_albedo) but then writing them to the same file e.g. for monthly restart. In this case there is either garbage for the previous timesteps in the NetCDF file, or there are timesteps missing and reading from this restart will fail.
I now try how this problem can be solved for the same skip feature in awicm-3-frontiers_parallel-restart

dsidoren:
> There might be an issue when first skipping some variables (e.g. ice_albedo) but then writing them to the same file e.g. for monthly restart. In this case there is either garbage for the previous timesteps in the NetCDF file, or there are timesteps missing and reading from this restart will fail.
> I now try how this problem can be solved for the same skip feature in awicm-3-frontiers_parallel-restart

the rule shall be that it does not happen within the same restart file.

PR #167, opened 2021-09-20
Refactoring tracer part2

dsidoren:
How do you like it, Elon Musk?


patrickscholz:
@dsidoren 
Should we make a splitting of horizontal and vertical mesh component?

dsidoren:
> @dsidoren
> Should we make a splitting of horizontal and vertical mesh component?

i was also thinking about it but but both are often used together. I would suggest to limit upto ~10 new derived types for now and then split if needed.

PR #170, opened 2021-10-03
Update oce_ale.F90

patrickscholz:
Fix bug when making restart with cavity from partial cell off --> partial cell on


koldunovn (approved):
Checks are passes, so fine with me :)

PR #188, opened 2021-10-20
Chunking commented out again

mathanase:
Commenting out lines 621-623 in the routine `src/io_meandata.F90`.

This change is **needed on Mistral** (only).


hegish:
What is the problem with chunking on mistral? Is it a compile time or runtime problem?

JanStreffing:
It is a runtime problem. I think @dsidoren suspected that the MPI installations on mistral where somehow done without support for the feature. I understood that once the output is transposed this becomes obsolete, correct?

hegish:
> It is a runtime problem. I think @dsidoren suspected that the MPI installations on mistral where somehow done without support for the feature.

I do not see how this is tied to MPI.

> I understood that once the output is transposed this becomes obsolete, correct?
YES!

JanStreffing:
out of date, shall not be merged.

PR #193, opened 2021-11-01
be able to write transposed 3D output files via preprocessor definition

hegish:
as in 3bcd313, but intentionally without a switch to turn it off


trackow:
Great to follow your progress!

PR #200, opened 2021-11-05
Refactoring branch cray compatible (compiles with OpenMP)

trackow:
I mostly had to change the attributes, to make them consistent between the interface and subroutine definition. One slightly curious thing is that continuation of characters/strings requires an "&" as first nonblank character in the next line on cray.

I also changed the CMakeLists.txt file in src for the CRAY option. I added -homp (seems to be default for cray) and the line target_link_libraries(${PROJECT_NAME} OpenMP::OpenMP_Fortran) is ignored for CRAY (failed).


hegish (approved):

koldunovn:
Thomas, can you pull the latest version from refactoring?

PR #224, opened 2021-11-11
Merged ifs interface!

trackow:


hegish:
@trackow, this does not compile with BUILD_FESOM_AS_LIBRARY=ON
@koldunovn, I did not include the library build with the default executable build on purpose, but seeing this it may be a good idea if we a) always build the library together with the executable or b) have a test where the code is compiled with `cmake BUILD_FESOM_AS_LIBRARY=ON ...`

trackow:
I have been working on including a "BUILD_FESOM_AS_LIBRARY=ON" test all morning, give me some more time :) It compiled on cray before I made the pull request though, but I didn't test with intel or gnu yet.

Having said this, building both at the same time (library and binary) would be great, but didn't you say that you do not want to have two targets since it is error prone, @hegish ?

hegish:
> Having said this, building both at the same time (library and binary) would be great, but didn't you say that you do not want to have two targets since it is error prone, @hegish ?

Yes, so I have to think of a clever way to do it...

hegish:
You can do a custom build whenever you like. This script is intended for the casual user and looks a bit too complicated this way. But feel free to keep a copy and compile this way.

trackow:
Yes, agreed. But maybe while you are looking for a solution to compile both, think about the possibilty to have the libs (parms, metis, libfesom) seperate somewhere; this way we wouldn't need a custom configure at all.

trackow:
Or maybe having two build directories in the end

PR #231, opened 2021-11-13
add test case in fesom2.1.yml for compilation as library

trackow:


.github/workflows/fesom2.1.yml
```diff
@@ -19,10 +19,17 @@ jobs:
...
       run: |
         bash -l configure.sh ubuntu
-    
+
+    - name: Compile model (library)
+      run: |
+        bash ./test/ifs_interface/configure_lib.sh -l
+    - name: Library exists
+      run: |
+        bash ./test/ifs_interface/check_exist.sh
```
koldunovn, line 31: Thomas, did you check it it fails when library is not created?

trackow: I am returning exit code 0 in that case, so I was assuming this will be working.  From my local Docker, I did not know how to trigger the ./github/workflow

trackow: image

Issue #264, opened 2021-12-29
New restarts in refactoring not working on large meshes


koldunovn:
Attempt to run the latest `refactoring` branch on moderate (about 3M) mesh failed at the stage of restart writing. The stdout give:

```
Do output (netCDF, restart) ...
 initializing I/O file for sst
 initializing I/O file for m_ice
 initializing I/O file for a_ice
 initializing I/O file for vice
 initializing I/O file for MLD2
 initializing I/O file for MLD1
 error in line          101 async_threads_module.F90
 error in line          101 async_threads_module.F90
  --> actualize clock file to latest restart point
 error in line          101 async_threads_module.F90
 error in line          101 async_threads_module.F90
```

Sometimes also

```
error in line          101 async_threads_module.F90
 initializing I/O file for m_snow
 error in line          611 io_meandata.F90 NetCDF: File exists && NC_NOCLOBBER
```

The `namelist.config` settings are:

```
&restart_log
restart_length=1        !only required for d,h,s cases,  y, m take 1
restart_length_unit='m' !output period: y,  d, h, s, off
raw_restart_length=1
raw_restart_length_unit='m' ! e.g. y, d, h, s, off
logfile_outfreq=720     !in logfile info. output frequency, # steps
```

The setup can be found on ollie:
`/home/ollie/nkolduno/nextgems/refactoring/work_trop`

@hegish @patrickscholz @dsidoren

hegish:
- I could reproduce a similar error with core2
- I have fixed an issue (which I had introduced in the big merge of IO to refactoring)
- two runs with core2, 15 days, daily output and daily restart have been successful on ollie
- a job with trop mesh and the setup from above is pending in the queue

koldunovn:
Thanks a lot, @hegish ! So this is most probably not a problem of the big mesh then. So maybe you can make a PR with a fix without waiting the trop mesh to run? :)

hegish:
I would like to verify if the problems are really gone now. But this error has to be fixed anyway, so I did the PR already: https://github.com/FESOM/fesom2/pull/265

hegish:
... I forgot to add a file to the last commit. Next try...

hegish:
The run with trop mesh failed because FESOM tried to do a restart run. I do not know why yet...

koldunovn:
You probably should change the fesom.clock.

For me the `trop_mesh` run works, I will still have to check if the restarting itself works, but restarts are created :)

[...]

PR #272, opened 2022-01-20
use custom mkdir to avoid sporadic error with EXECUTE_COMMAND_LINE

hegish:


src/fortran_utils.F90
```diff
@@ -39,5 +39,30 @@ function int_to_txt_pad(val, width) result(txt)
     allocate(character(w) :: txt)
     write(txt,'(i0.'//widthtxt//')') val
   end function
+  
+  
+  ! using EXECUTE_COMMAND_LINE to call mkdir sometimes fail (EXECUTE_COMMAND_LINE is forked as an new process, which may be the problem)
```
dsidoren, line 44: Hi Jan, why do you write the C code for that? We generally use system command to avoid all these troubles which costs just 3 lines of code.
As an example:
    !    f%dump_dir='DUMP/'
    !    INQUIRE(file=trim(f%dump_dir), EXIST=f%L_EXISTS)
    !    if (.not. f%L_EXISTS) call system('mkdir '//trim(f%dump_dir))

hegish: Hi Dima, I did not really write any C code, I am just calling it from Fortran.
EXECUTE_COMMAND_LINE (or the older system) sometimes does not work and it is really hard to find out why. Fortran does it in a complicated way and has to fork a subprocess. My guess is that there are sometimes problems to do a fork due to memory problems or file count limitations. JSC support approves with the idea.
@koldunovn recently had this problem too, see https://github.com/FESOM/fesom2/issues/264#issuecomment-1015952096

koldunovn: @hegish But what happens if you just use `system`? It seem to just work fine in other parts of the code.

hegish: We also have EXECUTE_COMMAND_LINE in other parts of the code. I have also seen the error SYSTEM instead EXECUTE_COMMAND_LINE on aleph (I had them both active for a while). SYSTEM seems to behave similar to EXECUTE_COMMAND_LINE. The problem is, the error happens only rarely, see our discussion at https://github.com/FESOM/fesom2/issues/264.
Currently I have no better idea how to solve this.

koldunovn: I see. It's sad that one should have so much code for something that should be essentially a one liner :)

koldunovn (approved):

PR #278, opened 2022-01-25
Merging fesom branch in awicm3-v3.0 with latest fesom changes

mathanase:
Merging branches in order to integrate the latest fesom2 changes in the branch used in awicm3-v3.0

### DIFF
--- a/env/mistral.dkrz.de/shell-intel+openmpi
+++ b/env/mistral.dkrz.de/shell-intel+openmpi
@@ -1,0 +1,50 @@
+# make the contents as shell agnostic as possible so we can include them with bash, zsh and others
+
+module load gcc/4.8.2
+export LD_LIBRARY_PATH=/sw/rhel6-x64/gcc/gcc-4.8.2/lib64:$LD_LIBRARY_PATH # avoid GLIBCXX_3.4.15 not found error
+module unload intel && module load intel/18.0.4
+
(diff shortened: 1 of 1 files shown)


hegish:
is there only one changed file?

hegish:
I am not sure what the goal is here. Is it only the mkdir think? I could add these changes to the awicm3-frontiers-freeze-candidate-5 branch. Or should the awicm3-frontiers-freeze-candidate-5 branch also be up to date with all other changes in awicm-3-frontiers_parallel-restart?

mandresm:
I think the merge should be the other way around: `awicm-3-frontiers_parallel-restart` into `awicm3-frontiers-freeze-candidate-5`. @hegish , I think you are right, the only thing needed here is the mkdir changes, or all the changes necessary for getting rid of the restarting writing problems.

mathanase:
I am not quite sure actually. Maybe @JanStreffing would know better.

JanStreffing:
Ok I had a look. This is indeed the wrong way round. It's a merger of a branch that has been frozen since September into the one that @hegish is developing IO solutions in. I will close this here and make a reversed merge request. Then we have to check if we want everything in `awicm-3-frontiers_parallel-restart `for the physics frozen `awicm3-frontiers-freeze-candidate-5`, or if some things would have the model results.

PR #280, opened 2022-01-26
Awicm 3 frontiers parallel restart

JanStreffing:
I can see two commits here that we do not want in the frozen branch, as they change the model physics.
d968f0a3e3f664a6ee843489565240dd1a0edda4 96e8da16dbcd58083d91c355e4c2bc9935394a59
We could merge and revert manually. Is there an option to merge and exclude these here?
Are there any other changes that would impact the model physics? @hegish 

I can't find you in the list of users, @mathanase, so adding you here.


mandresm:
>

mandresm:
Sorry for closing it, I'm pretty clumsy... 

> I can see two commits here that we do not want in the frozen branch, as they change the model physics. [d968f0a](https://github.com/FESOM/fesom2/commit/d968f0a3e3f664a6ee843489565240dd1a0edda4) [96e8da1](https://github.com/FESOM/fesom2/commit/96e8da16dbcd58083d91c355e4c2bc9935394a59) We could merge and revert manually. Is there an option to merge and exclude these here? Are there any other changes that would impact the model physics? @hegish
> 
> I can't find you in the list of users, @mathanase, so adding you here.

There are a number of ways this can be done. One is to do the merge manually on git with `git merge --no-commit <branch>`. This does the merging but stops right before doing the commit, so that you can change things around, for instance, `git checkout HEAD <whatever_file_you_don't_want_to_get_modified>. Once you have only what you selected then you can go ahead and do `git commit` and `git push`. This will do the merge and GitHub will recognise that the merge had occurred externally. 

The other option is to cherry-pick from the `awicm3-frontiers-freeze-candidate-5`. For cherry picking I like the following flags: `git cherry-pick -x -n <ID>`. `-n` stops before committing the cherry-pick so you can check the changes with `git diff HEAD`. If everything is fine you can do `git commit` and the cherry pick is finished. `-x` appends the original commit ID  to the cherry-pick comment.

hegish:
> I can see two commits here that we do not want in the frozen branch, as they change the model physics. [d968f0a](https://github.com/FESOM/fesom2/commit/d968f0a3e3f664a6ee843489565240dd1a0edda4) [96e8da1](https://github.com/FESOM/fesom2/commit/96e8da16dbcd58083d91c355e4c2bc9935394a59) We could merge and revert manually. Is there an option to merge and exclude these here?

I would go for the obvious route: merge and then revert. I consider tinkering with the merge dangerous, as it is not version controlled. It is also possible to loose the connection to the parent branches this way.

> Are there any other changes that would impact the model physics? @hegish

Not that I know of.

hegish (approved):

koldunovn:
I would really advice against revert. Reverting means that if you decide to actually get some of the changes in the future, you will not be able to, there will be "canceling" commit in your branch. Had this trouble with refactoring branch.

JanStreffing:
Would making a branch, reverting there, and then merging keep everything in order?

hegish:
I we ever want to un-revert a revert, we can just cherry-pick it.

koldunovn:
> Would making a branch, reverting there, and then merging keep everything in order?

I don't think so, the branch will contain "reverting instructions". See here more explanation of the revert: https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt

hegish:
We are talking about reverting the two commits with unwanted changes to physics, right? Not a revert of a previous merge.

mandresm:
My favorite option for this task is cherry-picking, there you really select commits, without messing future merges, plus is less work IMO

koldunovn:
> We are talking about reverting the two commits with unwanted changes to physics, right? Not a revert of a previous merge.

Yes, but `revert` **will create a new commit that inverses the changes**, so there will be not possible (or rather hard) to add those commits bach, when (if) you need them. If you sure you never want them in your code, then revert is fine.

hegish:
I just tried it: git revert SHA and then git cherry-pick SHA works fine.

JanStreffing:
> > We are talking about reverting the two commits with unwanted changes to physics, right? Not a revert of a previous merge.
> 
> Yes, but `revert` **will create a new commit that inverses the changes**, so there will be not possible (or rather hard) to add those commits bach, when (if) you need them. If you sure you never want them in your code, then revert is fine.
This branch shall never contain them. It's the frozen branch for AWI-CM release 3.0. Only technical changes are allowed.

JanStreffing:
I ran a set of testruns on juwels with the new v3.0. Everything looks fine. @mathanase should you run into any trouble, let me know.

mathanase:
Do I need to change to OpenMPI as well (so recompile my awicm3 there)?

JanStreffing:
Hello Marylou, I don't think you have to switch MPI. But a git pull on fesom2 and a recompile should fix any mkdir related errors that you may have had. (I know I had some in the past)

PR #294, opened 2022-02-07
Ollie CI

pgierz:
Here is a start to CI directly on ollie. So far, the compile test goes through, the library one not yet. Everything else will need to be added step by step.


pgierz:
Closing and deleting branch, CI directly on HPC will probably not be happening anytime in the foreseeable future.

PR #307, opened 2022-02-21
Correct confusion of gamma's with gamma's_tra

tsemmler05:
The bug fix to correct the confusion of gamma0, gamma1, and gamma2 with gamma0_tra, gamma1_tra, and gamma2_tra should be put into the awicm-3-frontiers_parallel-restart branch.


JanStreffing:
Closed for inactivity. Not needed anymore.

PR #309, opened 2022-02-23
declare parms functions before they are used

hegish:
otherwise it can not be compiled with e.g. clang 13.0.0


JanStreffing:
Parms shall be removed in next version.

PR #321, opened 2022-03-15
activated coupling of ice_temp & ice_alb with IFS

dsidoren:
@trackow

### DIFF
--- a/src/gen_forcing_couple.F90
+++ b/src/gen_forcing_couple.F90
@@ -172,7 +172,7 @@ subroutine update_atm_forcing(istep, ice, tracers, partit, mesh)
      do i=1,nsend
          exchange  =0.
          if (i.eq.1) then
-#if defined (__oifs) || defined (__ifsinterface)
+#if defined (__oifs)
             ! AWI-CM3 outgoing state vectors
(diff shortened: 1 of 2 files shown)


trackow (commented):
The change in gen_forcing_couple.F90 is not related to the coupling of ice temp and albedo, is it? But I also think the __ifsinterface is not needed there.

Is ice_temp and ice_alb always computed? I would have expected an __oifs flag somewhere where a || __ifsinterface statement would need to be added...

dsidoren:
> The change in gen_forcing_couple.F90 is not related to the coupling of ice temp and albedo, is it? But I also think the __ifsinterface is not needed there.

right it is not related. everything with __oasis is not related to IFS. there was an __oasis directive above so the whole piece was deactivates in any case.
> 
> Is ice_temp and ice_alb always computed? I would have expected an __oifs flag somewhere where a || __ifsinterface statement would need to be added...

we will include the piece with computation from FESOM&OIFS (it is only in the master branch currently). all directives  __ifsinterface are complete as far as I could see.

trackow:
Okay, that answers everything. From my side it can go in

koldunovn (approved):

PR #325, opened 2022-03-25
adapting oasis coupling interface to new datastructures

JanStreffing:
Compiles and runs coupled to OpenIFS via OASIS. Please double check the style. It's the first time I look at the new data structures.


hegish:
looks OK to me regarding the new derived types. Are you aware of the other changes, e.g. in  ice_thermo_cpl.F90 h0min/h0max and more?

JanStreffing:
Yes, those changes were made so that this branch reflects what we learned while developing the AWI-CM3 release. It will have to be double checked by @trackow, as it would appear, that on this branch the same lines of code are also used for the IFS single executable coupling.

hegish (approved):

trackow:
> @trackow I changed the lead closing and ice albedo in the coupled case to match the AWI-CM3 DECK paper. Do you want to take those changes as well?

Yes, thanks Jan. I went through the commits and I think it is good to do this consistently in AWI-CM3 and IFS-FESOM.

trackow (commented):
Not sure why there are __ifsinterface statements inside the __oasis block, that is never reached for IFS


#if defined (__oasis)
     if (firstcall) then
        allocate(exchange(myDim_nod2D+eDim_nod2D), mask(myDim_nod2D+eDim_nod2D))
        allocate(a2o_fcorr_stat(nrecv,6))
        a2o_fcorr_stat=0.
        exchange      =0.
        mask          =0.
        firstcall=.false.
     end if
     do i=1,nsend
         exchange  =0.
         if (i.eq.1) then
#if defined (__oifs) || defined (__ifsinterface)

Looking at it more closely, I think all __ifsinterface blocks are not needed/potentially wrong in gen_forcing_couple.F90 as introduced [here](https://github.com/FESOM/fesom2/commit/013d7153ad7367db86ecc26635564a6dbf2886a7)

trackow:
Most of the gen_forcing_couple.F90 actually needs to be ignored for the IFS case. The sbc_do is correctly ignored, but also not sure right now whether the #endif was introduced too early [here](https://github.com/FESOM/fesom2/commit/aca0d0fda8b440d8ae7f6e5569c9695b334c669b)

trackow:
The wind stress was potentially already wrong during the Cycle 1 simulations [gen_forcing_couple.F90.txt](https://github.com/FESOM/fesom2/files/8366274/gen_forcing_couple.F90.txt) The OASIS block ends after the stress computation (#endif /* (__oasis) */), but there is unfortunately an ELSE case that IFS-FESOM still enters. I think this is a bug and could fit to something weird with the winds that @lzampier and @koldunovn once saw

      !__________________________________________________________________________
     dux=u_wind(i)-u_ice(i) 
     dvy=v_wind(i)-v_ice(i)
     aux=sqrt(dux**2+dvy**2)*rhoair
     stress_atmice_x(i) = Cd_atm_ice_arr(i)*aux*dux
     stress_atmice_y(i) = Cd_atm_ice_arr(i)*aux*dvy
  end do

  ! heat and fresh water fluxes are treated in i_therm and ice2ocean
#endif /* (__oasis) */

The #ifndef __ifsinterface case that was supposed to skip sbc_do() should probably end only here. Need to test this and confirm

JanStreffing:
Superseeded by #352

PR #326, opened 2022-03-30
AWIESM: Unified version for FESOM 2.1

pgierz:
We want to see if we can get a unified version of FESOM for inclusion in AWI-ESM. I could merge this by hand, but we could also let the CI system try if the basic tests still go through.


pgierz:
@koldunovn, any idea why the CI is failing?

@PengyangSong, these changes look mostly like what you had for the caverns. Can you make a look and double check please?

PengyangSong:
> @koldunovn, any idea why the CI is failing?
> 
> @PengyangSong, these changes look mostly like what you had for the caverns. Can you make a look and double check please?

I can have a try later today, but I am a bit slow on this.
I can give you a branch that can pass the test by the end of this week, if you are not hurry on that.
But I think the code itself do not have a bug.

pgierz:
@PengyangSong: passing tests would be great. I just merged this (for testing) and at least the tests I have described here https://github.com/esm-tools/esm_tools/pull/623 seem to be working. If the results are the same, I do not know yet, but the model does not crash 😄

koldunovn:
@pgierz The reason currently is very old `env` folder, that don't have ubuntu option. I suspect the same is true for `env.sh`.

pgierz:
@koldunovn, I already via email told Pengyang to take the version in the refactoring branch, that may solve the issue.

PengyangSong:
@koldunovn Can you have a look on this?
https://github.com/FESOM/fesom2/runs/5769281321?check_suite_focus=true
I have already updated env and env.sh, but somewhere related to cmake stopped this. Do you have any idea on which file?

koldunovn:
Something related to OASIS, is it something that should be used? I guess `CMakeLists.txt` in multiple places are wrong.

pgierz:
@PengyangSong I suspect you commit in a version that enabled the coupled mode flag by default. There should be in the main CMakelists file a -D that can turn that on or off, and the default should be off -- FESOM is designed standalone first, we just tend to enjoy abusing it for climate purposes ;-)

pgierz:
@PengyangSong OK, fine by me! :-) I only changed the bool OFF to ON, so you can use the version I just made, but this did not solve the problem.

pgierz:
Hi @PengyangSong: passing tests, hooray! It seems though that you modified some of Lars' iceberg code. Please get in touch with him to see if that is what we want to do. In the end, I want to support both caverns and icebergs, so, ideally you should not have to delete anything

PengyangSong:
@pgierz can you give me a hint on which file I changed? I think I took care of the merging very carefully.

koldunovn:
@PengyangSong and @pgierz If you change test results, you should really know what you are doing, and be sure this is what you want :)

pgierz:
@koldunovn: yes. I would advocate for putting any "extra physics" behind a namelist switch so that we don't intrude on the main model.

@PengyangSong, look at he diff here: for me it looks as there are many iceberg related changes that have been removed. https://github.com/FESOM/fesom2/compare/eb7e03558c7948d2d7fad5baad401082da02e651..f2d6ab33e66e1a6b1fc749a18f0ca7e2c999d100

PengyangSong:
@koldunovn I really want to check the test results of fesom2.1 branch but the log is missing, so it cannot compare. I think in physics it should be the same with fesom2.1 branch.
https://github.com/FESOM/fesom2/runs/2729145754?check_suite_focus=true
https://github.com/FESOM/fesom2/runs/2729145763?check_suite_focus=true

koldunovn:
The test results should match ones in the setups. For FESOM2.1 it would be https://github.com/FESOM/fesom2/blob/master/setups/test_pi/setup.yml

PengyangSong:
@pgierz These icb_* files are ahead of my branch (wiso & wiso-2.1 & fesom2.1 do not have these). So these are not deleted by me. Instead, these should be added. Maybe the merging strategy is a little stupid.

koldunovn:
> The test results should match ones in the setups. For FESOM2.1 it would be https://github.com/FESOM/fesom2/blob/master/setups/test_pi/setup.yml

By "should" I mean if you want to have exactly the same results as current FESOM2.1. If you intention to change physics, that's a different story :)

PengyangSong:
@koldunovn thanks for this information. Actually I kept all things unchanged. I do not know what causes these. But I will check anyway.

PR #332, opened 2022-04-19
Add -fpe0 to Intel compile options for IFS-FESOM

trackow:
Added -fpe0 to Intel compile options for running IFS-FESOM in the RAPS environment. Without it, the model often
fails with GM and Redi if those are not set to .false. (model blowup or floating point exceptions)


trackow:
Can we add this, @dsidoren ? :)

dsidoren (approved):

PR #333, opened 2022-04-26
switch off update_atm_forcing when coupled with operational IFS

dsidoren:


JanStreffing:
already in production_DE.

Issue #178, opened 2021-10-10
momix_length outputted by default


koldunovn:
One can't turn it off from the `namelist.io`.

koldunovn:
Closed by #351

PR #351, opened 2022-06-11
Refactoring update tke+idemix

patrickscholz:
- Move IDEMIX  mixing operation from vertices towards elements, since only on elements we are able to properly define the inflow of tidal energy from the bottom
- Do the same for TIDAL mixing (Simmons) same problem as in IDEMIX
- Fix Bug in io_restart.F90 --> before the variables for tke and iwe were not properly written, so after restart were initialised with zeros
- Add io_meandata.F90, so that data for TKE, IDEMIX, TIDAL and FORC (forcing) are written only when they are listed in namelist.io in iolist with the corresponding frequency and precision only when they are used or when the corresponding flag (ldiag_forc) is set to .true.

- `&nml_list
io_list =  ...
           'TKE         ',1, 'y', 4, 
           'IDEMIX    ',1, 'y', 4,  
           'TIDAL      ',1, 'y', 4,  
           'FORC      ',1, 'y', 4,  
           ...
/
`

Solves #178


koldunovn:
@patrickscholz should we merge this?

koldunovn:
I would gladly merge it, but there is a conflict now, that should be resolved :)

patrickscholz:
@koldunovn i think you can merge this, conflict are solved!

koldunovn (approved):

PR #300, opened 2022-02-14
UGRID Compliant Output

pgierz:
This adds metadata to the `fesom.mesh.diag.nc` and output files for [u-grid](https://ugrid-conventions.github.io/ugrid-conventions/#2d-flexible-mesh-mixed-triangles-quadrilaterals-etc-topology) compliance. 

If it works, it would close #273.

Example `ncdump -h` of the new `mesh.diag.nc` file:
```
$ ncdump -h test/output_pi/fesom.mesh.diag.nc
netcdf fesom.mesh.diag {
dimensions:
	nod2 = 3140 ;
	edg_n = 8986 ;
	elem = 5839 ;
	nz = 48 ;
	nz1 = 47 ;
	n2 = 2 ;
	n3 = 3 ;
	n4 = 4 ;
	N = 8 ;
	info = 1 ;
variables:
	int fesom_mesh ;
		fesom_mesh:cf_role = "mesh_topology" ;
		fesom_mesh:long_name = "Topology data of 2D unstructured mesh" ;
		fesom_mesh:topology_dimension = 2 ;
		fesom_mesh:node_coordinates = "lon lat" ;
		fesom_mesh:face_node_connectivity = "elements" ;
		fesom_mesh:edge_node_connectivity = "edges" ;
		fesom_mesh:edge_face_connectivity = "edge_tri" ;
		fesom_mesh:face_dimension = "elem" ;
		fesom_mesh:edge_dimension = "edg_n" ;

```

and for SST:

```
$ ncdump -h test/output_pi/sst.fesom.1948.nc
netcdf sst.fesom.1948 {
dimensions:
	nod2 = 3140 ;
	time = UNLIMITED ; // (1 currently)
variables:
	double time(time) ;
		time:long_name = "time" ;
		time:standard_name = "time" ;
		time:units = "seconds since 1948-01-01 0:0:0" ;
		time:axis = "T" ;
		time:stored_direction = "increasing" ;
	double sst(time, nod2) ;
		sst:description = "sea surface temperature" ;
		sst:long_name = "sea surface temperature" ;
		sst:units = "C" ;
		sst:location = "nodes" ;
		sst:mesh = "fesom_mesh" ;
```

I would ask @koldunovn to check the code, and @Chilipp to check if this works nicely with plotting tools already available.

PR #354, opened 2022-06-15
Refactoring better metadata

pgierz:
Replacement for #300, as my private fork was somehow messed up.


pgierz:
More to do before merge: remove the "2" files created by my cloud backup. Sorry about that

pgierz:
Hi @koldunovn: this is good to go from my side. I restored the old dimension names `nod2` and `elem` instead of `nnode` and `nface` (that was the example on the U-Grid website); but that would have broken a ton of downstream tools. Maybe it is nicer not to do that ;-)

koldunovn:
@pgierz Are we missing location of the centers of elements in `fesom.mesh.diag.nc`?

pgierz:
Should we have those as well? I can add it. The ugrid check passes, but if it useful no problem to include it.

koldunovn:
@pgierz That would be great! I don't think we should limit ourselves by ugrid minimum, and I would add as much info as possible, ideally most of the things from the mesh object :)

JanStreffing:
@pgierz will have time to look at this in April. Either we can resolve the MR conflict by ourselves or we need to push this MR to the next release.

JanStreffing (approved):

PR #366, opened 2022-08-18
Stress

dsidoren:


koldunovn:
The branch initially supposed to add functionality to wind stress, but also possibility to add sea ice as initial conditions were developed here as well. So the status before merge is:
- The wind stress functionality is not tested, and might just not work.
- The sea ice initial conditions are tested and works fine.

PR #374, opened 2022-11-03
Draft: fixing the finalize for the fesom2-openifs coupling, as well as verbose coupled

JanStreffing:
I tried running a coupled AWI-CM3 simulation with the refactoring branch and verbose output today. Here is a quick fix for the errors I encountered.

### DIFF
--- a/src/cpl_driver.F90
+++ b/src/cpl_driver.F90
@@ -66,7 +66,6 @@ module cpl_driver
   REAL(kind=WP), POINTER                          :: exfld(:)          ! buffer for receiving global exchange fields
   real(kind=WP), allocatable, dimension(:,:)      :: cplsnd
 
-  real(kind=WP)              :: time_send(2), time_recv(2)
 
   integer, dimension(1,3)    :: iextent    
   integer, dimension(1,3)    :: ioffset   
@@ -571,13 +570,6 @@ contains
     end if
     t3=MPI_Wtime()
     
-    if (ind==1) then
-       time_send(1)=t3-t1       
-       time_send(2)=t2-t1
-    else
-       time_send(1)=time_send(1)+t3-t1
-       time_send(2)=time_send(2)+t2-t1
-    endif	         
   end subroutine cpl_oasis3mct_send
 
   ! ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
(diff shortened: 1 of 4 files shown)


JanStreffing:
Some tests are failing. Those seem unrelated?

JanStreffing:
its not unrelated. Interesting, I was compiling for me before. I'll have another look

patrickscholz:
You edit some things in gen_modules_partitioning.F90 which might be the problem. He doesnt seem to compile gen_modules_partitioning.F90 --> so no mod_parsup.mod file that needs to compile gen_halo_exchange.F90. 

This whole thing looks weired...
```
#if defined (__oifs)
!OIFS coupling doesnt call prism_terminate_proto but cpl_oasis3mct_finalize through which it calls oasis_terminate
  use cpl_driver 
  implicit none
  integer,           intent(in)   :: COMM
  integer,           intent(in)   :: mype
  integer, optional, intent(in)   :: abort
  integer                         :: error
  if (present(abort)) then
    if (mype==0) write(*,*) 'Run finished unexpectedly!'
    call MPI_ABORT(COMM, 1 )
  else
    if (mype==0) write(*,*) 'Finalizing MPI communication through oasis'
    call cpl_oasis3mct_finalize
    if (mype==0) write(*,*) 'MPI communication closed down'
  endif
#endif
```
I dont think you can use 
```
 use cpl_driver 
  implicit none
  integer,           intent(in)   :: COMM
  integer,           intent(in)   :: mype
  integer, optional, intent(in)   :: abort
  integer   
```
at the end of the subroutine, this needs to be defined in the header of the routine !

JanStreffing:
I thought everything before is under an #ifndef __oifs block. So that's actually the start of the par_ex subroutine in the __oifs case.

JanStreffing:
I was able to reproduce the behavior from before. If the compile without cleaning up, adding the `use cpl_driver` works. If I then clean and compile again I get the error as in the CI test.

This is what tripped my up. I guess it's a circular dependency with the use statements?

JanStreffing:
I can make it work by calling the oasis_terminate directly in par_ex. I guess this is what we did with the old prism_terminate_proto block for AWI-CM1&2 as well.

While I'm here, I have a mind to clean this routine up a bit and make it more readable with regards to preprocessor statements.

patrickscholz (approved):

PR #375, opened 2022-11-08
Refactoring bugfix restart zstar2linfs

patrickscholz:
- Make sure that the restart works from zstar towrads linfs
- add another echange_nod when Sergey Solver is used


koldunovn:
@patrickscholz fine to merge?

patrickscholz:
@nikolay Yes should be ok to merge !

PR #376, opened 2022-11-08
Nvfortran subarray workaround

hegish:
@patrickscholz, not sure if this also has be changed for reading/writing the dwarf memory dumps?


WiltonLoch:
Hello, 
This is my personal Github account which I'll be using for the porting of FESOM.
The problem we are having now happens in the io_restart_derivedtype.F90 file when running the tracer dwarf.
Apparently there are no changes to this file in this PR. Do you think that after the fix is merged, if we deactivate the current PGI check, the remaining part of the code will work as expected?

dsidoren (approved):

hegish:
@patrickscholz, can you have a look if we need to implement the workaround fur the dwarf memory dumps as well?
@WiltonLoch, is this version still failing at runtime?

PR #377, opened 2022-11-10
 make sure when we cutoff a_ice=0.0 that we also cutoff m_ice and m_snow

patrickscholz:
When we cut off the ice concentration towards zeros we should make sure that we also cutoff the m_ice and m_snow, when there is no concentration than there can be also no ice and snow height. Somehow this issue slipped into the only coupled case. The difference is not much, its just a small inconsistency but its enough to most likely blow up the test case. 

Its not a burning issue,  for the case we next time have to redo the testcase we can also merge that issue!


koldunovn:
Do you know why it was defined only for the IFS case? Not to break the tests? :)

JanStreffing:
@koldunov, could you evaluate and compare
`/work/ab0246/a270092/runtime/fesom-2.5-refactoring/reference/outdata/fesom/`
and
`/work/ab0246/a270092/runtime/fesom-2.5-refactoring/bugfix_ice_cutoff/outdata/fesom/` 
?

JanStreffing:
Here is the tripyview evaluation of a 1958 to 2017 standalone run with the new cutoff values removed vs the same period with the current main branch. I have nowhere to host the html, so as a pdf instead:
[fesom26rc_bugfix_ice_cutoff.pdf](https://github.com/user-attachments/files/16387584/fesom26rc_bugfix_ice_cutoff.pdf)

JanStreffing:
As expected, the changes are minute. I will go ahead, update the truth and merge.

JanStreffing:
Requires review to merge @dsidoren, @koldunovn, @trackow.

koldunovn (approved):

JanStreffing:
Ty :)

PR #385, opened 2022-11-26
vertical smooth for bvfreq. S. Koldunov reported noise in bv in verti…

koldunovn:
…cal direction.


koldunovn:
@dsidoren We checked results on CORE simulations - they are different, but differences are tiny. If @dsidoren is fine with it we can merge after I update the tests. Probably make sense to combine with #377

dsidoren (approved):

PR #388, opened 2022-11-28
Cycle3 updates

trackow:
The current updates for Cycle 3 include

- coupling of surface ocean currents (took nodal velocities instead of velocities on elements)
- optimization of coupling interface (packing of fields, p2p instead of alltoallv) following what was done for NEMO4
- extra diagnostics are available on IFS grid (MLD, ave T 300m, ave S 300m, 20C isotherm, etc) and thus in GRIB output
- reduced standard I/O output in single-executable runs

To come:
- include snow enthalpy fix


JanStreffing:
> The current updates for Cycle 3 include
> 
>     * coupling of surface ocean currents (took nodal velocities instead of velocities on elements)

Hello Thomas, I'd be interested in a few more details on this one. What are the advantages of doing so, and should we do that for the oasis coupling too?

trackow:
> > The current updates for Cycle 3 include
> > ```
> > * coupling of surface ocean currents (took nodal velocities instead of velocities on elements)
> > ```
> 
> Hello Thomas, I'd be interested in a few more details on this one. What are the advantages of doing so, and should we do that for the oasis coupling too?

Hey Jan, it was mostly being pragmatic since then we don't need to generate a separate set of remapping weights from the element mid points to the Gaussian grid, just for the velocities. So now we just take them from the nodes, similar as for the tracers. Also, we had faced an issue with the global numbering of elements, since elements can belong to more than one processor, right? All in all it seemed easiest to go this way for the moment.

JanStreffing:
I see. I had used `dynamics%uvnode(:,:,:)` Anything speaking against that?

trackow:
Yes, I think that is perfectly fine. For the IFS coupling, fesom%dynamics%uvnode(1,1,:) and fesom%dynamics%uvnode(2,1,:) are now used

dsidoren (approved):

PR #391, opened 2022-12-02
Refactoring albedo env

patrickscholz:
- add preliminary ALBEDO environment and job submission files


koldunovn:
ready to merge?

patrickscholz:
@koldunovn  FESOM compiles and runs on albedo. Problem that remains is that at least two albedo nodes have to be used. On only one albedo node the model gets stucked at the io stream definition. Further i had the impression that the IO is a bit slow  in general. There we have to see if its due to the preliminary state of albedo or if we still need to tweak something. But to have a starting i think it can be merged than also others can make some test on albedo.

koldunovn (approved):

PR #400, opened 2022-12-21
Refactoring albedo env

patrickscholz:
- Fix 3d output issues on albedo using jans workaround for aleph
- cleaned up and commented io_meandata.F90 so i was able to debug it


koldunovn (approved):

hegish:
Why do we disable multithreading by default in Fesom? I do not think any regular user will bother to explicitly activate it and so it will be just slow from now on.

patrickscholz:
This might be a mistake from my side to switch it back after my debugging!!!

Issue #407, opened 2023-01-10
allocatable character


dsidoren:
https://github.com/FESOM/fesom2/blob/5baae1788b02a56354a26ff90908d5c7560437d9/src/io_netcdf_file_module.F90#L8

As noticed by @trackow ftn cray compiler does not like it and requires static allocation. We need to do it static or come up with an alternative solution the earlier the better @hegish

koldunovn:
I would say we urgently need a version of refactoring that works with lumi version of ftn...

PR #412, opened 2023-01-16
an unelegant fix to run with CCE (FTN fortran compiler) on LUMI. Checks

dsidoren:
for _CRAYFTN macros. "character(:), allocatable :: name" -> "character(1000)           :: name"

### DIFF
--- a/src/io_netcdf_file_module.F90
+++ b/src/io_netcdf_file_module.F90
@@ -5,7 +5,11 @@ module io_netcdf_file_module
   private
 
   type dim_type
+#if !defined(_CRAYFTN)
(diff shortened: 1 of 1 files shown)


koldunovn (approved):

hegish:
The compiler error is only for this file? Strange...

dsidoren:
> The compiler error is only for this file? Strange...

yes, we had an issue on that https://github.com/FESOM/fesom2/issues/407

hegish:
> > The compiler error is only for this file? Strange...
> 
> yes, we had an issue on that #407

But is this the issue we talked about last Monday?

Issue #21, opened 2020-10-09
Corners for SCRIPR/CONSERV 1st order conservative remapping


JanStreffing:
Currently we are using global conservative flux residual redistribution for AWICM (1,2 & 3). Never versions of OASIS3 (MCT3 & MCT4) are offering a locally conservative remapping. I had a short discussion with the EC-Earth community and their idea is to switch from GAUSWGT remapping mit conservative redistribution to CONSERV remapping. @helgegoessling was also interested

1. I think this would in principle also be of interest to us. Maybe @dsidoren, can comment.
2. Similar to how we already write out the center lon/lat (`coord_nod2D(1, i)` & `coord_nod2D(2, i)`) for the current remapping schemes we would need corner lon/lats for CONSERV.

Would this from oce_mesh be the right vector? `mesh%x_corners(n,mesh%nod_in_elem2D_num(n))`

For reference, this is what the [OASIS manual states](http://www.cerfacs.fr/oa4web/oasis3-mct_4.0/oasis3mct_UserGuide.pdf):
> **If the SCRIPR/CONSERV remapping is specified, longitudes and latitudes for the source and target grid corners must also be available in the grids.nc file as double precision REAL arrays dimensioned(nx,ny,4) or (nbrpts,1,4) where 4 is  the  number  of  corners  (in  the  counterclockwize sense, starting by any corner).** The names of the arrays must be composed of the grid prefix and the suffix “.clo” or “.cla” for respectively the grid corner longitudes or latitudes.  As for the other grid information, the corners can be provided in grids.nc before the run by the user or directly by the component code through specific calls (see section 2.2.4). **Longitudes must be given in degrees East in the interval -360.0 to 720.0. Latitudes must be given in degrees North in the interval -90.0 to 90.0.** Note that if some grid points overlap, it is recommended to define those points with the same number (e.g. 360.0 for both, not 450.0 for one and 90.0 for the other) to ensure automatic detection of overlap by OASIS3-MCT. The  corners  of  a  cell  cannot  be  defined  modulo  360  degrees.   For  example,  a  cell  located  over Greenwich will have to be defined with corners at -1.0 deg and 1.0 deg but not with corners at 359.0deg and 1.0 deg.

helgegoessling:
[...]

PR #432, opened 2023-03-10
Feat/oasis corners 2nd try

JanStreffing:
related to issue #21

While this adds corners, it does only add the element center corners. The center of element vertices corners are missing. Introducing them would make the control volume convex, which OASIS does not like:

Image descriptionImage description

The features is good enough for stochastic coupling based on area fraction but not accurate enough for conservative remapping.


JanStreffing:
back to wip, adding edge centers.

patrickscholz (approved):

JanStreffing:
wait still wip

JanStreffing:
It writes the corners correctly:

image


but then crashes afterwards. I still need to look into it.

PR #440, opened 2023-03-21
setting Intel bach as default compiler for Levante. Replacing env/lev…

dsidoren:
…ante.dkrz.de/shell.nvhpc->shell will allow to compile with NV & OpenACC support

### DIFF
--- a/env/levante.dkrz.de/shell.nvhpc
+++ b/env/levante.dkrz.de/shell.nvhpc
@@ -2,22 +2,14 @@
 export LC_ALL=en_US.UTF-8
 export CPU_MODEL=AMD_EPYC_ZEN3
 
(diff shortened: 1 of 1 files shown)


env/levante.dkrz.de/shell
suvarchal, line 1: does it make sense to add multio directories to it ?

suvarchal (changes requested):
may be multio directory pointing to a user's directory is not elegant?

koldunovn (approved):

env/levante.dkrz.de/shell
suvarchal: resolving for now as it is harmless

suvarchal (approved):
moving on for now

PR #441, opened 2023-03-22
change atos compile options and add platform strategy

trackow:
Compile options that work stably on Atos ECMWF with IFS-FESOM2.5 (TCo1279-NG5 and TCo399-FORCA025)


suvarchal (changes requested):
configure_any.sh is not any more relevant? if we skip that we can merge src/CMakelists.txt safely.

JanStreffing:
@trackow Does this still make sense to merge for 2.6?

JanStreffing:
@trackow And also the question for this one: Merge or ignore?

trackow:
If it was up to me, I think we should just make sure everything from this tag is in refactoring https://github.com/FESOM/fesom2/blob/2.5.0.3_de/src/CMakeLists.txt

JanStreffing:
I did a test merge from that tag to refactoring. 100% of that tag is converd by https://github.com/FESOM/fesom2/pull/571 already. I suggest we merge that and close here.

PR #442, opened 2023-03-23
Fesom2.5 intel fixes

trackow:
Test to see it worked...

### DIFF
--- a/src/io_meandata.F90
+++ b/src/io_meandata.F90
@@ -973,6 +973,8 @@ subroutine write_mean(entry, entry_index)
         ! allocate global 2d array in which local data are gathered
         if(entry%p_partit%mype==entry%root_rank) then
             if(.not. allocated(entry%aux_r8)) allocate(entry%aux_r8(size2))
+        else
+            if(.not. allocated(entry%aux_r8)) allocate(entry%aux_r8(0))
         end if
         
         !_______________________________________________________________________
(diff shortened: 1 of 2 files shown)


JanStreffing:
@trackow Does this still make sense to merge for 2.6?

JanStreffing:
@trackow Any update on this PR?

trackow:
The io_meandata.F90 changes by Dima and the added k-loop in oce_fer_gm.F90 I don't know what they are for, but we should check with him @dsidoren .     

```
 !Intel vectorisation did something strange in the above lines hence we had to code it more explicitely
        nzmax=mesh%nl
        nzmin=1
        do k=1, nod_in_elem2D_num(n)
           nzmax=min(nzmax, nlevels(nod_in_elem2D(k, n)))
           nzmin=max(nzmin, ulevels(nod_in_elem2D(k, n)))
        end do
```

The rest, for example, MPI_BARRIERs and OMP barriers I would not touch as they worked well in https://github.com/FESOM/fesom2/tree/2.5.0.3_de. So would again suggest to just merge that branch instead

JanStreffing:
We leave this open until after merge of https://github.com/FESOM/fesom2/pull/571 and check what's left then. @patrickscholz said the io_meandata changes might be good still and not covered elsewhere.

JanStreffing:
This had been left over after 2.6. Not much is left from the original PR. @trackow, of the bits that are left, they fall in two chunks: The MPI Barrier in oce_ale_pressure_bv.F90, and the cmake changes.

Essentially we disable omp with intel compiler for a few files with Multithreading is on. I think this is still needed, correct, @trackow?

JanStreffing:
Closed until confirmed that this is still needed. @suvarchal is working on parallel netcdf output redone.

PR #443, opened 2023-03-25
extend  configure build system to take CMAKE options in commandline and add frame work to directly use multiple shells for a machine

suvarchal:
- extend configure.sh to take cmake command line build options and FESOM specific options like, ENABLE_OPENMP, ACC etc as command line arguments to configure.sh. for instance: `./configure.sh -DENABLE_OPENMP -DUSE_ICEPACK etc` without needing to edit/change CMakeLists.txt. This makes building model and testing easier for developer who frequently builds model with different options and to test building of FESOM with different  compile options in  CI without needing to change the CMakeLists files. 

- add a template via  configuration for levante  to be able to add and use different shells (compilers) for a given machine without needing to copy the shell.xyz file to shell manually. For instance on levante a user can now compile FESOM using intel compiler using:  `./configure.sh  levante.intel` or `./configure.sh levante.nvhpc` for pgi/nv compiler or `./configure levante.gnu`. The default shell is a soft link to shell.intel used when `./configure.sh` is used in levante. Basically, a developer can add different shells as shell.anyname in env and use it  in cmd line as `./configure.sh machine.anyname`

- added levante shell for gnu compiler.


hegish:
I notice you put a question about whether env.sh is being sourced or executed in the code. I would like to clarify this in a comment. Should I add the corresponding commit it to this PR?

koldunovn (approved):

PR #457, opened 2023-04-20
fixes depth dimension for nextGEMS upper variables

koldunovn:
Add proper dimension to the "upper variables"


JanStreffing:
Not used anymore.

PR #458, opened 2023-04-21
Initial OpenACC porting for the dynamics

wiltonloch:
Initial ported version with decorations and data movement optimization for some of the dynamics subroutines.

The ported version is also running with the full model.
In the development environment the results are bit identical. For this to be achieved the compute_neutral_slope subroutine also has to be disabled, as the hyperbolic tangent function is not supported in the uniform CPU/GPU math algorithms.

Let me know if there are any changes required for the merge.


JanStreffing:
@koldunovn: Can this be re-used in the current generation conversion efforts?

JanStreffing:
@wiltonloch I had this run through some initial check with Claude. It appears some work would be left. Is this something you would still have time to persue yourself?

wiltonloch:
Sorry for the delay @JanStreffing! But no, we don't have capacity do pursue this now.

PR #459, opened 2023-04-21
Version of tracer advection porting with loops collapsed

wiltonloch:
Tracer advection OpenACC ported version with loops collapsed. This builds on top of the current porting.

This version is also running with the full model.
In the development environment the results are bit identical between CPU and GPU (deactivating the atomics + no FMA + gpu math uniform algorithms).

Let me know if there are any changes required for the merge.


JanStreffing:
This PR will be harvested in other means.

PR #460, opened 2023-04-26
do a raw restart at the end of the simulation if __ifsinterface is de…

dsidoren:
…fined!


mandresm:
Hi @koldunovn, I'm manually testing that this reproduces the expected behavior right now.

PR #472, opened 2023-06-23
Fix restart: when tke or iwe files do not exist, make existence of iwe and tke optional

patrickscholz:
make existence of tke and iwe files optional for restart so that the restart also works when the files dont exist like in a restart from TKE to TKE+IDEMIX. If the files would be needed but they do not exist they are simply initialised with zeros for the first time step

allow for restart from:
- tke  --> tke+idemix
- kpp --> tke 
- kpp --> tke+idemix  

should also allow restart for:
- viscop5 --> viscopt8
- viscop6 --> viscopt8
- viscop7 --> viscopt8

### DIFF
--- a/src/io_restart.F90
+++ b/src/io_restart.F90
@@ -82,15 +82,15 @@ subroutine ini_ocean_io(year, dynamics, tracers, partit, mesh)
   !___Save restart variables for TKE and IDEMIX_________________________________
 !   if (trim(mix_scheme)=='cvmix_TKE' .or. trim(mix_scheme)=='cvmix_TKE+IDEMIX') then
   if (mix_scheme_nmb==5 .or. mix_scheme_nmb==56) then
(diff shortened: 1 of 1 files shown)


JanStreffing:
Commenting to keep this MR in mind, as I plan to do the same for sea ice temperature and albedo in the coupled setup. Those can be missing if one switches from standalone FESOM2 to coupled FESOM2.

patrickscholz:
@JanStreffing i think ice_temp and albedo are already optional at least in the latest refactoring branch
 `
#if defined (__oifs)
  call ice_files%def_node_var_optional('ice_albedo', 'ice albedo',    '-',   ice%atmcoupl%ice_alb, mesh, partit)
  call ice_files%def_node_var_optional('ice_temp', 'ice surface temperature',  'K',   ice%data(4)%values, mesh, partit)
#endif /* (__oifs) */
`

koldunovn (approved):

JanStreffing:
ah, nice I guess that was a case of the user not using an older branch then. ty!

PR #478, opened 2023-07-19
AWIESM 2.5 FESOM branch

pgierz:
Combines work of @YuchenSun33 and @PengyangSong. Let us see what the automatic tests do.

Also CC as info for @gknorr and Martin Werner (who's GitHub user name I cannot find right now)

If this combination (plus any fixes) works correctly, it should become frozen as the new default, and please communicated to the group to make sure that everyone uses this one. I can help with any git-fiddling if needed.


JanStreffing:
It's been two years, is this PR still needed?

PR #487, opened 2023-08-11
Refactoring razvan multio

dsidoren:


src/ifs_interface/ifs_interface.F90
```diff
@@ -9,6 +9,46 @@ MODULE nemogcmcoup_steps
...
+    INTEGER :: icomm
+    LOGICAL :: lnemoioserver
+    INTEGER :: irequired, iprovided
+    LOGICAL :: lmpi1
+ 
+    CALL mpp_io_init(icomm, lnemoioserver,  irequired, iprovided, lmpi1)
+END SUBROUTINE nemogcmcoup_init_ioserver
+ 
+ 
+SUBROUTINE nemogcmcoup_init_ioserver_2( icomm )
```
suvarchal, line 28: should we remove this? subroutine?

suvarchal (approved):
@dsidoren  say "I Agree" to everything in the comments and it is automatically merged.

src/ifs_interface/ifs_interface.F90
dsidoren: indeed, nowhere else used

PR #488, opened 2023-08-21
Add same information from GH readme to RTD docs

kinow:
Hello,

I was working on some review feedback in an eFlows4HPC, where we were asked to add the acronym of FESOM in the document.

I accessed the RTD site but I couldn't find it. I found it later on fesom.de. The GitHub README has this information, so I copied that over here, along with the introductory paragraph from GH's README.

Local preview:

image

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,6 +6,15 @@
 FESOM2 documentation
 ====================
 
+The Finite volumE Sea Ice-Ocean Model (FESOM2).
+
+Multi-resolution ocean general circulation model that solves
+the equations of motion describing the ocean and sea ice using
+finite-volume methods on unstructured computational grids. The
+model is developed and supported by researchers at the Alfred
+Wegener Institute, Helmholtz Centre for Polar and Marine
+Research (AWI), in Bremerhaven, Germany.
+
 Authors
    -------
 


docs/index.rst
```diff
@@ -6,6 +6,15 @@
 FESOM2 documentation
 ====================
 
+The Finite Element Sea Ice-Ocean Model (FESOM2).
```
koldunovn, line 9: The FESOM 2 is finite-volume version, so: "Finite volumE Sea Ice-Ocean Model (FESOM2)"

docs/index.rst
```diff
@@ -6,6 +6,15 @@
...
+
+Multi-resolution ocean general circulation model that solves
+the equations of motion describing the ocean and sea ice using
+finite-element and finite-volume methods on unstructured
```
koldunovn, line 13: remove finite-element, that's for FESOM 1 :)

koldunovn:
Thanks a lot, @kinow , it totally make sense! I made a couple of suggestions, so things are valid for FESOM2 :)

docs/index.rst
kinow: Done!

docs/index.rst
kinow: Done (updated paragraphs to balance the text & linebreaks)

kinow:
> Thanks a lot, @kinow , it totally make sense! I made a couple of suggestions, so things are valid for FESOM2 :)

Thanks @koldunovn for the review and explanation on the different meaning for FESOM1 and FESOM2. I think it could be useful to have that too somewhere in the docs (if not already, like in a "History" section). But probably for another issue. Edited commit and push forced with the changes. Thanks!!!

koldunovn (approved):

PR #493, opened 2023-09-04
Refact/compiler as input in configure

mandresm:
@basava70 and myself could not compile FESOM with NVHPC using `configure.sh` unless we made this modification in the `configure.sh` to give as an argument `levante.nvhpc`.

With this change one can run:
```bash -l configure.sh```
To run the default compiler for the machine or:
```bash -l configure.sh levante.<compiler>```
To select which `env/levante.dkrz.de/levante.<compiler>` you want to use

### CODE REVIEW

**Comment by suvarchal** [comment]
I totally second this, i personally don't find what we have elegant to first degree where we expect user to manually copy or link their compiler env to shell file before configure.sh. that said above change breaks our current workflows used for development like ./configure.sh ubuntu etc so that $1 argument has to parsed cautiously before passing. also any solution should also work for our ./configure_any.sh for using fesom as library.

**Review by suvarchal — COMMENTED** [review]
did you forget to add and commit env.sh ? because something needs to be done to $SHELLFILE.

**Comment by mandresm** [comment]
> that said above change breaks our current workflows used for development like ./configure.sh levante.dkrz.de or ubuntu(tests) or many regexes we use in env.sh

Why? You can still give no argument and get the exact same result as before: `bash -l ./configure.sh`

> did you forget to add and commit env.sh ? because something needs to be done to $SHELLFILE.
Aren't those lines already there?
https://github.com/FESOM/fesom2/blob/0c2de0af0c9d80bf8e5814c0ed39ebd66bf4dbe0/env.sh#L31-L37

**Comment by patrickscholz** [comment]
Why this change lets the testcases fail ?

**Comment by mandresm** [comment]
> Why this change lets the testcases fail ?

and most importantly, how did it work before I added the change? Maybe I'm just doing something silly. Before this change, how would you typically compile in refactoring with a different compiler than the default?

**Comment by patrickscholz** [comment]
... should this not be done automatically via cmake depending on what compiler module gets provided via the environement files?!

**Comment by mandresm** [comment]
Yes but there are more than one `env/levante.dkrz.de/shell` file, and you need to select which one you want to use. Without this change I cannot source the `env/levante.dkrz.de/shell.nvhpc`.

**Comment by patrickscholz** [comment]
Most likely the levante NATESM people just changed the link from hand ?
```
~/fesom2_refactoring/env/levante.dkrz.de/
drwxr-sr-x  2 pscholz hpc_user 4096 May  9 14:47 .
drwxr-sr-x 18 pscholz hpc_user 4096 May  9 14:47 ..
lrwxrwxrwx  1 pscholz hpc_user   11 May  9 14:47 shell -> shell.intel
-rwxr-xr-x  1 pscholz hpc_user 1645 May  9 14:47 shell.gnu
-rwxr-xr-x  1 pscholz hpc_user 1582 May  9 14:47 shell.intel
-rwxr-xr-x  1 pscholz hpc_user 1143 May  9 14:47 shell.nvhpc
```

we could provide a further argument to ./configure.sh, like ...
./configure -c intel to enforce a certain compiler if the corresponding environment file exists and change the link with configure.sh?

**Comment by suvarchal** [comment]
> Aren't those lines already there?
> 
> https://github.com/FESOM/fesom2/blob/0c2de0af0c9d80bf8e5814c0ed39ebd66bf4dbe0/env.sh#L31-L37
> 
> The problem was that `env.sh` had all the machinery to do precisely what we need, but `configure.sh` was not passing any arguments to `env.sh`

env.sh sourced inside configure.sh should get all the args passed to configure.sh, we consume what we want and pop that(using shift) now back in configure.sh rest of the arguments should be available. so passing $1 to env.sh should not necessary in bash at least, do you have zsh? can imagine it is same but don't know. 

So my guess why test fails is some how args get messed up and cmake gets some other build argument? just just tests, i can't use ./configure.sh ubuntu on my machine when we do this. 

on comment on if you forgot to add env.sh I thought you had some generic fix for future for all envs like i have ubuntu/shell.gcc shell.intel in my local env. I would love to use it like you intend but cant imagine elegant way of doing it for all platforms.

**Comment by suvarchal** [comment]
> Most likely the levante NATESM people just changed the link from hand ?
> 
> ```
> ~/fesom2_refactoring/env/levante.dkrz.de/
> drwxr-sr-x  2 pscholz hpc_user 4096 May  9 14:47 .
> drwxr-sr-x 18 pscholz hpc_user 4096 May  9 14:47 ..
> lrwxrwxrwx  1 pscholz hpc_user   11 May  9 14:47 shell -> shell.intel
> -rwxr-xr-x  1 pscholz hpc_user 1645 May  9 14:47 shell.gnu
> -rwxr-xr-x  1 pscholz hpc_user 1582 May  9 14:47 shell.intel
> -rwxr-xr-x  1 pscholz hpc_user 1143 May  9 14:47 shell.nvhpc
> ```
> 
> we could provide a further argument to ./configure.sh, like ... ./configure -c intel to enforce a certain compiler if the corresponding environment file exists and change the link within configure.sh?



> Yes but there are more than one `env/levante.dkrz.de/shell` file, and you need to select which one you want to use. Without this change I cannot source the `env/levante.dkrz.de/shell.nvhpc`.

manually :laughing: linking shell to what they want. I also want a better solution then this :smile:

**Comment by mandresm** [comment]
> env.sh sourced inside configure.sh should get all the args passed to configure.sh, we consume what we want and pop that(using shift) now back in configure.sh rest of the arguments should be available. so passing $1 to env.sh should not necessary in bash at least, do you have zsh? can imagine it is same but don't know.

You are right, sorry with the confusion, we try several stuff and only stop when something worked. So ignoring my changes, in a pristine `refactoring` version I have the following problem:
```
$ bash -l configure.sh levante.nvhpc
Sourcing /work/ab0995/a270152/model_codes/fesom-ref/env/levante.dkrz.de/shell.nvhpc for environment
CMake Error: The source directory "/work/ab0995/a270152/model_codes/fesom-ref/build/levante.nvhpc" does not exist.
Specify --help for usage, or press the help button on the CMake GUI.
```

But as @suvarchal mentioned, the argument of `configure.sh` is passed to `env.sh`, so my proposed solution is not good, the problem is somewhere else, probably in the linking to `build`

mandresm:
> > Aren't those lines already there?
> > https://github.com/FESOM/fesom2/blob/0c2de0af0c9d80bf8e5814c0ed39ebd66bf4dbe0/env.sh#L31-L37
> > 
> > The problem was that `env.sh` had all the machinery to do precisely what we need, but `configure.sh` was not passing any arguments to `env.sh`
> 
> env.sh sourced inside configure.sh should get all the args passed to configure.sh, we consume what we want and pop that(using shift) now back in configure.sh rest of the arguments should be available. so passing $1 to env.sh should not necessary in bash at least, do you have zsh? can imagine it is same but don't know.
> 
> So my guess why test fails is some how args get messed up and cmake gets some other build argument? Not just tests, i can't use ./configure.sh ubuntu on my machine when we do this.
> 
> on comment on if you forgot to add env.sh I wondered if you had some generic fix for future for all envs in that case SHELLFILE would use shell.$compiler compiler is parsed from input first arg.

I think you pointed out this before, but also @patrickscholz did: now it seems that the first argument is used to specify the machine or machine+compiler. I think we probably want to separate those, as @patrickscholz is suggesting with a `-c` flag for specifying the compiler.

suvarchal:
> > on comment on if you forgot to add env.sh I wondered if you had some generic fix for future for all envs in that case SHELLFILE would use shell.$compiler compiler is parsed from input first arg.
> 
> I think you pointed out this before, but also @patrickscholz did: now it seems that the first argument is used to specify the machine or machine+compiler. I think we probably want to separate those, as @patrickscholz is suggesting with a `-c` flag for specifying the compiler.

I was thinking about this, problem to me is how to do it elegantly with current set of features. so now we can send ./configure.sh machine -DENABLE_OMP=on -DCMAKE_blah_blah. machine is popped out in env.sh if it exists and rest are passed to cmake. Now if we have -c we need to filter that out before we send to cmake, if we assume -c only happens in combination with machine and next arg then may be easy otherwise can get ugly. is there any other better way.....

patrickscholz:
@suvarchal in this context i see the simplicity of using "maschine+compiler" as first input argument for configure.sh, and we scan for the + to see if a different compiler is wanted to be used. But than this functionality needs to be documented and lets hope that there will be never a HPC maschine with a "+" in its name.

mandresm:
Thanks guys! This seems also a very good solution to me!

suvarchal:
made commits to this PR to support + option to specify compiler, tests pass and should be backwards compatible. So ./configure.sh levante+nvhpc or levante+intel should work now can you check. and then anyone @mandresm or @patrickscholz  can merge?

mandresm:
Seems to work pretty nicely for @basava70 and myself. Thanks a lot @suvarchal and @patrickscholz!

patrickscholz (approved):

PR #498, opened 2023-09-20
Fesom multio standalone

suvarchal:
To be able to use MULTIO from fesom standalone and few other bug fixes that came along with making fesom as library. 
TODO: 
- [ ] To test if FESOM with multio works when used with IFS.

### commit notes
- [x] modify mpp_io to initialize Multio server and finalize Multio for standaldone to be minimally invasive in fesom source.
- [x] stop multio server gracefully (although not the best way) 
- [x]  cmake module changes: add cmake module to find multio library (using MULTIO_INSTALL_PATH env variable) 
- [x] CMakelist changes: add multio defines, add multio api modules to include and libs to link while compiling.
- [x] CMakeList changes:make MPI necessary as it is needed any ways for FESOM as MPI calls are hardcoded and will be found automatically instead of setting MPI_HOME etc in shell.
- [x] use more robust cmake's find_package for netcdf and MULTIO
- [x] add comments to remind to use cmake toolchains in future instead of hard coding machine specific configurations in CMakeLists.txt.


sebastianbeyer:
Hey, I talked with Willem from ECMWF and he is currently rebasing some work that he has done on top of Razvan's FESOM-multio branch to be able to compile inside the bundle with and without multio enabled in FESOM. Maybe we could wait with this standalone thing and see if that already solves the IFS-side?

sebastianbeyer:
I think this are the changes he was talking about: https://github.com/wdeconinck/fesom2/commits/nextGEMS_c3_fesom2.5_climateD_multio

suvarchal:
> Hey, I talked with Willem from ECMWF and he is currently rebasing some work that he has done on top of Razvan's FESOM-multio branch to be able to compile inside the bundle with and without multio enabled in FESOM. Maybe we could wait with this standalone thing and see if that already solves the IFS-side?

I think, that should not be a problem, to deal with those changes. But there is still work to be done here meanwhile in terms of what  multio-api we use as it is changed significantly (15h ago there were substantial changes in multio api) i guess these changes are linked to multio detecting "ocean model" .

patrickscholz (approved):

PR #499, opened 2023-09-21
fixes #496 blas discovery 

### LINKED ISSUE #496: FESOM2 standalone simulation from documentation
I need to do a fesom standalone simulation for a new mesh representing the mid-Pliocene ocean conditions. This issue is to update on errors encountered while following the [documentation](https://fesom2.readthedocs.io/en/latest/getting_started/getting_started.html), as discussed with @koldunovn, @pgierz, @patrickscholz, and Sesh (I don't know his username).

Machine: `levante`
Branch: `refactoring`
Mesh: `/work/ab0246/a270179/runtime/awicm3-v3.1/input/fesom2/midpli/`
Model directory: `/work/ab0246/a270179/runtime/awicm3-v3.1/model_codes/fesom-standalone/fesom2/`

**First step: build model executable**

- To do the mesh partitioning, one has to switch from `master` to `refactoring` branch, otherwise it does not update the available machines for submitting the job. 

After doing this, the mesh partitioning was easy to perform following documentation.

**Second step: running the model**

- My mesh is of mid-Pliocene conditions based on COREII. Is there something to be done prior to running the model? Which naming should I use, `core2` or `midpli`?
- I created a link on `/work/ab0246/a270179/runtime/awicm3-v3.1/model_codes/fesom-standalone/fesom2/results/` for results and created a `fesom.clock` there using the start year of 1990 instead of 1958. Is that ok?
- changes made to `namelist.config`:
```
&timestep
step_per_day=32 !96 !96 !72 !72 !45 !72 !96
run_length=10 !62 !62 !62 !28
run_length_unit='y'             ! y, m, d, s
/

&clockinit              ! the model starts at
timenew=0.0
daynew=1
yearnew=1990
/

&paths
MeshPath='/work/ab0246/a270179/runtime/awicm3-v3.1/input/fesom2/midpli/'
ClimateDataPath='/pool/data/AWICM/FESOM2/INITIAL/phc3.0/'
ResultPath='/work/ab0246/a270179/runtime/awicm3-v3.1/experiments_testing/fesom-standalone/'
/
```
***Notes:*** there should be more description on having to define a result path in `namelist.config`, and of what is exactly the run length.

- changes made to `/work/ab0246/a270179/runtime/awicm3-v3.1/model_codes/fesom-standalone/fesom2/work/job_levante`
```
#SBATCH --job-name=midpli_test1
#SBATCH -p compute
#SBATCH --ntasks-per-node=108
#SBATCH --ntasks=432
#SBATCH --time=00:30:00
#SBATCH -o slurm-out.out
#SBATCH -e slurm-err.out
#SBATCH -A ab0246
```
***Notes:*** None of the above-mentioned lines are described in the documentation, except the number of tasks, which depends on the distribution chosen. 

It does not run and slurm messages and other files are kept in the work folder, instead of in the results folder. Is it ok?
Error message in `slurm-err.out`:
```
/var/spool/slurmd/job6868133/slurm_script: line 14: /work/ab0246/a270179/runtime/awicm3-v3.1/model_codes/fesom_standalone/fesom2/env/levante.dkrz.de/shell.intel: No such file or directory
217: fesom.x: error while loading shared libraries: libmkl_intel_lp64.so.2: cannot open shared object file: No such file or directory
```

suvarchal:
- don't use blas when not found as blas is optional for parms.
- use cmake's internal module to find blas as default way to find blas.

if imkl is used for lapack/blas configure.sh may need -DBLA_VENDOR=Intel10_64lp as arg to make blas discoverable.

### DIFF
--- a/lib/parms/CMakeLists.txt
+++ b/lib/parms/CMakeLists.txt
@@ -6,7 +6,10 @@ project(parms C)
 set(src_home ${CMAKE_CURRENT_LIST_DIR}) # path to src directory starting from the dir containing our CMakeLists.txt
 file(GLOB all_sources ${src_home}/src/*.c ${src_home}/src/DDPQ/*.c)
 
-include("${CMAKE_CURRENT_LIST_DIR}/../../cmake/FindBLAS.cmake")
+find_package(BLAS) # standard way to find blas
+if( NOT BLAS_FOUND) # try older way to find blas
+   include("${CMAKE_CURRENT_LIST_DIR}/../../cmake/FindBLAS.cmake")
+endif()
 
 # create our library (set its name to name of this project)
 if(${BUILD_FESOM_AS_LIBRARY})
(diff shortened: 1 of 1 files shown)


patrickscholz:
Hi @suvarchal, thats great ... one step further! But on levante the compiling still cant fully finish! There is still the error message:
```
CMakeFiles/fesom.dir/oce_ale.F90.o: In function `solve_ssh_ale_':
/home/a/a270042/fesom2/src/oce_ale.F90:2682: undefined reference to `psolve'
/home/a/a270042/fesom2/src/oce_ale.F90:2674: undefined reference to `psolver_init'
make[2]: *** [src/CMakeFiles/fesom.dir/build.make:1642: src/fesom] Error 1
make[2]: Leaving directory '/home/a/a270042/fesom2/build'
make[1]: *** [CMakeFiles/Makefile2:140: src/CMakeFiles/fesom.dir/all] Error 2
make[1]: Leaving directory '/home/a/a270042/fesom2/build'
make: *** [Makefile:139: all] Error 2
``` 
... it looks like without blas there is no parms and without parms there is no standard solver ?

patrickscholz (approved):

patrickscholz:
Does not using blas for parms has any effect on the performance of the solver ?

suvarchal:
> Does not using blas for parms has any effect on the performance of the solver ?

I guess it should, we could make BLAS as REQUIRED instead of being optional. but then for levante if we want it to be smooth installation we may have then add if levante, if intel block and set(BLA_VENDOR Intel10_64lp) before FindBLAS discovery which is ugly. another reason we should start using toolchains for differenet machines.

PR #504, opened 2023-10-04
together with Suvi here comeis the healpix remapping for the nodal fi…

dsidoren:
…elds. The target is globally alocated at 0 root. It might be not really efficient yet but needs to be tested.


JanStreffing:
done multio side

PR #521, opened 2023-10-25
Fix compilation issues using Hecuba (2.1_intel) on MN4

kinow:
Hi @suvarchal ,

Yolanda noticed the compilation issues with this branch, and suggest a fix (`uint32_t` instead of `int32_t`). I've updated the modules I used to compile on MN4, and we also enabled `c++11` as that's what Hecuba 2.11_intel uses.

With this we were able to successfully compile this branch.

Could you, please,

i) confirm this branch can be used with Alien4Cloud
ii) check if these changes look OK

Thanks!!
Bruno

### DIFF
--- a/env/bsc/shell
+++ b/env/bsc/shell
@@ -1,7 +1,4 @@
-
-module load intel/2020.1 impi/2018.4 mkl/2020.1 bsc/1.0 netcdf/4.2
-
+module use /apps/HECUBA/modulefiles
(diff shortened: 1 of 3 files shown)


suvarchal:
merging this Jorge wants to build new container with it and see. changes seem okay, i could compile it and run local workflow.

otherthings: i don't know if it works from alien4cloud as i dont have an account yet.

kinow:
I will compile again using this branch and let you know if it works from Alien4Cloud (but it should, in theory). Thanks @suvarchal !

PR #537, opened 2023-11-13
Make async_threads_module and fesom_C an OBJECT library

wdeconinck:
This PR responds to commit https://github.com/FESOM/fesom2/commit/d887a600b1537a55d8079fdded38762cd9372d96 by @dsidoren needed to recover performance. This was due to compile flags not being added to the Fortran module `async_threads_module`. 
That revert currently disallows shared libraries as the callback function `async_threads_execute_fcall` needs to be available for the library `async_threads_cpp` when built as shared library.
The proposed solution is to build `async_threads_cpp` as object library and have the objects included in libfesom. The same is proposed for libfesom_C (which is only one file `psolve.c`).


wdeconinck:
ping @dsidoren @suvarchal

PR #549, opened 2023-12-20
Remove old Cray exception for OpenMP - builds with tagged ifs-bundle

trackow:
When building the tagged ifs-bundle with FESOM, target_link_libraries(${PROJECT_NAME} PRIVATE OpenMP::OpenMP_Fortran) were for some reason not set for Cray compiler on LUMI. This normally sets OpenMP compiler flags, link flags, libraries automatically. 

target_compile_options() for Cray is however left because otherwise **-s real64** is missing, which will fail in the coupling interface for a couple of REALs that are then assumed SP per default.


suvarchal:
wonderful! good didn't know -s real64 was needed for coupled setup but some how missed commit to `target_link_libraries(${PROJECT_NAME} PRIVATE OpenMP::OpenMP_Fortran)` as it seems we have been secretly using it in lumi builds after new cray compiler.

wdeconinck:
Would be great if "someone" 😇 could make work of creating tests for coupling interface

suvarchal:
> Would be great if "someone" 😇 could make work of creating tests for coupling interface

agree, none of us get enough time to do it elegantly, always in a rush.

PR #552, opened 2024-01-10
Support FESOM single precision fields

raguridan:


sebastianbeyer:
I think, I have already added this here: https://github.com/FESOM/fesom2/commit/5d73eb422756cddb971c65803fd3237ad99d5578
not sure why github is not showing this change...

sebastianbeyer:
As I wrote in my earlier comment, I think it's already in there, so this can be deleted.

PR #559, opened 2024-01-22
adding find oasis for oasis compiled with yac. also adding switch to …

JanStreffing:
…select it instead of regular find oasis


JanStreffing:
Need to test wether this works with oasis that does not have yac still. Might need more if switched at linker stage control.

JanStreffing:
Tests worked. This branch compiles with:
`DFESOM_COUPLED=OF`
with
`-DOIFS_COUPLED=ON -DFESOM_COUPLED=ON`
and with
`-DOIFS_COUPLED=ON -DFESOM_COUPLED=ON  -DOASIS_WITH_YAC=ON`

Imo ready to merge.

PR #561, opened 2024-01-24
Fix typo (inizialization/initialization) in gen_surface_forcing.F90

kinow:
Found troubleshooting the output of the model for eflows4hpc - https://github.com/eflows4hpc/workflow-registry/pull/11


koldunovn (approved):
Thanks, Bruno!

PR #563, opened 2024-01-28
1. in the coupled mode the freshwater flux components were scaled wit…

dsidoren:
…h rhofwt/rhowat every timestep (not every coupled step). This couls lead to discrepancy but not much expectedly.

2. redisual freshwater flux treatment from Jan Streffing in coupled mode with OpenIFS is now also applied when coupled to IFS

### DIFF
--- a/src/MOD_ICE.F90
+++ b/src/MOD_ICE.F90
@@ -49,6 +49,7 @@ TYPE T_ICE_THERMO
     !___________________________________________________________________________
     real(kind=WP) :: rhoair=1.3  , inv_rhoair=1./1.3  ! Air density & inverse ,  LY2004 !1.3 AOMIP
     real(kind=WP) :: rhowat=1025., inv_rhowat=1./1025.! Water density & inverse
+    real(kind=WP) :: rhofwt=1000., inv_rhofwt=1./1000.! Freshwter density & inverse
     real(kind=WP) :: rhoice=910. , inv_rhoice=1./910. ! Ice density & inverse, AOMIP
     real(kind=WP) :: rhosno=290. , inv_rhosno=1./290. ! Snow density & inverse, AOMIP
     ! Specific heat of air, ice, snow [J/(kg * K)]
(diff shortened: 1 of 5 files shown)


JanStreffing:
Severity of the bug depends on ration R of coupling timestep/model timestep like so:

E = 1-(1+(1000/1025)^R)/2

Error table:
grafik

So for low resolution setups, where R is typically between 1-3, the error was small. For higher resolutions setups, where the ration might be as high as 12 or 15, the error was more severe.

Fix approved from my side. Has to also be applied to refactoring!

JanStreffing:
Additional info. The scaling was introduced here about 2 years ago.
https://github.com/FESOM/fesom2/commit/c209e2395b9d838bee391fbe11888740ddc50a3b

@patrickscholz, Perhaps we can retrace the thought process to ensure everything is in order after this merge.

JanStreffing:
Here is the patch file that can fix this error on other branches. 
[fw_bug.patch](https://github.com/FESOM/fesom2/files/14362870/fw_bug.patch)

To apply, run:  git apply --ignore-whitespace fw_bug.patch

For me, working on 2.5.1 based on tag 2.5 the first change MOD_ICE.F90 did not work. I had to do that one manually and remove that section from the patch.

JanStreffing:
WIP branch of AWI-CM3 v3.2.1 with WIP branch of FESOM2.5.1 compared directly to AWI-CM3 v3.2 with FESOM Tag 2.5:

Changes are small (std) of tos is a bit worse. A lot of other things are a bit better. Likely very little of this is statistically significant (not checked by tool). Pardon the missing 3D temp and salt eval. I had only written yearly 3D output with is not enough for cmpitool.
FW-Fix-WIP-2 5 1

Here is the comparison of before and after performance biases to CMIP6 biases. Underlying are the same obs fields:
 
FW-Fix-WIP-2 5 1
TAG-2 5

@ackerlar

I would not suggest anyone throw away any simulation over this error. It's small. Just write an errata.

PR #566, opened 2024-02-10
sync threads before and after MPI exchanges 

suvarchal:
1. ensuring thread synchronization if MPI exchanges are done within OMP parrallel regions
2. clean up needless OMP barriers that are outside OMP Parallel regions

more stable, ran 3 months and a month (differerent exp) in levante  without a crash.

validity of results need to be checked.


suvarchal:
No, skip it

PR #570, opened 2024-03-06
2.1 awiesm 2.1 with transient tracers

mbutzin:
This branch includes the abiotic transient tracers 14C, 39Ar, CFC-12, and SF6.
They are enabled and controlled in namelist.config, section &runctl, where the main switch is "use_transit". 
Equilibrium / time slice simulations of 14C and 39Ar can be carried out without additional atmospheric input files. 
Transient simulations of 14C, CFC-12, and SF6 require additional additional input data which are available on request and could be copied to a central pool directory.
CFC-12 and SF6 are highly recommended for model validation/tuning because their initial concentrations are zero, their transient atmospheric input history = FESOM integration period is short (less than 100 years), and there are many field observations available. The implementation of similar transient tracers (such as CFC-11, CFC-113, or CCl4) would be straightforward.


JanStreffing:
As expected, this branch is quite far from refactoring. I wonder also if some of the things done here are intentional:

```
new file:   lib/metis-4.0
deleted:    setups/souf
deleted:    test/meshes
```

For some reason the PR tries to remove additions made on the refactoring branch. I'm not sure this is easy to fix with git. This will be manual work I'm afraid.

JanStreffing:
Most of that seems to come from https://github.com/FESOM/fesom2/pull/570/commits/fb6eab8aa3bbdf1c0201359bbc841fdafd8507ee

mbutzin:
I presume that > 95% of the conflicts are not due to the transient tracer code (which can be identified when searching for "use_transit"). How should I/we proceed?

JanStreffing:
I had a chat with Patrick today. We had two ideas of how we can proceed:

1. Perhaps the best way forward would be to take the changes on your branch that we want to keep (probably the last 60 commits) and rebase them onto refactoring.
2. Another potential way forward would be to first merge in the Iceberg from @ackerlar then the WISO @mwerner-awi, and then reimplement what missing based on refactoring at that stage. 

p.s. From after 2.6. we will have a different merge strategy, which will not see such long times between PR opening and being merged. I think we placed a bit too much emphasis on maintaining bit-reproducibility since 2.1.

JanStreffing:
May be superseeded by https://github.com/FESOM/fesom2/pull/572. TBD

JanStreffing:
Superseded by https://github.com/FESOM/fesom2/pull/576

PR #575, opened 2024-03-14
Refactoring bugfix: fix three small bugs in the tracer part

patrickscholz:
Fix 3 small bugs in the tracer part of FESOM2:

1.  [which_hnode_in_tr_z] which hnode is used in the computation of the vertical tracer gradient: Since the vertical tracer gradient is used explicit vertical Redi diffusivity we have to use here the layerthickness hnode of time step n, not hnode_new of time step n+1.
2.  [hnode_in_difveR] which hnode is used in the computation explicit vertical Redi diffusivity: since its the explicit component we have to use here the layerthickness hnode of time step n, not hnode_new of time step n+1.
3.  [which_tr_4_tr_xy] actual in the code we use the Adams-Bashfort tracer to compute the horizontal tracer gradient and the tracer at time step n to compute the vertical tracer gradient that is later used to compute the horizontal and vertical explicite Redi diffusivity. It doesnt makes sense here to have different tracers for the horizontal and vertical tracer gradient. Sergey recommended to use for the horizontal and vertical tracer gradient the tracer at time steps n. 

All three bugfixes have only minor changes on the outcome see below after one JRA55 cycle (temperature anomalies) !
grafik
grafik
grafik
grafik


JanStreffing (approved):

JanStreffing:
Here is the tripyview evaluation of a 1958 to 2017 standalone run with the tracer bugfixes vs the same period with the current main branch. I have nowhere to host the html, so as a pdf instead:
[fesom_26rc1_bugfix_hnode.pdf](https://github.com/user-attachments/files/16388635/fesom_26rc1_bugfix_hnode.pdf)

JanStreffing:
Differences are miniscule and numerical in nature. I will update the truth and merge.

PR #582, opened 2024-03-25
Draft: remove parms library

JanStreffing:
oce_ale needs work still. Some things from #ifdef parms seemingly are required outside of it.


JanStreffing:
Yeah no, we are not merging this one. This has to be redone based on refactoring. I had accidentally done it based on master.

PR #584, opened 2024-04-18
cmake cleanup

suvarchal:
cleanup of existing draft cmake.
- [x] tested standalone compile
- [x] tested awicm compile


JanStreffing:
I take it that the way to fix AWI-CM3 compile was to ensure that clean removes the cmake configure cache files and then add -fPIC, correct?

suvarchal:
> I take it that the way to fix AWI-CM3 compile was to ensure that clean removes the cmake configure cache files and then add -fPIC, correct?
yeah right. Let me know if it doesn't work.

PR #588, opened 2024-05-16
added computation of residual heal flux as proposed by Dmitry Sein.

dsidoren:


JanStreffing:
Please ensure actual physics tests e.g. 50 years constant 1990 with IFS/FESOM and OpenIFS/FESOM before merge. I can do the latter next week.

Issue #569, opened 2024-02-28
Remove parms library


JanStreffing:

sebastianbeyer:
The PR is against master, not refactoring, should we update that?

JanStreffing:
you are right. I see what happens when I fix that. might be easier to start over again from refactoring tough.

JanStreffing:
I closed https://github.com/FESOM/fesom2/pull/582. To be redone.

PR #597, opened 2024-06-10
Remove PARMS solver

sebastianbeyer:
Trying to address https://github.com/FESOM/fesom2/issues/569 here. 
I have removed the most obvious parts of PARMS, removed the C stuff from the CMakeLists.txt. Still need to do a proper test run, the CI does currently always fail with the other solver.  

- [ ] Test if this still gives good results
- [ ] Check the Make files that are not CMake. PARMS is still in there, but are they still in use?
- [ ] Check `src/oce_ale.F90` for unused variables
- [ ] Update CI tests with new solver results (if we are sure that the new solver works as intended, which it better does, because we ran nextgems and destine with it ;) )


sebastianbeyer:
From the CI it seems at least not terribly wrong
```
AssertionError: One or several tests have failed.
Variable: a_ice, current_value: 0.2692498167586123, master_value: 0.2692498167543513
FAIL!!!: a_ice DIFFERENCE: 4.261035968511351e-12.

Variable: salt, current_value: 23.944089812053637, master_value: 23.94408981205545
FAIL!!!: salt DIFFERENCE: -1.8118839761882555e-12.

Variable: sst, current_value: 8.526792796322084, master_value: 8.526792796340805
FAIL!!!: sst DIFFERENCE: -1.872102473043924e-11.

Variable: temp, current_value: 1.7018189804265962, master_value: 1.7018189804276316
FAIL!!!: temp DIFFERENCE: -1.035393992765421e-12.

Variable: u, current_value: -0.001431070160064853, master_value: -0.0014310701355284
FAIL!!!: u DIFFERENCE: -2.4536453164386574e-11.

Variable: v, current_value: 0.00014314235109323336, master_value: 0.0001431423767448
FAIL!!!: v DIFFERENCE: -2.5651566645541052e-11.
```

src/CMakeLists.txt
```diff
@@ -134,8 +134,6 @@ if(ENABLE_MULTIO)
    list(APPEND sources_Fortran ${src_home}/ifs_interface/iom.F90)
 endif()
 
-file(GLOB sources_C ${src_home}/*.c)
```
JanStreffing, line 137: Was this the only c code? I seem to recall some fesom fileserver code by Jan Hegewald that is in c. Though I think, we don't use it by default.

JanStreffing (commented):
Changes in oce_ale go beyond simple removal. I assume they are what was already used for DE and NG? If so, it looks good to me.

src/CMakeLists.txt
sebastianbeyer: Not sure what you mean about oce_ale. I removed the calls an logic surrounding the PARMS solver, but maybe I got carried away and removed too much? 
About the other C code, I will check again!

sebastianbeyer: The async_threads you mean? those are in CPP and in their own subdirectory.

sebastianbeyer:
I have discussed with Jan an we agreed that we will wait for the merge of production_DE into refactoring and then put this after that one.

sebastianbeyer:
why are there two more commits now in this? you have merged refactoring into this branch?

JanStreffing:
> why are there two more commits now in this? you have merged refactoring into this branch?

Yes, standard procedure. :)

sebastianbeyer:
I would prefer to rebase the feature branch on top of the latest develop (refactoring) branch to have a more readable history, but I am fine with merging if you prefer that :)

sebastianbeyer:
So for this we wait for the new test setup?

JanStreffing:
I consider this standard practise when there are multiple open pull requests. I would suggest not spending extra time on this. If you like, we can squash the history when merging, so that it appears as a single commit.

JanStreffing:
> So for this we wait for the new test setup?

We need Nikolay to be there to update the truth sequentially merge after merge for the remaining requests, as most of them will change the results.

JanStreffing (approved):

JanStreffing:
Note: Do not merge without updating truth.

JanStreffing:
Here is the tripyview evaluation of a 1958 to 2017 standalone run with parms removed vs the same period with the current main branch. I have nowhere to host the html, so as a pdf instead:
[fesom26rc_remove_PARMS.pdf](https://github.com/user-attachments/files/16387181/fesom26rc_remove_PARMS.pdf)

JanStreffing:
All changes appear to be within the expected range for purely numerical differences.

JanStreffing (approved):

PR #598, opened 2024-06-14
change some paths, remove ollie and mistral as HPCs are long gone

JanStreffing:


patrickscholz:
@JanStreffing: Should we set albedo or levante as default paths? At some point i was pretty sure i already did that at least onces for albedo

JanStreffing:
i think many more users are on levante compared to albedo.

patrickscholz (approved):

PR #603, opened 2024-06-18
fix landice fw bug

ackerlar:


src/ice_oce_coupling.F90
```diff
@@ -757,22 +757,6 @@ subroutine oce_fluxes(ice, dynamics, tracers, partit, mesh)
 
     !---wiso-code-end
 
-    !---fwf-code-begin
-    if(use_landice_water) then
-      do n=1, myDim_nod2D+eDim_nod2D
-         water_flux(n)=water_flux(n)-runoff_landice(n)*landice_season(month)
```
JanStreffing, line 763: This line is not needed anymore? I can't find it on the right?

ackerlar: Yes, this is correct. This loop was there twice, accidentally

JanStreffing (approved):

PR #86, opened 2021-03-25
Ice fct screening

dsidoren:

PR #612, opened 2024-07-10
Main ifs thermo

dsidoren:
change in the thermodynamics  proposed by Dmitry Sein which is required for more physical coupling with IFS type models. In the sea ice step we also neglect the divergent part since its contribution is negligible and sometimes caused errors on levante. The results might be not reproducible!


src/ice_setup_step.F90
```diff
@@ -248,13 +248,13 @@ subroutine ice_timestep(step, ice, partit, mesh)
 #endif
 #endif /* (__oifs) */
     if (flag_debug .and. mype==0)  print *, achar(27)//'[36m'//'     --> call ice_TG_rhs_div...'//achar(27)//'[0m'
-    call ice_TG_rhs_div    (ice, partit, mesh)
+    call ice_TG_rhs    (ice, partit, mesh)
```
dsidoren, line 251: this line will lead to the change of the test results

src/ice_setup_step.F90
```diff
@@ -248,13 +248,13 @@ subroutine ice_timestep(step, ice, partit, mesh)
...
 
     if (flag_debug .and. mype==0)  print *, achar(27)//'[36m'//'     --> call ice_fct_solve...'//achar(27)//'[0m'
     call ice_fct_solve     (ice, partit, mesh)
 
     if (flag_debug .and. mype==0)  print *, achar(27)//'[36m'//'     --> call ice_update_for_div...'//achar(27)//'[0m'
-    call ice_update_for_div(ice, partit, mesh)
+!   call ice_update_for_div(ice, partit, mesh)
```
dsidoren, line 257: this line will lead to the change of the test results

JanStreffing:
Here is the tripyview evaluation of a 1958 to 2017 standalone core2 run with these changes vs the same period with the current main branch. I have nowhere to host the html, so as a pdf instead:
[fesom26rc_ifs_thermo.pdf](https://github.com/user-attachments/files/16393949/fesom26rc_ifs_thermo.pdf)

JanStreffing:
Patterns look to be similar in magnitude and shape but inverted compared to https://github.com/FESOM/fesom2/pull/86

JanStreffing:
Besides the standalone evalulation in the previews comment I also did 50 years AWI-CM3 TCo95L91-CORE2 with constant year 2000 forcing. The results are good; cmpi-tool score is low:
main_ice_thermo
Comparison to before shows only minimal changes:
TCO95L91

I did not output the right MLD (MLD3). Deep ocean salt and temp make no sense to compare after 50 years.

JanStreffing (approved):

PR #613, opened 2024-07-15
Refactoring improve4ogurses calendar new

patrickscholz:
- redo avoid leapyear problem of ozgures 
- Solve special problem here: Ozgurs wants to repeat only a part of the JRA55 forcing periodically (e.g repeat the period 1958-1967) through re-linking of the original forcing files.

    1st problem: This will mess up the leap year cycle of the entire forcing, so the model has to run best without leapyear although the calendar system of the forcing remains gregorian (so with leapyear) This can be solved by makeing the treatment of the time-axes depending on the calendar of the forcing files, not alone from the include_fleapyear flag

    2nd problem: In the case of a periodically repeated forcing period through re-linking of the original files it happens that e.g. the year 1968 is linked to the file 1958. In that case the time axis of the 1968 forcing file remains that of the original 1958 file. So the time axis needs to be resetted properly to the right year they are supposed to represent

-  did this to ensure  their isnt a bug in it, and results remain identical to the default test case 

- year 1958 (non-leapyear)
image

- year 1960 (1st leapyear, flpyrF is with out 29. Feb 1960)
image

### DIFF
--- a/src/gen_modules_clock.F90
+++ b/src/gen_modules_clock.F90
@@ -196,11 +196,18 @@ contains
     flag=0
 
     if(.not.include_fleapyear) return
+    call is_fleapyr(year, flag)
+  end subroutine check_fleapyr
 
+  subroutine is_fleapyr(year, flag)
+    implicit none
+    integer, intent(in) :: year      
+    integer, intent(out):: flag
+    flag=0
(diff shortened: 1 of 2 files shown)


patrickscholz:
from my point of view this can be merged i recheck everything against a clean refactoring branch, and everything seemed to work now !

JanStreffing (approved):

PR #615, opened 2024-07-29
Fixed "refactoring" is not compiling on Levante (in CPU). 

basava70:
refactoring is not compiling on Levante, either on shell.inter or shell.gnu.
blas not found error.
Fixed it by commenting export commands in env.sh.
Also added option to choose which compiler to use on Levante. It is set to shell.gnu. I copied Suvi's code to add compID regex to give arguments like " ./configure levante.intel ",  " ./configure levante.gnu", " ./configure levante.nvhpc".


JanStreffing:
Please note that refactoring branch shall not be used anymore as the target of commits. See my email from 21.6.2024

> If you are not a FESOM model user, please ignore this email. 
> 
> If you are, you might know, that the default FESOM2 master branch had been falling behind the refactoring branch with no merger of refactoring into master for about two years. Thus refactoring had become an unofficial de-facto default branch. Today we created an official new default branch: main. All Merge Requests with the base refactoring were rebased onto main. If you develop on the FESOM2 code, from here on out please branch off from and merge back into main. Thank you!

Do you wish to merge these changes back, then they shall be merged into the main branch.

mandresm:
Please, do not close @basava70's PRs, they might need to be merged into other branches, in that case we need to edit the PR and target another branch, but the PR should still be available so that we can discuss the merging.

mandresm:
> Do you wish to merge these changes back, then they shall be merged into the main branch. In that case rebase to main please.

Is `main` used in DestinE? Because it's great that @basava70 changes make it to main but it's a requirement that they make it to the branch used in DestinE

JanStreffing:
This I don't know. It might be production_DE, which I overzealously closed, sorry for that. I don't think it's refactoring, which was the misnamed general default branch before we made main. If it was refactoring, then it would now be main.

sebastianbeyer:
DestinE is currently using the tag `2.5.0.6_de`, but will be using 'regular' tags from main soon. So, if this is still required, please target main

mandresm:
@basava70, what work is missing in this branch?

JanStreffing:
Are we sure the latest main branch is not compiling on levante without these changes?

basava70:
Ok, the code in the main branch does pass the test and compiles on Levante without any changes. However, it compiles only using shell.gnu script and doesn't take shell.intel (or shell.nvhpc for gpu computation). To enable this, we need to add code in env.sh as given in this PR.
PS. This code is provided by @suvarchal long back and I cannot find which commit that was but I copied it by hand in this PR.

env.sh
```diff
@@ -112,4 +117,4 @@ else
    echo "Sourcing $(realpath $SHELLFILE) for environment" 
    source $SHELLFILE
    echo "$(realpath ${SHELLFILE})" > $DIR/bin/current_shell_path
-fi
+fi
```
mandresm, line 120:
```suggestion
fi

```
Just to minimize the number of unnecessary changes, so that the git history and the blame function gives more clear output.

mandresm: you can commit this suggestion by clicking on the `Commit suggestion` button above

JanStreffing (approved):

PR #619, opened 2024-08-16
get 2.5.0.6_de back into main

sebastianbeyer:
This is what we are currently using in DestinE. The branch name (updates_2.5.0.4_de) is different from the tag name (2.5.0.6_de), sorry for that! 
This is a draft and I need to clean up some stuff and some of the changes we need to discuss first, I think.


src/MOD_DYN.F90
```diff
@@ -18,13 +18,13 @@ MODULE MOD_DYN
     integer       :: lutype  = 2
     real(kind=WP) :: droptol = 1.e-8
 !!! PARMS Solver
-    real(kind=WP) :: soltol  = 1e-10  ! default for PARMS
-    logical       :: use_parms = .TRUE.
-!!!
-!!! Sergey's Solver
-!   real(kind=WP)  :: soltol  = 1e-5  ! default for PARMS
-!   logical        :: use_parms = .FALSE.
+!    real(kind=WP) :: soltol  = 1e-10  ! default for PARMS
```
JanStreffing, line 21: This can be changed, PARMS is gone.

JanStreffing (changes requested):

src/MOD_DYN.F90
JanStreffing: The rest looks good, mostly multio and some OMP changes.

JanStreffing (approved):

src/MOD_DYN.F90
sebastianbeyer: yes, I will try to remove the unnecessary changes (also some commented out stuff from Suvi), but not exactly sure how to do it without breaking git ;) 

The thing I was wondering about is the additional MPI barriers, I am not sure if they are needed, if they harm performance etc.

sebastianbeyer:
I was surprised to see this part:
```
!SUVI: disable overflow, underflow for entire model when used  in coupled with ifs
!      bad practice
#if defined  __ifsinterface
    call ieee_set_halting_mode(ieee_overflow, .false.)
    call ieee_set_halting_mode(ieee_underflow, .false.)
    call ieee_set_halting_mode(ieee_invalid, .false.)
#endif
```
We probably do not want the in the code? or is that required for our runs @suvarchal ?? would be kind of bad xD

sebastianbeyer:
Ah, I saw in a later commit it is commented out. So I will leave out that commit.

PR #515, opened 2023-10-11
The great cmake refactoring

wdeconinck:
@sebastianbeyer @suvarchal @dsidoren

Here is the promised branch with CMake refactoring.
A few things this now enables

- Use MPI as library without using compiler wrappers
- Improved NetCDF detection from various netcdf modules on various HPC (LUMI, ECMWF-ATOS)
- Installation of fesom as a cmake package that other projects like e.g. IFS could do `find_package(fesom)`
- Building fesom with shared libraries (default now)
- Only recompile the version-info file when the sha1 has changed, avoiding a relinking cascade.

Rename BUILD_FESOM_AS_LIBRARY to ENABLE_IFS_INTERFACE. Fesom can always build as library and have a driver executable for the library. Remaining TODO is however a run-time configuration switch to use the IFS-interface or internally provided forcings from file, for this to be working as a single binary.

After this is merged in, it will be easy to merge the [multio-climate-dt branch](https://github.com/wdeconinck/fesom2/tree/nextGEMS_c3_fesom2.5_climateD_multio) which I have refactored with similar changes before. The changes on this branch make multio completely optional, so it can be merged in the main refactoring branch.

PR #621, opened 2024-08-23
The great cmake refactoring MK2

JanStreffing:
Copy PR of #515, but from https://github.com/FESOM/fesom2/ remote, so that we can work on it.


sebastianbeyer:
I still see some parms stuff in here

PR #622, opened 2024-08-27
Turn on netcdf compression by default

JanStreffing:


patrickscholz (approved):

pgierz (approved):
Yes, default compression is a good thing. One small caveat that might pop up in automatic tests: checksums of compressed files will be different, so if anything runs correctness tests in that manner, the test might break.

JanStreffing:
It appears that's not the case. The tests passed.

PR #627, opened 2024-09-16
update mpif.h to use mpi

sebastianbeyer:
Locally compiles, let's see what the tests say. Also, do we test compiling with e.g. `ifs_interface` enabled? Otherwise we might be missing something ;) 
I will try `use mpi_f08` as well if this works and also try to build on levante and lumi...

### DIFF
--- a/src/MOD_PARTIT.F90
+++ b/src/MOD_PARTIT.F90
@@ -5,12 +5,12 @@ USE O_PARAM
 USE, intrinsic :: ISO_FORTRAN_ENV, only : int32
 USE MOD_WRITE_BINARY_ARRAYS
 USE MOD_READ_BINARY_ARRAYS
+USE mpi
 #if defined(_OPENMP)
     USE OMP_LIB
 #endif
 IMPLICIT NONE
 SAVE
-include 'mpif.h'
 integer, parameter   :: MAX_LAENDERECK=16
 integer, parameter   :: MAX_NEIGHBOR_PARTITIONS=32
 
--- a/src/fortran_utils.F90
+++ b/src/fortran_utils.F90
@@ -1,5 +1,6 @@
  ! synopsis: basic Fortran utilities, no MPI, dependencies only to INTRINSIC modules
 module fortran_utils
+  use mpi
   implicit none
 
 contains
@@ -48,7 +49,6 @@ contains
     integer mype
     integer npes
     integer mpierr
-    include 'mpif.h'
 
     call MPI_Comm_Rank(mpicomm, mype, mpierr)
     call MPI_Comm_Size(mpicomm, npes, mpierr)
--- a/src/gen_modules_partitioning.F90
+++ b/src/gen_modules_partitioning.F90
@@ -113,7 +113,7 @@ subroutine par_ex(COMM, mype, abort)       ! finalizes MPI
 #ifndef __oasis
   if (present(abort)) then
      if (mype==0) write(*,*) 'Run finished unexpectedly!'
-     call MPI_ABORT(MPI_COMM_WORLD, 1 )
+     call MPI_ABORT(MPI_COMM_WORLD, 1, error)
   else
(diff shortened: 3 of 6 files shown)


sebastianbeyer:
I could build and run on levante and I also noticed that `mpi_topology_module.F90` already contains `use mpi` in one place ;)
`use mpi_f08` takes more effort (for some reason the general mpi routines can not be found, but if I call everything with `_f08` postfix it does find it...) so I would suggest to merge this now and look at `f08` later again.

JanStreffing (approved):

JanStreffing:
Standalone looks good. I assume you tested this with IFS interface by now? Our test suit checks if FESOM builds as a library, but obviously it can't check if it runs as a library in IFS.

sebastianbeyer:
I am testing with IFS now, but it takes a little, because we are behind main quite a bit and now I need to add all of the namelist groups that did not exist before to the config on the HPCs...

sebastianbeyer:
okay, now I tested on levante with IFS-FESOM and it still works. So from my point of view this can be merged

sebastianbeyer:
You need to fix the call, probably you need to use the correct type for MPI_COMM_WORLD

sebastianbeyer:
ah, not it's not that... 
looks like this file is not always compiled. Seems like it does call MPI subroutines, but there is no `use mpi` nor `include mpif.h` anywhere. strange.

sebastianbeyer:
Okay, it looks like the file `cpl_driver.F90` does not contain either `use mpi` nor `include mpif.h`. This was never correct, but before it did work, because when that file included another module which did have `include mpif.h` it would also get a copy of the subroutines. So the solution is to rather add `use mpi` to the respective modules in `cpl_driver.F90`.

sebastianbeyer:
and also update the CI, so that this file is also built during in the CI builds.

ackerlar:
Unfortunately, I still get the same error during compilation

sebastianbeyer:
Then please give me a way to replicate thatAm 09.10.2024 um 13:56 schrieb Lars Ackermann ***@***.***>:﻿
Unfortunately, I still get the same error during compilation

sebastianbeyer:
But that requires access to  https://git.geomar.de/foci/src/oasis3mct.git  if I understand correctly. And I don't have that...

JanStreffing:
That doesn't sound right, awi-esm2 should not use geomar oasis... should be from https://gitlab.dkrz.de/modular_esm/oasis3-mct i think

ackerlar:
Apparently, oasis and echam come from the AWI gitlab:

https://gitlab.awi.de/paleodyn/Models/oasis3-mct.git
https://gitlab.awi.de/paleodyn/Models/echam6.git

sebastianbeyer:
ah, I did not see the other possible repositories. 

Okay, but this is too tedious, sorry

JanStreffing:
I downloaded the source codes, you can copy from here /work/ab0246/a270092/model_codes/awiesm-2.1-paleodyn-2.5 and do esm_master comp-awiesm-2.1-paleodyn-2.5 in the copy you copy it to.

mandresm:
You should have access now @sebastianbeyer

sebastianbeyer:
Okay, the line that you show: `CALL MPI_Comm_Rank ( MPI_COMM_WORLD, commRank, ierror )` 
has `commRank` and that is a `logical`! but `MPI_COMM_RANK` expects an `integer` here! that is the reason to do `use mpi` instead of `include mpif.h`! Type safety! 
What is the purpose of `commRank`? what should be its type??

ackerlar:
It seems to me that `commRank` is only used for few write statements. I changed its type and adapted the if statements (I don't know who included this part but I guess it should read `ioerrer .eq. oasis_ok`) and now the models compiles. I a doing a quick test run and push the changes. Thanks @sebastianbeyer

PR #630, opened 2024-09-22
delete unused files

sebastianbeyer:
I collected a few files where I thought they are not needed anymore. Please check if they are really not in use!


sebastianbeyer:
what about `test_pgierz` @pgierz ? should that be moved to tests or is it not in use? (I saw in some of the logs from the CI that we run some test on ugrid compliance, I think...

pgierz:
Sorry for the late reply. If there is a unified place for test, that should go there. If I remember correctly, a ugrid compliance test is run automatically, I'll check.

pgierz:
https://github.com/FESOM/fesom2/blob/e900ff016b807654abcd91e46f24f9b39f6eceae/.github/workflows/fesom2.1.yml#L72-L77

I'll move the text files in my test folder into the main test folder, they might be handy to keep around for reference.

sebastianbeyer:
okay, so this is it for now I guess?

JanStreffing:
OK from my side, but I would wait for Patrick. He knows better what is used / not used by FESOM core developers.

JanStreffing (approved):

patrickscholz (approved):

PR #635, opened 2024-10-08
Update mpi again

sebastianbeyer:
Same as last time, but now I added `use mpi` to `src/cpl_driver.F90` as well, this should fix the error that @ackerlar reported here https://github.com/FESOM/fesom2/pull/627#issuecomment-2399957341 . However, I can not test it, because I can not build with oasis and/or oifs (no access to oasis and if I try to just define `__oifs` I do get 
```
/Users/sebeye001/climateDT/fesom2/src/io_restart.F90:180:90:

  180 |   call ice_files%def_node_var_optional('ice_albedo', 'ice albedo',    '-',   ice%atmcoupl%ice_alb, mesh, partit)
      |                                                                                          1
Error: 'atmcoupl' at (1) is not a member of the 't_ice' structure
```
)

### DIFF
--- a/src/MOD_PARTIT.F90
+++ b/src/MOD_PARTIT.F90
@@ -5,12 +5,12 @@ USE O_PARAM
 USE, intrinsic :: ISO_FORTRAN_ENV, only : int32
 USE MOD_WRITE_BINARY_ARRAYS
 USE MOD_READ_BINARY_ARRAYS
+USE mpi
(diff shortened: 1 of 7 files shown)


JanStreffing:
@ackerlar Can you commit the changes you had to make to get this running with oasis to this branch?

ackerlar (approved):

PR #637, opened 2024-10-18
Workbench mainf26 icepack

patrickscholz:
- fix freshwater balancing with icepack thus making icepack volume conserving with ALE zstar
- fix icepack restart for netcdf restarts and raw restarts, make sure that restart are also volume conserving
- fix a couple of bugs in the driver interface of icepack 
- fix two bugs in icepack itself see: https://github.com/lzampier/Icepack/tree/icepack_fesom2_bugfix


JanStreffing (approved):
A couple of debug write statements are left with !PS. However, since @fkauker might still work more on this, I think it's ok to merge them in, commented out.

PR #642, opened 2024-10-26
Improve error handling for open files for clock file and mesh file

sebastianbeyer:
Make a proper error message with the path where the file was expected and do proper mpi_abort.

### DIFF
--- a/src/gen_modules_clock.F90
+++ b/src/gen_modules_clock.F90
@@ -2,6 +2,8 @@ module g_clock
   !combining RT and Lars version
   !
   use g_config
+  use iso_fortran_env, only: error_unit
+  use mpi
   implicit none
   save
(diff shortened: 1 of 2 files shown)


sebastianbeyer:
What do you say, @JanStreffing ? Is it enough to check the `rpart.out` file, or should we add checks to every file (It would be nice to check everything, but it means a lot of additional code)?

trackow:
I think checking for the presence of the rpart.out file should cover a majority of issues in the IFS-FESOM context when running on different machines ;) Maybe just checking for the presence of the right distXXX folder could already be enough.

Also great to see that you check for the presence of the clock file.

sebastianbeyer:
Hey, can we add this, @JanStreffing ?

patrickscholz (approved):

PR #644, opened 2024-10-29
Double Gyre testing

alexwilms:
Added Double Gyre setup


JanStreffing:
Hey @alexwilms. Its seems something did not work out with the merge request. It would add more than 30000 lines mostly in the form of two libparms.so library binaries. These should not be merged into the main, because we do not want to ship binaries and they would greatly increase the repo size. Please have a look at `Files changed` up top. I would recommend making a new branch from main, and creating a new version of your double gyre additions.

PR #649, opened 2024-11-15
Fix/wave erosion

ackerlar:
Change wave erosion parameterization in iceberg module so that wave erosion only occurs if the upper-most ocean layer can uptake heat fluxes from melting of eroded ice volume


ackerlar:
Can we merge this branch?

JanStreffing (approved):

PR #654, opened 2024-12-03
Double gyre setup

alexwilms:
Implemented double gyre setup with nemo-namelists, jobscript and env for nesh.

### DIFF
--- a/config/namelist.config.toy_nemo
+++ b/config/namelist.config.toy_nemo
@@ -1,0 +1,79 @@
+! This is the namelist file for model general configuration
+
+&modelname
+runid='fesom'
+/
+
+&timestep
+step_per_day=45 !96 !96 !72 !72 !45 !72 !96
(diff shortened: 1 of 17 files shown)


patrickscholz:
should we really name it "nemo" in the code or better something more saying like "dbgyre" or something like this ... . And did you test this on more than one maschine ?

alexwilms:
> should we really name it "nemo" in the code or better something more saying like "dbgyre" or something like this ... . And did you test this on more than one maschine ?

Sorry for the late answer, I will rename it to dbgyre or just gyre and we are currently testing it on the AWI machine. I will let you know when this is done.

alexwilms:
@patrickscholz I changed nemo to dbgyre for the setup specific parts and we tested it on levante at dkrz aswell

patrickscholz (approved):

PR #660, opened 2024-12-19
added heat content computation as required by destinE GSV. Shall be a…

dsidoren:
…ctivated in namelist.io as ldiag_destinE=.TRUE.

### DIFF
--- a/config/namelist.io
+++ b/config/namelist.io
@@ -9,6 +9,7 @@ ldiag_dMOC       =.false.
 ldiag_DVD        =.false.
 ldiag_forc       =.false.
 ldiag_extflds    =.false.
+ldiag_destinE    =.false.
 ldiag_trflx      =.false.
 ldiag_uvw_sqr    =.false.
 ldiag_trgrd_xyz  =.false.
--- a/src/io_meandata.F90
+++ b/src/io_meandata.F90
@@ -80,7 +80,7 @@ module io_MEANDATA
 !
   integer, save                  :: io_listsize   =0
   logical, save                  :: vec_autorotate=.FALSE.
-  logical, save                  :: lnextGEMS=.FALSE.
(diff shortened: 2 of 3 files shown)


config/namelist.io
```diff
@@ -9,6 +9,7 @@ ldiag_dMOC       =.false.
 ldiag_DVD        =.false.
 ldiag_forc       =.false.
 ldiag_extflds    =.false.
+ldiag_destinE    =.false.
```
JanStreffing, line 12: Would be great to use a name that tells the user what diagnostic will be created.

src/io_meandata.F90
```diff
@@ -80,7 +80,7 @@ module io_MEANDATA
 !
   integer, save                  :: io_listsize   =0
   logical, save                  :: vec_autorotate=.FALSE.
-  logical, save                  :: lnextGEMS=.FALSE.
```
JanStreffing, line 83: Same for this one.

JanStreffing (changes requested):

src/io_meandata.F90
sebastianbeyer: when I try to run this with `ldiag_destinE=.TRUE.`, I get the following:

```
  --> FESOM STARTS TIME LOOP
 Updating SSS restoring data for month                1
  --check opt_visc--> Mean Ratio Resol/Rrossby =    5.0000000000000000
 Subroutine exchange_nod3D not implemented for           3 layers.
 Adding the MPI datatypes is easy, see oce_modules.F90.
 Run finished unexpectedly!
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 2 in communicator MPI_COMM_WORLD
  Proc: [[10799,1],2]
  Errorcode: 1
```

sebastianbeyer:
replaced by #673

PR #661, opened 2024-12-23
Abort model cleanly in ifsinterface and standalone on errors

suvarchal:
Some of old features lost in my branches that didn't make it into main, part 1.

- clean and exit model with par_ex instead of abort, this avoids a few annoying model hanging in ifs_interface.
- checks for namelists exist and show line where error happens and abort cleanly. 

TODO in another PR:
- model stop with par_ex should also be used consistently in checks should also be used in cvmix routines and ib routines. 
- add namelist checks in all other modules in style similar to this PR

### DIFF
--- a/src/gen_modules_partitioning.F90
+++ b/src/gen_modules_partitioning.F90
@@ -107,25 +107,29 @@ subroutine par_ex(COMM, mype, abort)       ! finalizes MPI
   integer, optional, intent(in)   :: abort
   integer                         :: error
 
-! For standalone runs we directly call the MPI_barrier and MPI_finalize
+ 
+! For standalone runs we
+! when there is error par_ex should be called with abort argument to abort abruptly,
+! in all other cases model will be finalized here, call the MPI_barrier and MPI_finalize
 !---------------------------------------------------------------
-!TODO: logic is convoluted here, not defined oasis and model needs to abort doesn't happen using par_ex 
 #ifndef __oasis
   if (present(abort)) then
      if (mype==0) write(*,*) 'Run finished unexpectedly!'
-     call MPI_ABORT(MPI_COMM_WORLD, 1, error)
+     call MPI_ABORT(COMM, 1, error)
   else
-          ! TODO: this is where fesom standalone, ifsinterface etc get to 
-          !1. there no abort actually even when model calls abort, and barrier may hang
-          !2. when using fesom as lib using finalize is bad here as there may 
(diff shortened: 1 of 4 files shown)


src/gen_modules_partitioning.F90
```diff
@@ -107,25 +107,29 @@ subroutine par_ex(COMM, mype, abort)       ! finalizes MPI
...
+          !3. or change all the cases in fesom that call par_ex to properly use abort in all cases other then normal closure.
      call  MPI_Barrier(COMM, error)
      call  MPI_Finalize(error)
   endif
 
-#else !  standalone
-! TODO logic below is also convoluted really not really for standalone
+#else ! 
+! TODO logic below is convoluted, COMM that is passed should be used for MPI_ABORT
+! changes are easy but need to be tested with coupled configurations
```
JanStreffing, line 132: Does that mean changes are were easy to make or would be easy?

src/oce_mesh.F90
```diff
@@ -244,7 +244,7 @@ SUBROUTINE read_mesh(partit, mesh)
        write (unit=error_unit, fmt='(3A)') &
          '### error: can not open file ', file_name, &
          ', error: ' // trim(errmsg)
-       call MPI_Abort(MPI_COMM_FESOM, 1, ierror)
+ !call MPI_Abort(MPI_COMM_FESOM, 1, ierror) # dont call abort here yet called bellow
```
JanStreffing, line 247: Keep or remove fully

JanStreffing (changes requested):
Excellent changes. This should make missing namelists etc. easier to catch. I have just a few small changes requested.

src/gen_modules_partitioning.F90
suvarchal: It handles core parts and there are other parts like ice modules, blowup that may need more inspection on aborts, clean kind to stop normally vs really abort kind. this can be easy to do once we go through those parts, but i would leave it here for now, going into all those parts to me is another issue.  @JanStreffing  thanks for confirming communicator in oifs-fesom case, then nothing about that configuration to worry about with these changes.

src/oce_mesh.F90
suvarchal: removed

src/gen_modules_partitioning.F90
JanStreffing: i want to do one small test with OpenIFS-FESOM2 before I merge.

JanStreffing (approved):

PR #662, opened 2025-01-10
apply patch with MB changes

ackerlar:
transient tracers fixed with @mbutzin changes but based on latest 'main' branch


JanStreffing:
Looks like it has the same issue. The problem is already in the commits.

PR #664, opened 2025-01-12
Small fixes for iceberg module, that allow compiling with intel debug flags

koldunovn:
@ackerlar can you, please have a look if this makes sense and if I did not break anything? :)

### DIFF
--- a/src/icb_allocate.F90
+++ b/src/icb_allocate.F90
@@ -1,3 +1,20 @@
+module allocate_icb_interface
+  interface
+     subroutine allocate_icb(partit, mesh)
+        use iceberg_params
+        use g_config
+        use g_comm
+        use g_comm_auto
+        use o_param
+        use MOD_PARTIT
+        use MOD_MESH
+        implicit none
+        type(t_partit), intent(inout), target :: partit
+        type(t_mesh),   intent(in)   , target :: mesh
+     end subroutine allocate_icb
+  end interface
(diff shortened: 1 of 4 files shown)


src/icb_dyn.F90
```diff
@@ -1167,7 +1168,7 @@ subroutine iceberg_avvelo(mesh, partit, dynamics, uo_dz,vo_dz,depth_ib,iceberg_e
 	
     if(dz < 1) then
       !write(*,*) 'z coord of up node', n_up, ':', coord_nod3D(3, n_up), 'z coord of low node', n_low, ':', coord_nod3D(3, n_low)
-      call par_ex
+      call par_ex (partit%MPI_COMM_FESOM, partit%mype)
```
suvarchal, line 1171: yes this is what it should be.

src/icb_elem.F90
suvarchal, line 1: yes yes

src/icb_allocate.F90
suvarchal, line 1: not sure about this,  shouldn't the subroutine be inside the module, and use this module when we use alllocate_icb. anyways module it is not being used anywhere in this PR. so is redundant.

suvarchal (approved):

src/icb_allocate.F90
suvarchal: but lets fix it another day and merge.

PR #665, opened 2025-01-12
Add a bunch of interfaces to make intel happy

koldunovn:
That's a relatively big one, but should not be too compicated. It mostly adding the interfaces, that Intel compiler was complaining about, and did not let me compile with debug flags.

This PR depends on #664 - as for icebergs I need to do more than just adding interfaces and don't want to break anything for @ackerlar 

Would be also great if things work for recom as well (@ogurses )

### DIFF
--- a/src/MOD_ICE.F90
+++ b/src/MOD_ICE.F90
@@ -788,6 +788,23 @@ end subroutine ice_init
 !
 !_______________________________________________________________________________
 ! initialise derived type for sea ice
+
+module ice_init_toyocean_dummy_interface
+    interface
+       subroutine ice_init_toyocean_dummy(ice, partit, mesh)
+          USE MOD_ICE
+          USE MOD_PARTIT
+          USE MOD_PARSUP
+          USE MOD_MESH
+          USE o_param, only: WP
+          implicit none
+          type(t_ice)   , intent(inout), target :: ice
+          type(t_partit), intent(inout), target :: partit
+          type(t_mesh)  , intent(in)   , target :: mesh
+       end subroutine ice_init_toyocean_dummy
+    end interface
+ end module ice_init_toyocean_dummy_interface
+
 subroutine ice_init_toyocean_dummy(ice, partit, mesh)
     USE MOD_ICE
     USE MOD_PARTIT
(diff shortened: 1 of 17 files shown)


koldunovn:
> @koldunovn I checked the changes made in #664 and it runs fine with interactive icebergs

Thanks a lot!

ackerlar:
I get this error:

'''
   6219 681:   --> subroutine vert_vel_ale --> found Nan in Wvel after cumulativ summation(...)
   6220 681:   mype =         105
   6221 681:   node =           1
   6222 681:   Wvel(nzmin:nzmax, n)=                     NaN                     NaN                     NaN                     NaN                     NaN                     NaN                     NaN
                        NaN
'''

and the model crashes: '/work/ab0246/a270124/esm-experiments/awiesm-2.6/test03/log/test03_awiesm_compute_13500101-13501231_14756612.log'

patrickscholz:
Does this happen in the first step of the model? Since it happen after the cumulativ sumation of Wvel means that your UV from which you compute the divergences already have some NaN. This could mean that there are already issues with computing your deta and eta --> which could link to the water_flux. Any idea how the water_flux looks like ?

ackerlar:
Yes, it happens in the first time step. Is there any way to check `water_flux` after the crash? I run AWI-ESM (i.e. with ECHAM) and icebergs + cavities. The same setup I used for testing [https://github.com/FESOM/fesom2/pull/664](https://github.com/FESOM/fesom2/pull/664) which worked fine for me

patrickscholz:
In optimal case it should have created a blowup file fesom.blowup.nc ...

ackerlar:
There is no blowup file

patrickscholz:
You have to check where the model exactly quits because the message you posted is just a warning the model doesnt stop there.  If the model comes to the point at the end of the ocean step where it checks for the blowups and finds it. It should be able to write a blowupfile where the most important variables at the moment of blowup are in.

koldunovn:
Please don't check this at the moment - I totally broke something :)

JanStreffing:
> Please don't check this at the moment - I totally broke something :)

Did @suvarchal redo this? https://github.com/FESOM/fesom2/pull/736, is this still needed @koldunovn

PR #668, opened 2025-01-14
updated Juwels env file

vasmue:
the env file for Juwels was outated since the last update on their system. Here is the content for the updated file:

```
#!/usr/bin/bash
# ENVIRONMENT used in new_test15_compute_20000101-20000101.run
# Use this file to source the environment in your
# preprocessing or postprocessing scripts

module --force purge
module use $OTHERSTAGES
module --force purge
module load Stages/2023
module load GCC/11.3.0
module load OpenMPI/4.1.4
module load CMake/3.26.3
module load Python/3.10.4
module load imkl/2022.1.0
module load Perl/5.34.1
module load git
module load CDO
module unload ecCodes
module load libaec FFTW imkl cURL netCDF netCDF-Fortran
module list
```

### DIFF
--- a/env/juwels/shell
+++ b/env/juwels/shell
@@ -1,44 +1,20 @@
-##########
-module --force purge 
-module use /gpfs/software/juwels/otherstages   
-module load Stages/2020
-module load Intel/2020.2.254-GCC-9.3.0
-module load ParaStationMPI/5.4.7-1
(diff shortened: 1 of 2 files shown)


suvarchal (approved):
all seems to work on juwels after this.

PR #671, opened 2025-01-23
Remove comma in write statements

sebastianbeyer:
Against standard and does not compile with cray otherwise


suvarchal (approved):
good catch

PR #672, opened 2025-01-23
Add missing close statement

sebastianbeyer:
add missing close(fileunit). 
improves sbc_ini a little.


suvarchal (approved):
We should do it probably everywhere else too with action=read

sebastianbeyer:
Yes!!! That will also ease stress for the file system; and might allow more optimization, because we don’t need a file lock!

PR #673, opened 2025-01-23
Add DestinE phase2 heatcontent variable

sebastianbeyer:
this is a replacement for the other heat content PR.

This requires to set ldiag_destinE=.TRUE. in namelist.io then, the following variables can be written:

hc300m, heatcontent300m, Vertically integrated heat content upper 300m,   J m**-2
hc700m, heatcontent700m, Vertically integrated heat content upper 700m,   J m**-2
hc,     heatcontent,     Vertically integrated heat content total column, J m**-2

### DIFF
--- a/config/namelist.io
+++ b/config/namelist.io
@@ -9,6 +9,7 @@ ldiag_dMOC       =.false.
 ldiag_DVD        =.false.
 ldiag_forc       =.false.
 ldiag_extflds    =.false.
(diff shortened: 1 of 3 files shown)


JanStreffing:
I can imagine this might be of quite some interested for non DestinE simulations as well. Are the DestinE diagnostics expensive? Is the heat content diag expensive?

sebastianbeyer:
I don’t know. @dsidoren probably does!

PR #674, opened 2025-01-24
Mark issues and PRs as stale after two years of inactivity.

JanStreffing:
Closes issues 30 days later if no further update is made. Check is made daily.

Copied from: https://github.com/esm-tools/esm_tools/blob/JanStreffing-patch-1/.github/workflows/stale.yml


sebastianbeyer:
I like that!

sebastianbeyer (approved):

koldunovn (approved):

pgierz:
There is a critical mistake here, this will never run: the file needs to end with `.yml` for it to be picked up by GitHub actions

PR #680, opened 2025-01-29
Ke energy conversion

koldunovn:


dsidoren (approved):

src/io_meandata.F90
```diff
@@ -1343,45 +1343,43 @@ subroutine ini_mean_io(ice, dynamics, tracers, partit, mesh)
     if (dynamics%ldiag_ke) then
        call def_stream((/nl-1, elem2D/), (/nl-1, myDim_elem2D/), 'ke_adv_u_xVEL', 'work of advection   [u]',        'm2/s2', dynamics%ke_adv_xVEL(1,:,:),  io_list(i)%freq, 'y', 8, partit, mesh)
        call def_stream((/nl-1, elem2D/), (/nl-1, myDim_elem2D/), 'ke_adv_v_xVEL', 'work of advection   [v]',        'm2/s2', dynamics%ke_adv_xVEL(2,:,:),  io_list(i)%freq, 'y', 8, partit, mesh)
+       !call def_stream((/nl-1, elem2D/), (/nl-1, myDim_elem2D/), 'ke_cor_u_xVEL', 'work Coriolis  :)  [u]',         'm2/s2', dynamics%ke_cor_xVEL(1,:,:),  io_list(i)%freq, 'm', 8, partit, mesh)
```
JanStreffing, line 1346: Can these comment out lines be removed? Or is this just to turn a particular output off? If the latter, I would suggest default off, to be turned on via namelist.io. If the former, then please remove.

JanStreffing (changes requested):

src/io_meandata.F90
JanStreffing: @koldunovn, can you look at this one more time and ensure that all KE energy output that is still needed can be activated by the diag flag, and all output that is not needed (anymore) is removed, not commented out?

JanStreffing: It would be good to bundle behind ldiag_ke, and place that switch into namelist.io

PR #682, opened 2025-02-03
added fix for old meshes

christian-stepanek:
for old method meshes that are not read in the new fesom due to contour computation - relevant code parts are encapsulated with new switch compute_oasis_corners


pgierz:
Hi @christian-stepanek,

The example namelists still need to be updated, so far this is only directly in the FESOM source code.

JanStreffing:
> Hi @christian-stepanek,
> 
> The example namelists still need to be updated, so far this is only directly in the FESOM source code.

The idea was that this is not needed. If someone wants the corners (currently no one does), they can activate that code by adding and the switch to the namelist.config. Otherwise it is false, as set in the code.

JanStreffing (approved):

PR #683, opened 2025-02-05
Add yac interface to couple to ICON

nils3er:
This PR adds yac as a coupling library to FESOM, such that it can be coupled to ICON.


nils3er:
@dsidoren @sveta-loza Can we sort out what needs to be done for merging this.
Especially the build system is a bit convoluted. Maybe @JanStreffing can help to streamline the changes with the existing coupling stuff?

src/gen_modules_partitioning.F90
```diff
@@ -152,10 +160,8 @@ subroutine par_ex(COMM, mype, abort)       ! finalizes MPI
   
   if (mype==0) print *, 'FESOM calls MPI_Finalize'
   call MPI_Finalize(error)
-#endif
-         ! oifs/echam
-#endif
-         ! oasis
+#endif ! oifs/echam
```
JanStreffing, line 163: I don't think all compiler will allow comments on the end of preprocessor statements. This would be reverted

JanStreffing (changes requested):
First, the checks are failing:
https://github.com/FESOM/fesom2/actions/runs/13158564989/job/36721281053?pr=683#step:6:915

After that goes through I'll give this branch a try as coupled to openifs through oasis. Then someone also needs to test coupled to echam through oasis. @christian-stepanek maybe?

JanStreffing:
I got around to testing this branch as part of AWI-CM3 (meaning oasis true, yac false). I get a compile time error:

```
/work/ab0246/a270092/model_codes/awicm3-v3.4/fesom-2.6/src/io_restart.F90:195:90:

  195 |   call ice_files%def_node_var_optional('ice_albedo', 'ice albedo',    '-',   ice%atmcoupl%ice_alb, mesh, partit)
      |                                                                                          1
Error: 'atmcoupl' at (1) is not a member of the 't_ice' structure
```

Full logfile attached.
[comp.log](https://github.com/user-attachments/files/19393257/comp.log)

JanStreffing:
Any update on this?
There seems to be interest in the community to use this interface.

PR #694, opened 2025-03-03
Frank ice

koldunovn:


JanStreffing:
Should we merge this as is, or wait till the next major release and change the default settings for standalone FESOM?

koldunovn:
I will clean it a bit - there are a couple of variables that we can delete. Then we can merge. The plan is also to do some tests with new namelist in a separate issue, but it is irrelevant for this merge. Change to WIP.

JanStreffing (approved):

koldunovn:
@JanStreffing I cleaned up a bit. In order to be able to use Frank's new settings, though, we have to solve the problem with h0. Now it is hardcoded, and setting in namelist do not have an effect:

https://github.com/FESOM/fesom2/blob/d1fa4f6b88fd69bd72e1660bbd40b7e8437075a3/src/ice_thermo_oce.F90#L283-L290

I am for having two separate h0 values. We can add `h0_n` and `h0_s`. @dsidoren @qiangclimate @patrickscholz - any thoughts?

qiangclimate:
it was practically used, to have different h0 values for NH and SH, if needed.

koldunovn:
> it was practically used, to have different h0 values for NH and SH, if needed.

Sure, but:
- one needs to recompile to change the values
- It's confusing, as setting the value in the namelist does not have an effect.

JanStreffing:
I'd say we should 

1. Expose both hemisphere variables to the namelist
2. Ensure that they are by default the same for both. 

I'm not too knowledge with the standalone model, but with coupled model I found years ago that lead closing parameter tuning, especially hemispherical was given the right concentration and thickness, but very wrong T2M. I ended up tuning all three right with hemispherical albedo changes, parameterizing the different melt pond states. See: https://github.com/AWI-ESM/project_management/issues/14

koldunovn:
I did the split of h0, and add namelist for ERA5.

JanStreffing:
ready for merge?

koldunovn:
Yes we can merge it.

PR #695, opened 2025-03-05
capitalization messes up f90nml reader in esm_tools. 

JanStreffing:
need lowercase at the end. Seems to be limitation of fortran namelist to dict parser?

Needs to be tested for DE.


sebastianbeyer:
This is only the variable name which is not case sensitive, or is this different in namelists?

sebastianbeyer (approved):

sebastianbeyer:
we should merge this in any case, we can update our namelists in DestinE, I was just wondering if this makes any difference...

PR #699, opened 2025-04-08
change awicm3 calving / enthalpy logic. Oasis field 13 brings mass in…

JanStreffing:
…stead of energy flux. local computation of enthalpy


tsemmler05 (changes requested):
That looks good to me. Do we need a minus sign in line 372, though?

tsemmler05:
> That looks good to me. Do we need a minus sign in line 372, though?

In addition, the runoff mapper needs to be changed accordingly (runoff_mapper_mod.F90, lines 183/184. In those lines the operation is performed that now should be performed in FESOM).

JanStreffing:
Hello Tido, could you approve the FESOM2 part?
The runoff mapper part is in rnfmap version 1.1.

tsemmler05 (approved):

PR #700, opened 2025-04-08
change default output for temp and salt to monthly

JanStreffing:
I think this would be nice for https://github.com/JanStreffing/cmpitool/
Though it would increase data volumes. Opinions?


koldunovn:
In general fine, but will make quick tests with core2 a bit slower, with a bit more data.

koldunovn (approved):

JanStreffing:
I don't think Nikolay meant the online tests. They never run for a month anyway. Just two days.

koldunovn:
Indeed, the CI is not a concern here - just simple tests. Merging.

PR #701, opened 2025-04-14
Cavity bugfix: from linfs to rigid lid

PengyangSong:
With linfs in cavity, the model cannot run for >100 years due to `ssh` blowup. Therefore we need to change it back to rigid lid.
In `oce_ale.F90`, it works like:

- `compute_ssh_rhs_ale` compute `ssh_rhs`, cavity values are not zero.
- `solve_ssh_ale` solves `d_eta`, cavity values are not zero.
- `update_vel` includes barotropic pressure gradient of `d_eta`.
- `compute_hbar_ale` makes no change in cavity layer thickness `hbar`.
- At cavity nodes, we skip updating `eta_n`, so it is always zero.


JanStreffing (approved):

PengyangSong:
@patrickscholz Could you have a look on my commits here?

patrickscholz:
@PengyangSong: Can you provide a bit more information, whats happening with the blowup? Are we speaking about standalone or coupled, which setup, where exactly does it blowup and under which conditions. Usually a model doesnt blowup after +100 years where it already run stable ? Is the ssh drifting ? ...

src/oce_ale.F90
```diff
@@ -2054,6 +2053,7 @@ subroutine compute_hbar_ale(dynamics, partit, mesh)
     if (.not. trim(which_ALE)=='linfs') then
 !$OMP PARALLEL DO
         do n=1,myDim_nod2D
+            if (ulevels_nod2D(n)>1) cycle
```
patrickscholz, line 2056: this change here affects the zstar case the upper change in line 1931 will affect the linfs case ?

PengyangSong: Yes.

PR #705, opened 2025-05-19
Add option to build mesh partitioner with main build

sebastianbeyer:
Adds a new option BUILD_MESHPARTITIONER to the main fesom build that will build the mesh partitioner as well.

This makes it easier to build it together with the model in the ifs-bundle, so that we can add some automation to generate missing partitioning files. 

I would like to keep this as the only way to build the partitioner to have only one central CMakeLists.txt but I left the current option because I don't know if anything (ESMtools...? @mandresm ) relies on this being build this way.

PR #707, opened 2025-05-26
Build meshpart

suvarchal:
PR to test if CI works with this, then it is @sebastianbeyer 's choice how to use it. :D 

- to build partitioner along with fesom.x one can use ./configure.sh ubuntu  -DBUILD_MESHPART=ON, also add this to ifs bundle this option. old way to build mesh partitioner also works by cd mesh_part;./configure.sh ubuntu.


JanStreffing:
merged https://github.com/FESOM/fesom2/pull/705 instead.

PR #722, opened 2025-06-19
Fix meshpart build

suvarchal:
-  currently meshpart build has by default meshpartitioner is [on always](https://github.com/FESOM/fesom2/blob/ab846d07b93ad2b879b5dae72de062690900849a/src/CMakeLists.txt#L91) and cant be disabled because of its use in [fesom export](https://github.com/FESOM/fesom2/blob/ab846d07b93ad2b879b5dae72de062690900849a/src/CMakeLists.txt#L484), which is fine but mesh partitioner is only build for [gnu and intel](https://github.com/FESOM/fesom2/blob/ab846d07b93ad2b879b5dae72de062690900849a/src/CMakeLists.txt#L468-L476) so it creates issue when used with other compilers as  @basava70 was using nvhpc on levante, this PR addresses minor fix to it.

- cmake minimum version change: use of block in new cmake is supported only from [cmake 3.25](https://cmake.org/cmake/help/latest/command/block.html), can be problematic in some machines, (for instance recent ubuntu LTS has cmake 3.23) we can address this later when issue arises, but it is better to delegate error message to cmake minimum version  required is 3.25 then block statement not found. As we increase cmake versions older versions will be depricated so metis cmake minimum version was bumped up without issues.

not addressed: Partitioner fortran interfaces are not installed because of fesom_export. I plan to have more install targets in the future, it is better to address this later.


suvarchal:
tested on levante with nv compilers and locally, PR is ready.

JanStreffing:
Adding @sebastianbeyer as reviewer, as he last worked on this part of the code.
I see that the default case for mesh-partitioner has changed from ON to OFF. Is this still needed with your fixes? Does it still not build with some compilers?

Furthermore:
```
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.25)
```
Version 3.25 released in November 2022. This seems quite strict. I think there might be some HPC systems out there that have older versions of cmake. What part(s) are missing in 3.16?

suvarchal:
totally agree, thats what i meant above (and others like it blocks compilation for other compilers).
use of block statement https://github.com/FESOM/fesom2/blob/ab846d07b93ad2b879b5dae72de062690900849a/src/CMakeLists.txt#L454 is only supported from 3.25 onwards if we use before then cmake throws error on lines of unrecognized statement setting minversion to 3.25 shows better error that cmake version is incompatible.  #707 doesn´t have that issue but needs more polishing. 
we can also use block if we wrap in a function and specify we can only use it in cmake version above etc, I will address this later when we have more build targets (planned).

JanStreffing:
Pardon, I did not read what you wrote. Just looked at the diff. My bad.

JanStreffing:
Quick scan of supported HPC systems showed none of them use cmake older than 3.26. Chances of newer systems having older cmake version is slim.

suvarchal:
and even so cmake gives binaries that work out of the box for most machines, https://github.com/Kitware/CMake/releases

JanStreffing:
So we found the first maschine that has only make 3.20.2 and not module. It's albedo.

PR #724, opened 2025-06-24
Add ctest automation framework

suvarchal:
adds better test automation framework then our custom automation we have to test fesom.
- mainly support for agentic development, just build and run standard ctest to see if changes worked.
- uses standard ctest framework and dir name: tests
- test data follows directory structure similar to levante
- can easily test matrix tests with different initial data, forcings, and meshes.
- currently only integration tests, but leaves more options for future.
- this will be optional tests for now, eventually we can get rid of current `test` dir till that switch is made we keep both.


sebastianbeyer:
Pretty cool! I did not yet find the integration tests themselves, but I will keep looking xD 
Do you already cover all of the tests that we currently have? Once that is true, we can remove the current github tests, correct?

JanStreffing (approved):

PR #730, opened 2025-06-26
Alternative meshpart

suvarchal:
- remove need for cmake 3.25 and block statement making mesh partitioner as a tool from mesh_part.
- reuse mesh_part infrastructure for building it without duplication. 
- a bug fix: we missed compiling mesh part C sources with proper defines which makes mesh part non deterministic. Basically we didn't use metis preprocessor flags -DMETIS_VERSION=5 -DPART_WEIGHTED -DMETISRANDOMSEED=35243 after fesom used shared library.   

@sebastianbeyer can you check if the fesom_meshpart is in the bin dir of ifs build please? 

I will add ctest for mesh part in another PR.


tests/data/MESHES/pi/elvls.out
JanStreffing, line 1: Was this meant to be deleted?

Issue #721, opened 2025-06-18
6.2.1) --> bugfix: correct the sign of the realsalt flux


patrickscholz:
# Bugfix: sign of real salt flux: 

In the code, the realsalt flux is weighted with a minus sign ...

fesom2/src/oce_ale_tracer.F90](https://github.com/FESOM/fesom2/blob/144f8a399814fe516daf51a520e5d4b6f0001eed/src/oce_ale_tracer.F90#L1509-L1513

The restoring- and real- salt flux analysis from #720 have shown, that in this case the real saltflux has the wrong direction, it should be ... 
```
CASE (2)
    ! --> real_salt_flux(:): salt flux due to containment/releasing of salt
    !     by forming/melting of sea ice
    bc_surface= dt*(virtual_salt(n) & !--> is zeros for zlevel/zstar
                          + relax_salt(n) + real_salt_flux(n)*is_nonlinfs)
```
In the Weddel Sea this could lead to a significant increase in its surface density and instability favoring these very deep mixing events.
Here i already corrected this bug and looked for the magnitude of its effect!

patrickscholz:
- GMOC eddy streamfunction (from bolus_w): 1. ctrl (cm=3, Kgm=1000, Kr_taper=on, S_cr=0.2e-2, ...), 2. with bugfix, for all 3 spinup cycles
![Image](https://github.com/user-attachments/assets/f51ef4b7-c453-4cfb-8cf6-a0be85d56e96)

- GMOC streamfunction (from w+bolus_w): 1. ctrl (cm=3, Kgm=1000, Kr_taper=on, S_cr=0.2e-2, ...), 2. with bugfix, for all 3 spinup cycles
![Image](https://github.com/user-attachments/assets/e3e5cc50-0ba6-485f-b9d2-94a5d3399e74)

- AAMOC streamfunction (from w+bolus_w): 1. ctrl (cm=3, Kgm=1000, Kr_taper=on, S_cr=0.2e-2, ...), 2. with bugfix, for all 3 spinup cycles
![Image](https://github.com/user-attachments/assets/d9c8071a-7d85-424d-886a-4f0585e5ee75)

- IPMOC streamfunction (from w+bolus_w): 1. ctrl (cm=3, Kgm=1000, Kr_taper=on, S_cr=0.2e-2, ...), 2. with bugfix, for all 3 spinup cycles
![Image](https://github.com/user-attachments/assets/f391c206-ba53-47bb-b5d5-1ad026127f56)

[...]

Issue #720, opened 2025-06-18
6.2) --> alternate surface salinity restoring (important found bug!!!)


patrickscholz:
# Check effect of sea surface salinity restoring:

Sea surface salinity (SSS) restoring is a common technique used in standalone ocean models to improve the realism and stability of simulations over long time periods. In the absence of a dynamically coupled atmosphere, standalone models rely on prescribed surface fluxes—such as precipitation, evaporation, river runoff, and ice melting/freezing—from observational datasets or reanalyses. However, these fluxes often contain biases, are incomplete, or do not perfectly balance globally, leading to gradual drift in surface salinity.

Salinity plays a crucial role in determining seawater density, which in turn influences key components of ocean circulation, such as the thermohaline circulation and deep water formation. If surface salinity drifts unrealistically, it can cause errors in density fields, potentially altering circulation patterns, stratification, and vertical mixing. These changes can propagate throughout the ocean, undermining the physical reliability of the simulation.

To mitigate this, SSS restoring applies a relaxation term that nudges the model’s surface salinity toward observed or climatological values. Mathematically, this is implemented as a Newtonian relaxation:

$$
\text{Restoring term} = -w_p *  (S_{\text{model}} - S_{\text{target}})
$$

$$
w_p = h_{\textrm{surface}} /  {\tau}
$$

where $w_p$ is the restoring rate or piston velocity, with $h_{surface}$ being the surface layer thickness and ${\tau}$ being the restoring time scale (typically ranging from \~30 to 300 days), and $S_{\text{target}}$ is the observed or climatological surface salinity. This term is added to the surface freshwater budget and effectively mimics a virtual source or sink of freshwater, depending on the salinity bias.

Restoring should be applied carefully: if it's too strong, it can suppress natural variability and introduce artificial feedbacks; if too weak, it may not prevent drift effectively. In practice, moderate restoring ensures that the model retains its ability to simulate variability while remaining anchored to realistic surface conditions. This is especially important for climate simulations, historical ocean state reconstructions, or ocean-only sensitivity studies where accurate representation of the mean state is crucial.

In summary, SSS restoring is a pragmatic solution to compensate for missing or imperfect atmospheric feedback in standalone ocean models. It helps maintain physically consistent salinity and density fields, supports realistic ocean circulation, and prevents the accumulation of long-term biases due to errors in surface freshwater forcing.

[...]

PR #733, opened 2025-06-30
Found Bug in the sign of the realsalt flux

patrickscholz:
# Bugfix: sign of real salt flux: 

In the code, the realsalt flux is weighted with a minus sign ...

fesom2/src/oce_ale_tracer.F90](https://github.com/FESOM/fesom2/blob/144f8a399814fe516daf51a520e5d4b6f0001eed/src/oce_ale_tracer.F90#L1509-L1513

The restoring- and real- salt flux analysis from #720 have shown, that in this case the real saltflux has the wrong direction, it should be ... 
```
CASE (2)
    ! --> real_salt_flux(:): salt flux due to containment/releasing of salt
    !     by forming/melting of sea ice
    bc_surface= dt*(virtual_salt(n) & !--> is zeros for zlevel/zstar
                          + relax_salt(n) + real_salt_flux(n)*is_nonlinfs)
```
In the Weddel Sea this could lead to a significant increase in its surface density and instability favoring these very deep mixing events. Here i already corrected this bug and looked for the magnitude of its effect!

See Issue #721


JanStreffing:
CI test results need to be updated, then it can be merged.

dsidoren:
> Great find! The literature is clear, the results are clear. Coupled test is not needed, as we don't do salt relaxation.

it is not about relaxation but about the salt stored in sea ice. Coupled simulations shall become better then? Right, @patrickscholz ? Crazy bug! Thank you Patrick :). Linear free surface is unaffected?

JanStreffing:
I'll do a coupled test then.

JanStreffing:
Given the changes in this branch, is this line correct?
https://github.com/FESOM/fesom2/blob/9bc7e43aba62dc2b4909e5fcbda8e06e6bbb7add/src/oce_ale_tracer.F90#L529

JanStreffing:
New test results on main test are still the same as with yesterdays fix. So I assume that while the fix was needed, we were not using this routine?

patrickscholz:
@JanStreffing: yes the usage of the explicite vertical diff is switched of by namelist flag, we usually use by default only the implicite vertical diff!

JanStreffing:
@sveta-loza

JanStreffing:
A first look at the bugfix in AWI-CM3 v3.3.0. Looking at last 15 years of 30 year long const 1990 forcing runs. 

Biases with fix as fraction of biases without fix:

Salinity improvements in upper 100m in Antarctic, upper 10m in Arctic. Arctic also has better siconc and tas. Nino34 region got colder and has larger biases. Looks dramatic, but is not.

Biases with and without fix as fraction of mean CMIP6 biases:


Overall, a slight model performance increase with regional differences. If you would like to make your plots, the runs are at:

- /work/bb1469/a270092/runtime/awicm3-v3.3.0/REF_V330/outdata/fesom/
- /work/bb1469/a270092/runtime/awicm3-v3.3.0/RSFIX_V330/outdata/fesom/

I will extend these to 50 years still, and also make a set based on awicm3-develop with OpenIFS 48r1.

PR #740, opened 2025-07-11
Runid fixes

suvarchal:
tiny fixes for when runid is used it should consistently be used in naming (useful for testing and benchmarking without need to create new result dirs)
- fixes name of meshdiag not to have empty spaces when runid is smaller then its defined size. 
- fixes names of bin restart and raw restart dirs to use runid.

### DIFF
--- a/src/io_restart.F90
+++ b/src/io_restart.F90
@@ -300,12 +300,12 @@ subroutine restart(istep, nstart, ntotal, l_read, which_readr, ice, dynamics, tr
   ! initialize directory for core dump restart 
   if(.not. initialized_raw) then
     initialized_raw = .true.
-    raw_restart_dirpath  = trim(ResultPath)//"fesom_raw_restart/np"//int_to_txt(partit%npes)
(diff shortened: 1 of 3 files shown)


src/io_restart.F90
```diff
@@ -300,12 +300,12 @@ subroutine restart(istep, nstart, ntotal, l_read, which_readr, ice, dynamics, tr
   ! initialize directory for core dump restart 
   if(.not. initialized_raw) then
     initialized_raw = .true.
-    raw_restart_dirpath  = trim(ResultPath)//"fesom_raw_restart/np"//int_to_txt(partit%npes)
-    raw_restart_infopath = trim(ResultPath)//"fesom_raw_restart/np"//int_to_txt(partit%npes)//".info"
+    raw_restart_dirpath  = trim(ResultPath)//trim(runid)//"_raw_restart/np"//int_to_txt(partit%npes)
```
sebastianbeyer, line 303: I am not sure if this will break the restarts how we currently use them in RAPS...

sebastianbeyer: ah, now I understand, yes, that should not be a problem then!

sebastianbeyer (approved):

PR #745, opened 2025-07-23
Fix netcdf interfaces

suvarchal:
use fortran interfaces (and specifically fortran90 interfaces).  use of these interfaces means netcdf does data conversions and calls to writing data and attributes are simpler. For instance, use of netcdf interfaces how data variables is defined eliminating need for our own interfaces for writing like for real4 and real8. Additionally we need this for netcdf parallel io and for mixed/single prec mode otherwise we need to keep writing needless new interfaces. 

**viola! and all the compiler warnings for netcdf are gone!**   

Todo later: we have too many custom netcdf interfaces defined at many places need to consolidate them at one place, restart writing needs also a overhaul to use netcdf90 interfaces.


pgierz:
I'm going to trust the CI here...the signatures are similar enough that my brain is too human and not machine enough to see anything after a few lines. That's what the robots are for ;-)

pgierz (approved):

PR #748, opened 2025-07-28
 fix bug in the definition of the sea-ice metric_factor term

patrickscholz:
- Qiang found a stupid bug in the definition of the metric_factor term that is used by the sea-ice
- I actually perform some runs to see how big is the effect!

### DIFF
--- a/src/oce_mesh.F90
+++ b/src/oce_mesh.F90
@@ -2510,7 +2510,7 @@ t0=MPI_Wtime()
  center_x(n)=ax
  center_y(n)=ay
  mesh%elem_cos(n)=cos(ay)
- mesh%metric_factor=tan(ay)/r_earth
+ mesh%metric_factor(n)=tan(ay)/r_earth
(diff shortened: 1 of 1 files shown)


patrickscholz:
- MLD2
abs_MLD2_NH_y1999-2019_m3
abs_MLD2_SH_y1999-2019_m9

JanStreffing:
Should we collect bug fixes and tunings that change results until we are ready to merge them all and call it 2.7?

patrickscholz:
- vertical profilers temp, salt pdens  NH
temp_vline_NH_y1999-2019
salt_vline_NH_y1999-2019
pdens_vline_NH_y1999-2019

patrickscholz:
- vertical profiles, temp, salt, pdens SH
temp_vline_SH_y1999-2019
salt_vline_SH_y1999-2019
pdens_vline_SH_y1999-2019

PR #750, opened 2025-08-08
Bugfix oce_ale_tracer.F90

mbutzin:
Bugfix for wrong CFC-11 input which affects all tagged versions >= 2.6


a270105 (commented):
Should '- real_salt_flux(n)*is_nonlinfs)' be '**+** real_salt_flux(n)*is_nonlinfs)'? Patrick found the bug some weeks ago, right?
Sorry I can't say much about Martin's update since I am not using CFC.

PR #752, opened 2025-08-13
adapt the metadata for multio 2.6.0

stefaniereuter:
We want to update the bundle to a new version of Multio. One of the keywords have changed and would require this fix in Fesom. I have based the PR on the current version that is used in the ClimateDT bundle. I don't know if there is a planned update for the next run so potentially this could be just a hotfix on 2.6.7.1 .


reuterbal:
Hi, any chance we could get a new release with this change included for use in the Climate DT?

PR #755, opened 2025-08-25
Fix/icb and cav

ackerlar:
Fix heat flux of icebergs and icebergs avoiding cavities


PengyangSong:
Hi @ackerlar and @JanStreffing 
I would not suggest to put the change from src/gen_forcing_couple.F90 in the code. 
This is just an expedient for a dynamic Antarctic Ice Sheet. We can use it in our own code but it could be easily overlooked and become a dangerous block in future developments.
The right solution is to forbid glacial runoff from AIS in ECHAM, right?

JanStreffing (approved):

PR #758, opened 2025-09-02
place debug option behind cmake switch buildtype=debug, instead of co…

JanStreffing:
…mments

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -308,18 +308,31 @@ if(${CMAKE_Fortran_COMPILER_ID} STREQUAL  Intel OR ${CMAKE_Fortran_COMPILER_ID} 
       target_compile_options(${PROJECT_NAME} PRIVATE -xHost)
    endif()
    
-   # # debugging related compiler flags
-   # target_compile_options(${PROJECT_NAME} PRIVATE -qopenmp -g -traceback -check all,noarg_temp_created,bounds,uninit ) #-ftrapuv ) #-init=zero)
-   # target_compile_options(${PROJECT_NAME} PRIVATE -g -traceback -check all,noarg_temp_created,bounds,uninit ) #-ftrapuv ) #-init=zero)
+   # debugging related compiler flags (activated for Debug build type)
+   if(CMAKE_BUILD_TYPE STREQUAL "Debug")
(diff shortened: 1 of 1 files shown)


configure.sh
suvarchal, line 1: i wonder if we should just remove -DCMAKE_BUILD_TYPE=.... from ./configure.sh because 1. ./configure.sh also takes cmake args so gives flexibility for user to use ./configure.sh blahblah -DCMAKE_BUILD_TYPE=Debug  if they want to. 2. in CMakeLists.txt build type is set to be Release in case the variable is not set.

src/CMakeLists.txt
```diff
@@ -357,29 +371,34 @@ elseif(${CMAKE_Fortran_COMPILER_ID} STREQUAL  GNU )
...
-                           #-pg 
-                           #-fbacktrace 
-                           #-fcheck=all,bounds 
-                           #-finit-real=snan 
-                           #-finit-integer=-9999 
-                           #-fsanitize=undefined,address
-                        #)
-   ## if use -fsanitize=undefined,address you also need ... PRIVATE -lubsan)           
-   #target_link_libraries(${PROJECT_NAME} PRIVATE -lubsan)
+   # debugging related compiler flags (activated for Debug build type)
```
suvarchal, line 385: we should use address sanitizer definitely for testing new code to check if we are allocating badly and creating leaks but I will skip -fsanitize for now as it need libasan to also be installed and present, this possibly need a cmake module to check this. so would skip this for now just, and ubsan seems to be a typo? should be -lasan or -static-libasan to compile options. More over any use of asan also need a env variable or someting to trace specific issue like ASAN_OPTIONS=detect_leaks=1 for detecting leaks. so this should be possibly done differently for it to be useful.

suvarchal (approved):
G.Catch!! 

TODO:  Some what related to this there is some  nesting and repetition of compilers and machine blocks for instance for the same compiler: gnu options on levante and on ubuntu differ some parts are machine specific and is needed but others should be common like link time optimization like -flto and it does make a difference as #775  showed runs in ci env and failed with gnu on levante.  when organized better so all the machine specific can go into env files and rest in src/CMakeLists.txt so that we remove clutter.

PR #759, opened 2025-09-02
added the missing include 'netcdf.inc' statement to the gen_surface_f…

JanStreffing:
…orcing.F90 file. This will provide the Fortran 77 NetCDF interface constants like nf_noerr, nf_nowrite, etc. that the code is using. Code was not compiling with GNU and debug flags before.


suvarchal:
I would use use netcdf module instead which makes compiler do better type checking. will get back on this shortly.

JanStreffing:
This needs merging. @a270105 already found and fixed this independently again on here branch: https://github.com/FESOM/fesom2/pull/681/commits/37d16f97eebd17d978c8fee3cf9d58caf6491689

PR #770, opened 2025-09-09
Improve redi + retune default case

JanStreffing:


JanStreffing:
This branch works as part of AWI-CM3. However it does not work, both coupled and standalone, when one selects OMP threads larger than 1.

suvarchal (approved):
trusting @patrickscholz :D

PR #775, opened 2025-09-15
bugfix in reading elem2D_nodes. Fixes #652 and Fixes #653

### LINKED ISSUE #652: wrong dimention
https://github.com/FESOM/fesom2/blob/b6e661c8492caecc2d9b4d094512289295dda439/src/oce_mesh.F90#L2151C53-L2151C66
shall be without "+eXDim_elem2D"?

### LINKED ISSUE #653: wrong dimention
https://github.com/FESOM/fesom2/blob/b6e661c8492caecc2d9b4d094512289295dda439/src/oce_mesh.F90#L458
shall be allocate(mesh%elem2D_nodes(3, myDim_elem2D)

dsidoren:
it could potentially read to errors in reading mesh.


JanStreffing:
Unexpectedly, this seemingly minor fix changed results. And at first glance not in a minor way. Needs to be tested.

suvarchal:
with @dsidoren  ran core2 simulations differences in first year and differences for temp and sst are significant as illustrated below @dsidoren will do more analysis on differences and update.

```
  RESULT       MIN DIFF    MAX DIFF   REL ERR    MASK   VAR    DESCR
 ────────────────────────────────────────────────────────────────────
   ❌ FAILED   -1.79e+00   2.43e+00   3.72e+04   True   temp   -
   ✔ PASSED    0.00e+00    0.00e+00   0.00e+00   True   nz1    -
   ✔ PASSED    0.00e+00    0.00e+00   0.00e+00   True   time   -



             f1/sst.fesom.1958.nc vs f2/sst.fesom.1958.nc

  RESULT       MIN DIFF    MAX DIFF   REL ERR    MASK   VAR    DESCR
 ────────────────────────────────────────────────────────────────────
   ❌ FAILED   -9.21e-01   1.07e+00   2.51e+02   True   sst    -
   ✔ PASSED    0.00e+00    0.00e+00   0.00e+00   True   time   -
```

suvarchal:
> Are we happy with the full cycle impacts of these fixes?
1. for mesh allocation fix
here are diffs after 60y for sss and sst from @dsidoren 
sss_2017
sst_2017

@dsidoren thinks the differences are like an ensemble. we could generate mesh differently with different seed and compare but regardless of differences we are convinced that this is a bug which needs to be fixed.

for another fix in this is another major bug for ice dynamics (EVP and maEVP) where metric factor is used for derivatives for 
previously used 
https://github.com/FESOM/fesom2/blob/72f0b986275ab75746b7efb8f90dec6ecf84e5cf/src/oce_mesh.F90#L2508-L2515
assigned metric_factor that varies for each point based on latitude to simply last value in partition, missing (n) in the loop, and based on how elongated the partition is in the latitude it make it can have significant influence and brings in sensitivity to number of partitions.
While we may want to understand what this does the fix is necessary:
https://github.com/FESOM/fesom2/blob/2302bd7aeb11d6a4a81faef3e988c98e377b7901/src/oce_mesh.F90#L2513

JanStreffing (approved):

PR #777, opened 2025-09-17
Melt pond parameterization for IFS/OIFS coupled runs

JanStreffing:
Getting desperate with the NH / SH, cold / warm dipol bias here. 

I decided to try and lift the meltpond scheme from icepack and use it in ice_thermo_cpl. The idea is, that NH should have more meltponds than SH, due to thicker multi year ice and less snow cover. Then I can increase base sea ice albedo values cooling SH and let the meltponds take care of lowering these back down for NH. My hope is that this reduces both issues.

First evaluation from a 4 year long coupled pi spinup run:
monthly_apnd_nh_sh
NH fraction reaches realistic and higher values than SH fraction.
monthly_hpnd_nh_sh
Meltpool depth also looks ok.
monthly_ipnd_nh_sh
Peak at 4m ice lid is a bit much for freezing in one winter imo. But mean values look fine.

Next I will do some longer runs with sea ice albedo output. Also checking the climate impact.

### DIFF
--- a/config/namelist.io
+++ b/config/namelist.io
@@ -37,6 +37,9 @@ io_list =  'sst       ',1, 'm', 4,
            'a_ice     ',1, 'm', 4,
            'm_ice     ',1, 'm', 4,
            'm_snow    ',1, 'm', 4,
+           'apnd      ',1, 'm', 4, ! melt pond area fraction (requires use_meltponds=.true.)
+           'hpnd      ',1, 'm', 4, ! melt pond depth (requires use_meltponds=.true.)
+           'ipnd      ',1, 'm', 4, ! pond ice lid thickness (requires use_meltponds=.true.)
            'MLD1      ',1, 'm', 4,
            'MLD2      ',1, 'm', 4,
            'MLD3      ',1, 'm', 4,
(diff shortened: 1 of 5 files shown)


JanStreffing:
So far looks to be effective. With +3% base albedo it's mostly a warming of NH. With +5% base albedo its a regional change in NH and a cooling of SH.

fesom_alb_comprehensive_comparison_diff_1351_1357
fesom_a_ice_comprehensive_comparison_diff_1351_1357
fesom_m_ice_comprehensive_comparison_diff_1351_1357
fesom_sst_comprehensive_comparison_diff_1351_1357

Overall seems like a good step forward.

lettie-roach:
I agree, the meltpond fraction area and mean depths look reasonable, peak depths seem a bit high. I'm not following that the different columns are in your spatial plots - e.g. meltpond off, tune on - is this tuning the ocean albedo?

suvarchal:
> Next: Trying to figure out why tests are failing. Any idea @suvarchal or @patrickscholz?

still weird why tests work after you comment out those 4 output lines

patrickscholz (approved):

PR #778, opened 2025-09-17
Mesh part cleanup

suvarchal:


suvarchal:
closed because of complex merges, replaced with #779

PR #780, opened 2025-09-18
add Intel compiler env for ubuntu

suvarchal:
- and build netcdf on the fly for new compiler with -DBUILD_NETCDF=ON , also needed for testing upcoming parallel netcdf support.
- add ci test for intel build with minimal test, optionally make it only for tags and releases later.


suvarchal (commented):
checked once again.

PR #785, opened 2025-10-02
Set metric_factor to zero as suggested by Qiang.

nwieters:
<p>This improves the Arctic sea ice thickness distribution in AWI-CM3.</p>
 before vs after
<p float="left">
  
  
  
  
</p>


koldunovn:
Note, that Sergey have very strong opinion against having this by default!

JanStreffing:
Why are do the results still differ? Did I make a mistake in the logic here?

PR #786, opened 2025-10-08
update restructure cvmix

patrickscholz:
- Update CVMix to the latest version 
- restructure CVMIx move interface wrapper file into ./src/cvmix_driver
- have own shell script get_cvmix.sh to download and install the CVMix library into ./src/cvmix_driver
- change CMakeList.txt so CVMix ist downloaded, installed and linked when fesom is compiled


patrickscholz:
@JanStreffing: Why renaming from USE_CVMIX to WITH_CVMIX, we already have plenty of compiler flags that start with USE_... i would rather unify it instead of starting to bring in another style with WIDTH_...

JanStreffing:
Feel free to undo.

PR #799, opened 2025-10-28
set missing value _FillValue attribute to netcdf outputs closes #289

### LINKED ISSUE #289: change missval from 0 to nan
Currently the missval in default fesom2 output where bathymetry cuts off the data is 0. The result is that addressing the wrong section of arrays does not lead to a crash. It seems to me that this removes a safety layer, as one might address values underground without noticing. One such case was: https://github.com/FESOM/fdiag/issues/10#issuecomment-1023701955

I suggest we switch to using nan or something like -9e+33.

suvarchal:
adds long needed setting missing values in handles/tested for both cases of output in single precision or double precision. closes #289.  Here it is works with cdo and xarray nansums. 

Before: 
temp:
image
a_ice:
image

After:
temp
image
a_ice
image

note missvals , mins, means


JanStreffing:
I will update the truths.

patrickscholz:
In tripyview it should not be a large problem, because there i added the land sea mask anyway from hand with nan, a step thats not needed now anymore. Only for this whole moc computation it was helpfull that all lsmask values were 0, because it was easier for the integration. This part i need to test

patrickscholz:
@suvarchal: why is the missing value number so odd 9.9692099683868690e+36_real32? Is this some Indian good look number or is it some sort of convention?

patrickscholz:
@suvarchal : attaching the grid to the data, should remain an optional argument? Especially for the very large grids, i guess we want to keep the data as slim as we can.

patrickscholz (approved):

chrisdane:
I am unsure if related but could it be that these changes result in a zero sea ice concentration being saved as nan to the output file? If yes, I wonder if this is correct. I think it makes more sense to save a zero sea ice concentration as zero, and not as nan @suvarchal @JanStreffing @patrickscholz (this is also the cmip cmor default). Thx!
(all sea ice variables would be affected of course)

patrickscholz:
@chrisdane: If i remember well (but correct me if im wrong @JanStreffing) we switched this off for the 2d fields! The problem is a bit that we set all occurences of zero to NaN or 3...e30 or something like this, we do not realy check for the land sea mask here. We just assume that only land values are zeros but it also occured to me that some ocean values (idemix energy related values) can remain zero which are than also setted to NaN.

JanStreffing:
@chrisdane is right. We are turning 0 Sea ice thickness and concentration to NaN..

PR #806, opened 2025-11-01
AWI-ESM3 CO2 coupling

### LINKED ISSUE #889: recom forcing files needed in both ocean-only and coupled
In 7e22dcf0, recom forcing routines were transferred from recom to `gen_surface_forcing.F90`s `sbc_do`, e.g. 
```
! ******** Fe deposition *********
    if (fe_data_source=='Albani') then
        if (update_monthly_flag) then
            if (mype==0) write(*,*) 'Updating iron climatology for month       ', i,' from ', trim(filename)
```

Until today, recom expects these forcing files, no matter if ocean-only or coupled mode.

@a270105 now realized that `sbc_do` is only called in ocean-only, and not in coupled mode. Hence, during the coupled run the recom forcing fields stay nan and and added to the rhs, why the output (e.g. `aFe`) is also nan.

@ogurses @JanStreffing: In current main 45026e58, the recom block in `gen_surface_forcing.F90:L1661ff` needs to be called in both modes, ocean-only and coupled 🙏

### LINKED ISSUE #971: fesom2.6_recom gives zoo3 output as coocs output in 2p3z2d option
Hi, I came across an inconsistency in naming the tracers in the recom part of 2.6.
If you run fesom2.6_recom_tra, the zoo3 tracer gets outputted as coccos. That means in commit 12f739a it reads as:
recom/recom_modules.F90

        if (enable_3zoo2det .and. enable_coccos) then
            expected_tracer_ids(32) = 1030 ! CoccoC
        else if (enable_3zoo2det .and. .not.enable_coccos) then
            expected_tracer_ids(32) = 1030 ! Zoo3C

But 
io_meandata.F90

      else if (tracers%data(j)%ID==1030) then
         if (use_REcoM) then
         call def_stream((/nl-1, nod2D/),  (/nl-1, myDim_nod2D/),  'CoccoC', 'CoccoC', '[mmol/m3]', tracers%data(j)%values(:,:), io_list(i)%freq, io_list(i)%unit, io_list(i)%precision, partit, mesh)

In 2.7 this is fixed by introducing an if in the io_meandata, which I now temporarily implemented in my version just for tracer 1030, but it would be nice to have it for all of the 2.6 recom branches.

### LINKED ISSUE #1003: Benthic Si and calcite are halved when num_fesom_groups > 1
With REcoM tracer parallelisation (`__recom` + `__usetp`), `BenSi` and `BenCalc` come out at exactly half their serial value when `num_fesom_groups=2`. `BenN` and `BenC` are unaffected.

Setup: pi test mesh, 1 day, 32 steps/day, KPP, `enable_3zoo2det=.true.` (30 BGC tracers). Two runs differing only in group count, with 2 ranks per group in both cases so the ocean decomposition is `dist_2` either way. One is `num_fesom_groups=1` on 2 ranks, the other `num_fesom_groups=2` on 4 ranks.

Comparing the whole restart (369 variables), 295 are bit identical and 72 differ at 1e-13 or below, which is round off on near zero tracers. Two do not:

| variable | ratio 2grp / 1grp |
|---|---|
| `BenN` | 1.000000 |
| `BenC` | 1.000000 |
| `BenSi` | 0.500000 |
| `BenCalc` | 0.500000 |

For those two, min = max = median, so the factor is exactly `1/num_fesom_groups` at every node. Reproduced on a clean `main` (45807953), so this is not specific to a feature branch.

Likely cause: `Benthos(n,1:4)` is accumulated inside the per tracer sinking loop in `recom_sinking.F90` (L115-L189), which under `__usetp` only covers the group's own tracer slice. The `recom_exchange_nod(Benthos(:,n), ..., MPI_COMM_FESOM, ...)` call at L288 is a group local halo exchange, so it does not merge contributions across groups. `Benthos_tr` is broadcast across groups in `oce_ale_tracer.F90` (L410, L434), but I cannot find anywhere that it is summed back into `Benthos`, and `io_restart.F90` writes `Benthos(:,1)` through `Benthos(:,4)` directly from group 0.

So the question is whether `Benthos` is meant to be reconstructed from the per tracer `Benthos_tr` buffers under `__usetp`, with that reduction currently missing. It would also be good to understand why slots 1 and 2 survive intact while 3 and 4 halve. My guess is that the tracers feeding N and C land in group 0's slice while diatom and detrital Si and calcite are split across groups, but I have not confirmed the slice boundaries.

`GlodecayBenthos` feeds the bottom fluxes for Si and for DIC and alkalinity, so a halved benthic pool halves the benthic return flux. In a one day run this is only a restart value, but it would compound over a long simulation.

Filed here rather than against REcoM because the REcoM sources contain no `__usetp` code at all. The group machinery and the `Benthos_tr` broadcast both live on the FESOM side, so any cross group reduction would have to be added here.

JanStreffing:
Adds interactive CO2 exchange between FESOM/REcoM and OpenIFS, plus the fixes needed to make the coupled 2p3z2d configuration run.

**CO2 coupling.** Extends the OASIS interface with one send and one receive. `FCO2_feom` carries REcoM's air-sea CO2 flux to the atmosphere, converted from `GloCO2flux_seaicemask` in mmolCO2/m2/s to kgCO2/m2/s and sign flipped to OIFS's downward-positive convention. `XCO2_oce` returns the atmospheric mixing ratio into REcoM's `x_co2atm`, so `constant_CO2` runs can be replaced by a prognostic one. Send count goes from 8 to 9, receive from 15 to 16.

**Sea ice skin temperature.** Adds an `ist_ref` anchor in `ice_thermo_cpl` and `MOD_ICE`, recording the ice surface temperature the atmosphere evaluated its ice-tile fluxes at, so the internal surface-temperature solve linearises around the same point the atmosphere used.

**Cavity NaN fixes.** `gen_ic3d` now skips climatology points that are NaN or fill valued when interpolating initial conditions near cavities, and `oce_ale` cleans up NaN produced at cavity nodes by tracer advection and diffusion where layer thickness goes to zero.

**REcoM output and restarts.** `io_meandata` resolves the configuration dependent tracer names through `tracer_ids%` instead of hardcoded ids, so ids 1023 to 1036 are labelled correctly in every `&parecomsetup` combination. This fixes #971. `io_restart` identifies REcoM tracers explicitly rather than by id range.

**Namelists.** Adds `config/namelist.recom.2p3z2d`, `config/namelist.io.recom.2p3z2d` and `config/namelist.tra.recom.2p3z2d` at the config root, and updates the existing `config/bin_2p3z2d` set. Bumps the REcoM submodule from 12f739a3 to bfa321d6.


JanStreffing:
This branch still needs some cleanup. I very optimistically add it to 2.7 for now.

JanStreffing:
@chrisdane, were you able to resolve the conflicts locally? Or need help?

JanStreffing:
This PR also fixes #971. The `tracer_ids%` lookups it introduces in `io_meandata.F90` replace the hardcoded tracer ids, so id 1030 is labelled `Zoo3C` under 3zoo2det instead of `CoccoC`, which is the mismatch reported there.

JanStreffing:
@chrisdane, you said there was nothing left to push on your side, but I had to doctor around quite a bit to get this to run. Is this all correct as is?

patrickscholz (approved):

PR #810, opened 2025-11-05
Complete standard namelists makeover providing more context to the user.

FinnHeu:
I reformatted the namelists and  added a lot of context that was previously hidden in the source code to the standard namelist files to make it easier for beginners to understand the various parameters of the namelists. 

Modifications include:
- A proper header for each namelist file
- Proper headings for each &-block in the respective namelists
- Context for each parameter that is set (AI searched the source code for information and helped a lot here)
- Consistent formatting
- In namelist.io, I added a complete catalogue of all possible outputs to the end of the file so one does not have to browse through io_meandata.F90 each time to look for outputs. 

As a consequence, the namelists are not as 'clean' as they were before but to me it's still an improvement which I would like to see merged in 2.7. Generally, /config became quite messy lately with a lot of additional adaptations of the standard namelists - I haven't touched those - but I could also do the same for the remaining namelists if that's appreciated.

What do you think?

### DIFF
--- a/config/namelist.io
+++ b/config/namelist.io
@@ -158,9 +158,9 @@ io_list =  'sst       ',1, 'm', 4,
 ! 'swice     ',1, 'm', 4,  ! swice [?]
 
 ! --- MIXED LAYER DEPTH ---
-! 'MLD1      ',1, 'm', 4,  ! Mixed Layer Depth [m]
-! 'MLD2      ',1, 'm', 4,  ! Mixed Layer Depth [m]
-! 'MLD3      ',1, 'm', 4,  ! Mixed Layer Depth [m]
+! 'MLD1      ',1, 'm', 4,  ! Mixed Layer Depth [m] Large et al. 1997, bvfreq(nz, node) > db_max
+! 'MLD2      ',1, 'm', 4,  ! Mixed Layer Depth [m] Levitus treshold, rhopot(nz)-rhopot(1) > 0.125_WP kg/m
+! 'MLD3      ',1, 'm', 4,  ! Mixed Layer Depth [m] Griffies 2016 , rhopot(nz)-rhopot(1) > 0.03_WP kg/m
 
 ! --- HEAT CONTENT (require ldiag_destine=.true.) ---
 ! 'hc300m    ',1, 'm', 4,  ! Vertically integrated heat content upper 300m [J m**-2]


config/namelist.config
```diff
@@ -1,77 +1,133 @@
...
+! ============================================================================
+! MESH, INITIALIZATION & OUTPUT PATHS
+! ============================================================================
 &paths
-MeshPath='/pool/data/AWICM/FESOM2/MESHES/CORE2/'
-ClimateDataPath='/pool/data/AWICM/FESOM2/MESHES/INITIAL/phc3.0/'
-ResultPath='/work/ab0246/$USER/runtime/fesom2/'
+MeshPath         = '/albedo/pool/FESOM2/core2/'               ! path to mesh files (nod2d.out, elem2d.out, etc.)
+ClimateDataPath  = '/albedo/pool/FESOM2/hydrography/phc3.0/'  ! path to initial conditions (temperature, salinity)
+ResultPath       = '<add path>'                               ! path for output files and fesom.clock file
```
patrickscholz: I think Resultpath must remain meaningful as it was otherwise the CI test seems to fail !

config/namelist.config
```diff
@@ -1,77 +1,133 @@
...
+! ============================================================================
 &timestep
-step_per_day=32 !96 !96 !72 !72 !45 !72 !96
-run_length=62 !62 !62 !62 !28
-run_length_unit='y' ! y, m, d, s
+step_per_day      = 96         ! number of time steps per day (determines dt = 86400/step_per_day seconds)
+                                ! common values: 32 (45min), 48 (30min), 72 (20min), 96 (15min), 192 (7min 30sec), 
+                                ! 216 (6min 40sec), 240 (6min), 288 (5min), 360 (4min), 720 (2min), 1440 (1min), 2880 (30sec)
+run_length        = 1          ! total length of simulation run
+run_length_unit   = 'm'        ! unit for run_length: 'y' (years), 'm' (months), 'd' (days), 's' (seconds)
```
koldunovn: 's' is not seconds, it's steps if something did not change from the times I run the model :)

config/namelist.io
```diff
@@ -1,46 +1,65 @@
...
+!   precision  = 4 (single precision) or 8 (double precision)
+! ============================================================================
 &nml_list
 io_list =  'sst       ',1, 'm', 4,
            'sss       ',1, 'm', 4,
-    	   'ssh       ',1, 'm', 4,
-           'uice      ',1, 'm', 4,
-           'vice      ',1, 'm', 4,
+           'ssh       ',1, 'm', 4,
+           'uice      ',1, 'd', 4,
```
koldunovn: Why change to `d` for uice and vice? I would say by default we should have months everywhere.

config/namelist.io
```diff
@@ -63,3 +82,254 @@ io_list =  'sst       ',1, 'm', 4,
            'fh        ',1, 'm', 4,
            'otracers  ',1, 'y', 4,
 /
+
+! ============================================================================
+! COMPLETE CATALOG OF ALL POSSIBLE OUTPUT FIELDS
```
koldunovn: OMG, @FinnHeu you are great! :)

koldunovn:
@FinnHeu thanks a lot, very useful indeed!

config/namelist.io
FinnHeu: Good point, I changed all default output frequencies to monthly, in both the standard output list and the catalogue below.

config/namelist.config
```diff
@@ -1,77 +1,133 @@
...
 &restart_log
-restart_length=1            ! --> do netcdf restart ( only required for d,h,s cases, y, m take 1)
-restart_length_unit='y'     !output period: y, d, h, s, off
-raw_restart_length=1        ! --> do core dump restart
-raw_restart_length_unit='y' ! e.g. y, d, h, s, off
-bin_restart_length=1        ! --> do derived type binary restart 
-bin_restart_length_unit='y' ! e.g. y, d, h, s, off
-logfile_outfreq=960         !in logfile info. output frequency, # steps
+restart_length          = 1        ! frequency for netCDF restart files (required for d,h,s; y,m use 1)
+restart_length_unit     = 'm'      ! unit: 'y' (years), 'm' (months), 'd' (days), 'h' (hours), 's' (seconds), 'off' (disabled)
```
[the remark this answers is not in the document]
patrickscholz: You can make it 100%!

config/namelist.config
suvarchal: it is still steps :)

config/namelist.config
```diff
@@ -1,77 +1,133 @@
...
+! ============================================================================
+! MESH, INITIALIZATION & OUTPUT PATHS
+! ============================================================================
 &paths
-MeshPath='/pool/data/AWICM/FESOM2/MESHES/CORE2/'
-ClimateDataPath='/pool/data/AWICM/FESOM2/MESHES/INITIAL/phc3.0/'
-ResultPath='/work/ab0246/$USER/runtime/fesom2/'
+MeshPath         = '/albedo/pool/FESOM2/core2/'               ! path to mesh files (nod2d.out, elem2d.out, etc.)
+ClimateDataPath  = '/albedo/pool/FESOM2/hydrography/phc3.0/'  ! path to initial conditions (temperature, salinity)
+ResultPath       = '/work/ab0246/$USER/runtime/fesom2/'       ! path for output files and fesom.clock file
```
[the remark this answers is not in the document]
suvarchal: I would use levante paths for consistency and we have most of the setups and forcings there.

JanStreffing:
> @FinnHeu: The general, cavity, channel and iceberg testcase keeps failing !

I think this is rather a problem with our checks, and the way we modify the default namelists.

FinnHeu:
@patrickscholz @JanStreffing, all checks approved. :)

PR #819, opened 2025-11-12
Separated liquid and solid runoff and allow them as individual outputs.

FinnHeu:
I separated the solid and the liquid freswater runoff and allowed them as individual outputs in the oIFS coupled case.


FinnHeu:
@JanStreffing: I branched off bugfix/awicm3-wind-coupling, which was merged to main in #791, for this feature. This is the last version of FESOM that I got AWI-CM3 to work with.

PR #824, opened 2025-11-17
DARS mesh namelist

TunaHim:


TunaHim:
I have now only kept files with significant changes.

JanStreffing:
Checked these against our CMIP7 HR configuration. `&param_tke` matches what CMIP7 uses exactly, which is the block that counts since CMIP7 runs `mix_scheme = 'cvmix_TKE'`, and `namelist.tra_DARS` lines up as well.

A few things we should adjust on our side before merging:

- `visc_gamma2 = 0.0285` in `namelist.dyn_DARS`, where both CMIP7 and the repo default use `0.285`. Looks like a decimal slip. Harmless while `opt_visc = 7`, since gamma2 only applies to easy backscatter, but worth correcting.
- `idemix_surforc_file` points at `/work/ollie/...`, which no longer exists, and the bottom forcing entries point at albedo paths. CMIP7 and the repo default use `/pool/data/AWICM/FESOM2/FORCING/IDEMIX/` for the surface file. The matching `idemix_surforc_vname` and `idemix_botforc_vname` are absent too.
- `AB_order` is unset in both files so it falls back to 2, while CMIP7 pins 3 for dynamics. Worth deciding what we want for this mesh.

There is also no `namelist.oce_DARS`, so nothing selects the mixing scheme and the IDEMIX and tidal blocks may never be read.

None of this is a problem with the contribution, it is mostly paths that have moved since it was opened.

JanStreffing:
Superseeded by #1004 Because I could not push to the fork here.

PR #826, opened 2025-11-20
Modify EVP settings for improved stability

patrickscholz:
Updated EVP solver type and stability constants for mEVP. change back to whichEVP=1 (mEVP) and adapt alpha, beta parameters accordingly


JanStreffing:
Also `alpha_evp      = 500` is not ok for core2 or core3. That is a value for something like DARS2

JanStreffing:
Closed needs to be revisited. @patrickscholz

PR #832, opened 2025-12-03
0D output and ldiag_cmor

JanStreffing:
This PR is a reimplementation of the cmor diagnostics from FESOM1.4


JanStreffing:
I ok, I found nothing. Ready for a test, @nwieters

PR #833, opened 2025-12-05
Reproducibility bug fix for IFS runs with frequent restarts

dsidoren:
Sea-ice surface temperature (and, less importantly, sea-ice albedo) were not properly restarted and were instead initialized with default values (265.15 K for the SIT). Each subsequent restart introduced an additional perturbation, which cumulatively resulted in an artificial cooling trend.

You can continue using your old run, as the new variables are declared optional. Just restart using NetCDF.


JanStreffing:
That's rather horrible...

Issue #854, opened 2026-02-16
Cavity freshwater/heat flux do not match


FinnHeu:
Hi there, 

I recognized the freshwater flux and the heat flux computed from the 3 equation model Holland & Jenkins in `cavity_param.F90` do not match. When adding some debugging write statements
```
!for saving to output:
net_heat_flux(node)=-heat_flux(node)   ! positive down
fresh_wa_flux(node)=-water_flux(node)

write(*,*) 'heat_flux', heat_flux(node)
write(*,*) 'water_flux from heat_flux', heat_flux(node) / (1000._WP * L) ! Convert heat flux [W] to freshwater flux [m/s]
write(*,*) 'water_flux', water_flux(node)
write(*,*) 'ratio', (heat_flux(node) / (1000._WP * L)) / water_flux(node)
```

I see a constant ration of ~1.13.

```
heat_flux  1.969518161113549E-002
water_flux from heat_flux  5.896760961417812E-011
water_flux -5.198769569582617E-011
ratio  -1.13426088279024     

heat_flux  8.249738257635762E-002
water_flux from heat_flux  2.469981514262204E-010
water_flux -2.177658007771997E-010
ratio  -1.13423756413859     

heat_flux  1.854380861213097E-002
water_flux from heat_flux  5.552038506626039E-011
water_flux -4.895161638870865E-011
ratio  -1.13418900461614     
```

The constant ratio of 1.13 points towards a missing density factor since rhor = rhoi/rhow ≈ 920/1028 ≈ 0.895, so 1/rhor ≈ 1.117 ≈ 1.13. 

I think the issue is a missing density factor in the water flux calculation:
```
heat_flux(node)  = rhow*cpw*gat*(tin-tf)      ! [W/m2]  ! positive for upward
water_flux(node) = gas*(sf-sal)/sf            ! [m/s]  <--- Missing factor here
```

It should be 
```
water_flux(node) = (gas/rhor)*(sf-sal)/sf            ! [m/s]  
```
which is already defined as `ep5 = gas/rhor` in the code but not used.

With this fix the ratio is reduced to ~1.01, almost perfectly 1.
The debug output reads:
```
heat_flux  0.185143478066233     
water_flux from heat_flux  5.543217906174626E-010
water_flux -5.457302525593441E-010
ratio  -1.01574319550332     
```

Cheers
Finn

FinnHeu:
#855

PengyangSong:
[...]

PR #855, opened 2026-02-16
Update water_flux calculation in cavity_param.F90

FinnHeu:
Added density factor in freshwater flux calculation. #854


FinnHeu:
I think the checks are failing because the results with the updated melt rates do not match the "assertation results" anymore since the fluxes changed. The updated ones are however the correct ones. 

How can these be updated @suvarchal ?

PR #862, opened 2026-02-24
Icepack CI

JanStreffing:
Activates what we had already for Icepack CI.


JanStreffing:
Some things I had to change in the CMakeLists file should be double checked. @patrickscholz

PR #865, opened 2026-03-02
hosing experiments subroutines

mzapponi:
I am including here the two subroutines I setup with the help of Dima and Patrick to add freshwater anomaly around the coasts of Antarctica (with different masks options), at the surface or at specific depths. 
I also included a passive tracer where the freshwater anomaly is imposed (tra_304) and it is possible to modify the temperature field to take into account the latent heat of fusion corresponding to the freshwater flux. 

In this case I use "linfs" as vertical coordinate so for the surface I modify the virtual salinity flux instead of imposing an actual freshwater volume.

One thing that would make it easier to use would be for sure to add two flags in the config files to call the experiments without having to do it manually and then recompile (but taking into account that part of it is only for linfs), I didn't do it because I had to modify manually the flux amount that was time-dependent so in the end I left it like this. 
I hope this is a good start at least for now.


JanStreffing (changes requested):
Thanks for putting this up @mzapponi, the masking logic and the volume-weighted vertical distribution look well thought through. The blocker at the moment is that the code runs in every simulation rather than only in hosing experiments, so it needs to become configurable before it can go in. Three things must change, then a few smaller ones.

**1. Duplicated surface boundary condition, `oce_ale_tracer.F90:1376-1384.** The line after the `end if` is unconditional:

```fortran
if (tracers%data(tr_num)%ID==304) then
    do nz=nzmin,nzmax
        tr(nz) = tr(nz) + bc_surface(...)
    end do
else
    tr(nzmin) = tr(nzmin) + bc_surface(...)
end if
 !for surface freshwater exp
tr(nzmin) = tr(nzmin) + bc_surface(...)
```

So for temperature, salinity and every other tracer the surface flux is added twice per timestep. That last line needs to be deleted and only the if/else kept.

2. Both routines need a namelist switch. fw_surf_anomaly is called from oce_fluxes and fw_depth_anomaly from fesom_runloop, neither guarded, both with hSv=0.1 hard-coded on the line above. Something like a &hosing group would cover it:

```fortran
&hosing
use_hosing       = .false.
hosing_mode      = 'surface'   ! 'surface' or 'depth'
hosing_hSv       = 0.1         ! Sv
hosing_lat_max   = -60.0       ! degrees
hosing_mask      = 'coastal'   ! 'coastal' or 'south_of_lat'
hosing_smooth_n  = 8           ! 0 disables the smoothing pass
hosing_nzmin     = 16
hosing_nzmax     = 19
hosing_latent    = .false.     ! remove latent heat of fusion
hosing_fw_temp   = .false.     ! add meltwater at freezing point
/
```

with the calls becoming if (use_hosing .and. hosing_mode=='surface') call fw_surf_anomaly(...) and likewise for depth. That also removes the hard-coded -60._WP and the nzmin=16, nzmax=19 level range, and turns the commented-out variants inside both routines into runtime options rather than edit-and-recompile toggles.

3. CASE (304) overwrites itself, `oce_ale_tracer.F90:2081-2082.

```fortran
bc_surface = dt*(hosing_flux3D(nzmin,n))
bc_surface = dt*(hosing_flux(n))
```

The second assignment kills the first, so the depth path is currently dead. It should select on the mode instead. Note also that mask3D already carries a factor dt from its construction in fw_depth_anomaly, so the 3D branch would double-count dt once it is re-enabled. Worth settling which of the two carries dt and saying so in a comment.

Smaller points

spar(100) is fixed size but indexed spar(:mesh%nl), which overflows on any mesh with more than 100 levels, so it should be sized from mesh%nl. fw_depth_anomaly has no interface block, unlike fw_surf_anomaly which you added to oce_fluxes_interface. The hfh and hfh3D streams are declared with unit W where the quantity is W/m2. The commented-out write(*,*) diagnostics would be better behind flag_debug than left as comments.

Suggested acceptance test

With use_hosing = .false. the CI truth values should come out bit identical to main. That is the direct check that the code no longer perturbs ordinary runs, and it would also have caught the duplicated bc_surface line. After that, one short run with hosing on where the area integral of hosing_flux comes back equal to hosing_hSv would confirm the normalisation.

JanStreffing:
@mzapponi I have pushed a commit to this branch that makes the hosing opt-in, which should fix the CI failures Patrick flagged.

Two things were causing them. Both hosing routines were called unconditionally with `hSv=0.1` hardcoded, so every run including all the default test cases received a 0.1 Sv Antarctic freshwater anomaly. Separately, the surface boundary term in `oce_ale_tracer.F90` was applied twice for every tracer, because the new `if/else` on tracer ID 304 was followed by an unconditional repeat of the line it replaced. That second one shifts results for temperature and salinity in every configuration, hosing or not. In `bc_surface`, `CASE (304)` also assigned twice, so the depth expression was always overwritten by the surface one. Both read like the manual switching you described in the PR text, with the alternative left uncommented rather than removed.

These are now driven from `&run_config` instead of by editing the source:

```
use_hosing   = .false.   ! off by default
hosing_mode  = 'surf'    ! 'surf' or 'depth'
hosing_hSv   = 0.0       ! magnitude [Sv]
```

I also moved `fw_surf_anomaly` and `fw_depth_anomaly` into `src/oce_hosing.F90` with a proper interface module. `fw_depth_anomaly` had no interface at all before, so its arguments were never checked at the call site in `fesom_module`.

One thing worth checking, since I had to make a judgement: I treated surface and depth as mutually exclusive modes. If you ever ran a configuration with both applied at once, that would need two independent flags instead, and my version changes what that experiment does.

JanStreffing (approved):

PR #869, opened 2026-03-10
fix: set YAC field timestep properly

nils3er:
Its not fully clear yet why this setup worked before and why yac hasn't complained.

@sveta-loza and sebastian are checking if the fix work.


sveta-loza (approved):

nils3er:
Ready! Sveta and Sebastian tested it.

PR #873, opened 2026-03-26
Stop model when using use_cavity=false with cavity mesh

### LINKED ISSUE #872: Cavities are treated as open ocean when use_cavity=.false.
When using a fesom mesh which contains cavity files (e.g. CORE2ice) and running the model with `use_cavity=.false.` in namelist.config the model will treat the cavity grid cells as open ocean without making the user aware of this. Users might think they can simply switch of the cavities by setting `use_cavity=.false.`. As it is certainly unintended to replace cavity cells with open ocean cells I suggest raising an error and stopping the model during the initialization when the mesh folder contains cavity files like cavity_elvls.out and `use_cavity=.false.` in namelist.config.

FinnHeu:
Raise an error and stop the model when the mesh folder contains cavity mesh files (cavity_elvls.out) but the user set `use_cavity=.false.` in namelist.config to prevent cavity grid cells being unintendedly treated as open ocean. #872 

```
____________________________________________________________________
  0:   ERROR: use_cavity=.false. but mesh contains cavity_elvls.out       
  0:          This mesh was prepared for cavity-resolving simulations.    
  0:          Running with use_cavity=.false. will treat cavity cells     
  0:          as open ocean, which is likely unintended.                  
  0:                                                                          
  0:          --> Either set use_cavity=.true. in namelist.config to      
  0:              enable cavities, or use a corresponding mesh            
  0:              without cavity files (e.g. CORE2 instead of CORE2ice).  
  0:  ____________________________________________________________________
  ```


JanStreffing (approved):

JanStreffing:
Mhm got a crash. I'll have a look.

JanStreffing:
Quite funny actually... Your check found a bug in our CI test.

JanStreffing:
The Icebergs CI failure here is this check doing its job rather than a regression: setups/test_pi_icebergs/setup.yml uses mesh: test_cavity (which ships cavity_elvls.out) together with use_cavity: False, so that test has been integrating the shelf cavity as open ocean from z=0, and its fcheck truth values encode exactly that.

Worth noting the iceberg code is itself cavity-aware — icb_coupling.F90:236 and icb_modules.F90:141 both branch on use_cavity. So those paths are currently dead in CI, which is arguably the combination most worth testing given icebergs calve from shelves.

The fix is probably use_cavity: True in that setup rather than an escape hatch here: same mesh as test_pi_cavity, and it needs regenerated truth values.

PR #876, opened 2026-04-02
Origin/tmp/apply patches for icb v3

ackerlar:
Some fixes for the iceberg module regarding:
 - lock-in mechanism in sea ice
 - vertical averaging and integration of temperature and salinity
 - heat fluxes feedback from icebergs to ocean


ackerlar:
can be merged from my side

JanStreffing (approved):

PR #880, opened 2026-04-09
Fix fesom.clock not updating when short segments cross year boundaries

### LINKED ISSUE #879: fesom.clock issue with short segment runs crossing year boundaries
There is an issue with fesom.clock when running FESOM2 in short segments (30 days) that cross year boundaries. I am working with the Soufflet toy ocean setup (365-day calendar, include_fleapyear=.false.).

**The problem:** 
When a 30-day segment crosses a year boundary (e.g., day 360 to day 390 of year 1975), the clock file is not updated correctly by clock_finish in gen_modules_clock.F90, and no new-year restart directory (e.g., fesom.1976.oce.restart) is created. Subsequent segments then start from the wrong time, effectively repeating the same simulation.

Within the same year say 1975 the clock works correctly, that is the clock advances properly from day 30 to day 360. But with a 30 day run which requires moving from day 360 → 390, crossing into year 1976:
 
- FESOM2 runs and produces output (split across [u.fesom.1975.nc] and [u.fesom.1976.nc]
- However, fesom.1976.oce.restart is never created
- fesom.clock remains at day 360, year 1975
- All subsequent cycles repeat from day 360 with identical results

**Possible Root cause (from gen_modules_clock.F90):**  (I think!)
In clock_finish (line 183), the year-rollover special case only triggers when the run ends exactly at the last day of the year:
 
    if ((dum_daynew==ndpyr) .and. (dum_timenew==86400._WP)) then
 
For a 30-day segment starting at day 360, the run ends at day 390. The clock subroutine (line 46) does handle the rollover during the run (daynew resets to 1, yearnew increments to 1976 when daynew > 365). But clock_finish writes yearold on line 1, which by this point has also been updated to 1976 through the timestepping loop (yearold=yearnew at line 35 of subroutine clock). So the written clock may not correctly reflect the transition.
 
Additionally, clock_finish writes the restart to ResultPath/runid.clock, but the restart directory naming (fesom.YYYY.oce.restart) appears to only create a new year's directory when a full year has been completed in one continuous run, not when short segments cross the boundary.

**Summary of the issue:** 
clock_finish and the restart I/O cannot handle arbitrary short segments crossing year boundaries without requiring the segment to end exactly at day 365.

JanStreffing:
Always write restart files and update the clock at the end of each run segment by adding (istep==ntotal) fallback to all restart type checks in write_initial_conditions. Previously only raw restarts under __ifsinterface had this safeguard, so yearly restart schedules with sub-yearly segments would leave the clock and restart files stale when the segment ended mid-year after crossing a year boundary.

Also pass the actual segment end step (ntotal) from fesom_runloop instead of f%total_nsteps, so the end-of-segment detection is correct for coupled setups where fesom_runloop is called multiple times.

Fixes #879

### DIFF
--- a/src/fesom_module.F90
+++ b/src/fesom_module.F90
@@ -768,7 +768,7 @@ contains
 #if defined (FESOM_PROFILING)
         call fesom_profiler_start("restart")
 #endif
-        call write_initial_conditions(n, nstart, f%total_nsteps, f%which_readr, f%ice, f%dynamics, f%tracers, f%partit, f%mesh)
+        call write_initial_conditions(n, nstart, ntotal, f%which_readr, f%ice, f%dynamics, f%tracers, f%partit, f%mesh)
 #if defined (FESOM_PROFILING)
         call fesom_profiler_end("restart")
 #endif
--- a/src/io_restart.F90
+++ b/src/io_restart.F90
@@ -515,24 +515,23 @@ subroutine write_initial_conditions(istep, nstart, ntotal, which_readr, ice, dyn
   ctime = timeold + (dayold - 1.0_WP) * 86400.0_WP
   
   ! Check whether restart will be written
-  is_portable_restart_write = is_due(trim(restart_length_unit), restart_length, istep)
-  
+  ! Always force a restart at the end of the run segment (istep==ntotal) to ensure
+  ! the clock file and restart files are up to date for the next segment, even when
+  ! short segments cross year boundaries without aligning with the restart schedule.
+  is_portable_restart_write = is_due(trim(restart_length_unit), restart_length, istep) .OR. (istep==ntotal)
+
   ! Should write core dump restart?
   if(is_portable_restart_write .and. (raw_restart_length_unit /= "off")) then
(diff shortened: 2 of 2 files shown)


JanStreffing:
seems to have fixed https://github.com/FESOM/fesom2/issues/879

Dr-Mansi-Singh:
> seems to have fixed #879

Forgot to mention that now with these updates, the restart folders seem to save the data both at the start and last timestep of the run segment. Earlier only the last time step was being saved.

Also in io_restart.F90, when raw_restart_length_unit='off' (or bin_restart_length_unit='off'), the code enters the else branch where the condition (istep==ntotal) forces is_raw_restart_write=.true. at the final timestep of every run.
This causes FESOM2 to attempt writing raw/binary dump files (e.g., fesom_raw_restart/np60/00.dump) even when these restart types are explicitly disabled. If the directory does not exist, the model crashes with forrtl: severe (29): file not found. Same issue applies to is_bin_restart_write.

patrickscholz (approved):

PR #884, opened 2026-04-20
Changes for XIOS3

JanStreffing:
Split xios_initialize (new io_xios_init_client) from xios_context_initialize and call the client stage right after OASIS init (NEMO pattern). Running both back-to-back deadlocked the XIOS server in createServerContextIntercomm on the EC-Earth v4.1.7 XIOS (dd74653e); interleaving par_init / mesh setup lets the server finish registering first.

Under __oifs, move io_xios_close + par_ex to the very end of fesom_finalize so no Fortran stats / profiler work runs after oasis_terminate finalizes MPI_COMM_WORLD. Non-__oifs ordering unchanged.

### DIFF
--- a/src/fesom_module.F90
+++ b/src/fesom_module.F90
@@ -166,6 +166,16 @@ contains
         call cpl_yac_init(f%partit%MPI_COMM_FESOM)
 #endif
 
+#if defined (__XIOS)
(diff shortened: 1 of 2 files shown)


JanStreffing:
XIOS 2.5 proved faster. Not upgrading to v3.0

PR #886, opened 2026-04-21
Add ability to perturb fesom2 for cold run or at restart

sebastianbeyer:
Done by @suvarchal in some other project. We are using it in the destine storylines, so it would be good to have it in main as well.


JanStreffing:
Sounds useful. Would be nice to be able to controll from namelist. Seems its all there but commented out.

sebastianbeyer:
Yes, it's commented out, to have it deactivated by default. We can also uncomment it and set 
`lperturb       = .false.`. That should do the same.

suvarchal (approved):

PR #888, opened 2026-04-21
Fesom growable io stream

JanStreffing:
@chrisdane Can you try locally merge this branch into your AWI-ESM3 recom branch?
I think this should solve the io_meandata issue. If so, let us now.


JanStreffing:
Oh this branch was not from main. sorry.

PR #890, opened 2026-04-23
Add namelist-configurable call frequency for dMOC diagnostic

JanStreffing:
diag_densMOC is called every timestep by default, adding ~43% overhead per step. Two new &diag_list parameters control this:

  dmoc_call_freq      = 1   ! default: every step (unchanged behaviour)
  dmoc_call_freq_unit = 's' ! 's'=steps, 'h'=hours, 'd'=days, 'm'=months

A new module function dmoc_is_due(mstep) dispatches to the existing step_event/hourly_event/daily_event/monthly_event helpers. Both the density_dMOC computation in pressure_bv and the diag_densMOC call in compute_diagnostics are gated by it. The dVdT divisor is corrected to use the actual time interval between calls (dmoc_dt_factor * dt).

### DIFF
--- a/src/gen_modules_diag.F90
+++ b/src/gen_modules_diag.F90
@@ -32,7 +32,8 @@ module diagnostics
             dvd_SD_chi_dif_ve, dvd_SD_chi_dif_sbc, dvd_xdfac,                                                     &
             ldiag_uvw_sqr, uv2, wvel2,                                                                            &
             ldiag_trgrd_xyz, trgrd_x, trgrd_y, trgrd_z,                                                           &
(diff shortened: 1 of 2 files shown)


patrickscholz (approved):

patrickscholz (approved):

patrickscholz:
did you check if the mean in the io_stream ist still correctly computed if not accumulate every time-step like we usually do

JanStreffing:
> did you check if the mean in the io_stream ist still correctly computed if not accumulate every time-step like we usually do

I will check.

PR #893, opened 2026-04-24
Fix restart written every coupling step under __ifsinterface

JanStreffing:
PR #880 (commit 28e51ddf) changed write_initial_conditions's segment-end marker from f%total_nsteps to the runloop-chunk end (ntotal). In standalone FESOM the two are identical (fesom_runloop is called once with all steps and from_nstep=1), so the year-boundary fallback from that PR still fires. But under __ifsinterface, fesom_runloop is called per IFS coupling step with a small substeps chunk, so ntotal == istep at the end of every call, forcing a restart write every coupling timestep.

Pass f%total_nsteps again so the (istep==ntotal) fallback only fires at the namelist-defined segment end. Standalone behavior is unchanged; under __ifsinterface FESOM never organically reaches f%total_nsteps (IFS drives the loop), so the fallback correctly stays silent and restarts revert to pure is_due cadence.

Preserves the fix for #879.

### DIFF
--- a/src/fesom_module.F90
+++ b/src/fesom_module.F90
@@ -788,7 +788,14 @@ contains
 #if defined (FESOM_PROFILING)
         call fesom_profiler_start("restart")
 #endif
-        call write_initial_conditions(n, nstart, ntotal, f%which_readr, f%ice, f%dynamics, f%tracers, f%partit, f%mesh)
+        ! Pass f%total_nsteps (namelist run length) as the segment-end marker
+        ! rather than the runloop-chunk end (ntotal). In standalone FESOM they
+        ! are identical (fesom_runloop is called once with all steps), so the
+        ! year-boundary fix from PR #880 still fires. Under __ifsinterface,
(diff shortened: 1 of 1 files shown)


sebastianbeyer:
That seems to mostly do it, but we still write netcdf and binary outputs as well at the end of the run, which we do not want to. If I set 
```
restart_length_unit='off' !output period: y,  d, h, s
raw_restart_length_unit='off' ! e.g. y, d, h, s, off
```
I do get

```
    1 : sys-2 : UNRECOVERABLE error on system request
        No such file or directory
        Encountered during an OPEN of unit -21387
        Fortran unit -21387 is not connected
```

JanStreffing:
I merged https://github.com/FESOM/fesom2/pull/887 recently, and after i made this branch i merged from main. Do you have the commit from that branch locally?

sebastianbeyer:
I have that one, but it's only about bin and raw, netcdf restarts are always being written. I'll try to address it

sebastianbeyer (approved):

PR #901, opened 2026-05-01
Drive FESOM xios_send_field cadence from XIOS XML via xios_field_is_active

JanStreffing:
## Summary

When `__XIOS` is on, FESOM today fires `xios_send_field` on every model step for every registered stream (the `freq=1, unit='s'` registration in `io_meandata.F90:ini_mean_io`). XIOS then accepts every sample, runs it through the `temporal_filter`, and gates output internally based on `output_freq`. The XIOS pipeline cost (`Streamfilter::inputField`, `Temporal filter::functor`, `Streamfilter::downstream`, `MPI_Isend`) is paid every step on every rank for every active field.

This PR matches the OpenIFS pattern (`arpifs/module/suxios.F90`) and asks XIOS — once per field per step — whether it actually wants a sample right now. The cadence policy moves from FESOM Fortran into the XIOS XML (field_def `freq_op` per file). With `freq_op` set to match `output_freq` (1d / 1mo), `xios_field_is_active` returns `.true.` only at the output boundary, so the heavy XIOS-side work runs once per output cycle instead of every model step.

Net effect (HR config, ratio=50/pools=2, dt=360):

| metric                     | before    | after        |
| --                         | --        | --           |
| streamData calls / rank/mo | ~558,000  | ~470 (1200×) |
| Streamfilter inputField    | 22.8 s    | 0.0 s        |
| Temporal filter functor    | 24.0 s    | 0.0 s        |
| Streamfilter downstream    | 37.0 s    | 4.9 s        |
| `runtime output` mean      | 228 s     | 90 s         |
| Wall                       | 28:54     | 25:32 (-3:22)|

CMIP7 conformance preserved: `cell_methods` stays `"time: mean"` because XIOS's `operation="average"` over a single sample reduces to identity and writes the FESOM-internal monthly/daily mean produced by `update_means`. As a side effect, dMOC density-class fields (`U_rho_x_DZ`, `std_dens_*`, …) that previously wrote near-fill garbage now contain correct physical values — the every-step send pattern was averaging
mid-recompute states.

### DIFF
--- a/src/fesom_module.F90
+++ b/src/fesom_module.F90
@@ -847,7 +847,10 @@ contains
     use mpp_io
 #endif
     ! EO parameters
-    real(kind=real32) :: mean_rtime(15), max_rtime(15), min_rtime(15)
+    ! 1..15 are the existing FESOM timers (ocean components, ice, output, etc.)
+    ! 16..20 are the io_meandata sub-decomposition: update_means, streamloop,
+    ! pack, mask, xsend (see rtime_om_* in io_meandata.F90).
+    real(kind=real32) :: mean_rtime(20), max_rtime(20), min_rtime(20)
     integer           :: tr_num
     
     ! Start finalization profiling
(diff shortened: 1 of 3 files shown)


JanStreffing:
Ready for review. ty! :)

patrickscholz (approved):

PR #902, opened 2026-05-01
ci: switch to FESOM-org docker image + add fesom2_xios end-to-end run-test

JanStreffing:
- Switches the build-test workflow container from the externally-owned `ghcr.io/suvarchal/fesom2-ci:latest` to the FESOM-org-owned `ghcr.io/fesom/fesom2_docker:fesom2_ci-master`, built from [FESOM/FESOM2_Docker#3](https://github.com/FESOM/FESOM2_Docker/pull/3). Same toolchain (ubuntu 22.04 + GCC + OpenMPI + NetCDF/HDF5 + LAPACK + OASIS3-MCT prebuilt at `/oasis`); the recipe was reverse-engineered from suvarchal's image so behavior is identical for the existing five matrix cells (default, coupled, coupled_yac, recom, ifs_interface).
- Adds a new `fesom2_xios.yml` end-to-end run-test mirroring the structure of `fesom2_recom.yml` / `fesom2_cavities.yml`. It runs inside `ghcr.io/fesom/fesom2_docker:fesom2_test_refactoring-master` (XIOS 2.5 install at `/xios` added in [FESOM2_Docker#5](https://github.com/FESOM/FESOM2_Docker/pull/5); attached-mode finalize segv fixed in [#6](https://github.com/FESOM/FESOM2_Docker/pull/6)), compiles FESOM with `./configure.sh ubuntu -DFESOM_WITH_XIOS=ON`, runs `mkrun pi test_pi -m docker`, stages a minimal standalone `iodef.xml` (`using_server=false`, no OASIS) plus a 6-field `file_def_fesom.xml` (`sst`, `a_ice`, `temp`, `salt`, `u`, `v` — matching the field set of the standard `test_pi` bit-identical check), runs `./job_docker_new`, and verifies all six XIOS-output netcdf files were produced per rank.
- Aligns `fesom2_openmp.yml` with the other run-test workflows by switching its container tag from `fesom2_test_refactoring-nightly` to `-master`. The nightly tag is only refreshed by a monthly cron, which let it lag behind master on the recent setuptools / pkg_resources fix from [FESOM2_Docker#4](https://github.com/FESOM/FESOM2_Docker/pull/4).
[...]

PR #903, opened 2026-05-03
build: integrate CVMix via FetchContent instead of download script

sebastianbeyer:
We do need an offline-functionality for building on MN5, so I took [this inspiration](https://github.com/FESOM/fesom2/pull/818), read a bit into cmake and (together with claude), I came up with this simplification / extension:

Replace the custom download_cvmix.sh + add_custom_command + IMPORTED-target plumbing with FetchContent_MakeAvailable. CVMix is now built as a regular CMake subproject inside the build tree (no source-tree pollution, no separate sub-cmake invocation).

For builds on internet-restricted machines, point at a manually placed checkout with 
`-DFETCHCONTENT_SOURCE_DIR_CVMIX=/path/to/CVMix-src` 
and pass 
`-DFETCHCONTENT_FULLY_DISCONNECTED=ON`.


sebastianbeyer:
hm... looks like the issue is in the ci image in this case

sebastianbeyer:
ah, should be okay, once #902 is merged

src/CMakeLists.txt
```diff
@@ -199,65 +199,30 @@ target_include_directories(${PROJECT_NAME} PUBLIC $<BUILD_INTERFACE:${CMAKE_Fort
...
-   add_dependencies(cvmix_external  cvmix_autobuild)
-   
-   # Link CVMix library to fesom, build chain becomes: fesom → cvmix_external → cvmix_autobuild
-   target_link_libraries(${PROJECT_NAME} PRIVATE cvmix_external)
-   
+
+   include(FetchContent)
+   set(CVMIX_BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
+   set(CVMIX_BUILD_DRIVER      OFF CACHE BOOL "" FORCE)
+   FetchContent_Declare(cvmix
```
patrickscholz, line 214: How does cmake know where to iinstall/put the CVMix files ? Or where does the clones repo end up ?

patrickscholz:
@suvarchal: We also had this improvement to the CVMix process, where suvi used a buildin Cmake function or something like that, https://github.com/FESOM/fesom2/pull/818/, but apparently we never merged it. I think problem was that it was dependent on a rather actual CMake version, that not everywhere might be available. Sebastian solution looks very short and clean, but i honestly have no ideas about its pro/cons ...

src/CMakeLists.txt
sebastianbeyer: The repo will be put into `build/_deps/cvmix-src` and then built in `build/_deps/cvmix-build`.  It will be added to the main fesom build through `add_subdirectory()`, that is why cmake can then link it.

sebastianbeyer:
yes, that is the original PR that I found and that made me come up with this one. using FetchContent simplifies this a lot and does also handle the offline part very nicely. 
It was added in Cmake 3.11 originally and the way we use it was added in 3.14. Fesom already required cmake 3.16. So that should be fine.

suvarchal:
> yes, that is the original PR that I found and that made me come up with this one. using FetchContent simplifies this a lot and does also handle the offline part very nicely. It was added in Cmake 3.11 originally and the way we use it was added in 3.14. Fesom already required cmake 3.16. So that should be fine.

wonderful, thanks for finishing this!

patrickscholz (approved):

PR #908, opened 2026-05-07
ice dyn: fix strength_ice diagnostic (zero/stale for mEVP/aEVP, off by 2x for EVP)

### LINKED ISSUE #907: strength_ice diagnostic is broken: zero or stale for mEVP and aEVP, off by factor 2 for EVP
The `strength_ice` output stream has two diagnostic-level bugs. The simulated dynamics are unaffected; only the NetCDF field labelled "ice strength" is wrong.

First, when the namelist selects mEVP (`whichEVP = 1`) or aEVP (`whichEVP = 2`), `ice%work%ice_strength` is never written. Both solvers fold the strength into local intermediates inside the rheology iteration and never persist the value. In `EVPdynamics_m` (`src/ice_maEVP.F90`) the strength enters a precomputed `pressure_fac(el) = det2 * P* * h * exp(-C*(1-A))`, after which the bare strength only reappears as a local scalar `pressure = pressure_fac(el)/(delta+delta_min)` inside the substep loop. In `EVPdynamics_a`, `stress_tensor_a`, and `find_alpha_field_a`, the strength only ever exists as a local scalar that is recomputed each substep. The `def_stream` registration in `src/io_meandata.F90` is unconditional on the active solver, so the diagnostic file silently contains zeros or whatever was left in memory from initialisation.

Second, even with the standard EVP, the value stored in `ice%work%ice_strength` is half the canonical Hibler (1979) ice strength. The relevant lines in `EVPdynamics` in `src/ice_EVP.F90` (both ALE branches) are

```fortran
ice_strength(el) = ice%pstar*msum*exp(-ice%c_pressure*(1.0_WP-asum))
ice_strength(el) = 0.5_WP*ice_strength(el)
```

The first line is the canonical strength `P = P* * h * exp(-C*(1-A))` from Hibler (1979). The second line halves it, because the EVP iteration in `stress_tensor` consumes the strength as `zeta = ice_strength*delta_inv` (giving `P/(2*delta)`) and as `r1 = zeta*(eps11+eps22) - ice_strength*Tevp_inv` (giving the Hunke and Dukowicz (1997) replacement pressure term `-P/(2*T)`). So the 0.5 is folded into the stored value as an inlining optimisation for the EVP substep. The variable is then misnamed: what `ice_strength` actually holds is pressure `P/2`, not the strength `P`. Bouillon et al. (2013), whose discretisation FESOM follows, write the same formulation with the 0.5 explicit at the point of use.

The convention the rest of the community uses is the unhalved `P`. CICE / Icepack expose `strength` in N/m as the full Hibler (1979) value, and that is the convention used throughout the mEVP and aEVP papers (Kimmritz et al. 2015; Kimmritz et al. 2017). A user comparing FESOM `strength_ice` against CICE, against published values, or against the icepack `strength` field already routed through `def_stream` for the icepack build path will see a factor of 2 mismatch with no warning.

Suggested fix, in two parts:

1. Populate `ice%work%ice_strength` in mEVP and aEVP using the same expression already evaluated by the rheology, so the diagnostic stops being silently zero. For mEVP this is one extra line in the existing `pressure_fac` precompute. For aEVP it needs a small precompute pass at the top of `EVPdynamics_a`. Match whatever convention EVP currently stores so the three solvers agree and EVP results stay bit-identical.

2. Fix the convention so `strength_ice` reports the canonical `P`, matching Hibler (1979), Hunke and Dukowicz (1997), CICE, and Icepack. The cleanest option is to move the 0.5 from storage to the use sites in `stress_tensor`. This is mathematically identical, but floating-point reassociation means EVP output may not be bit-identical to prior runs, so it should be coordinated with whoever owns rheology validation. A reproducibility-preserving alternative is to multiply by 2 only at output time.

References:
- Hibler, W. D. (1979). A Dynamic Thermodynamic Sea Ice Model. *J. Phys. Oceanogr.* 9, 815-846. https://journals.ametsoc.org/view/journals/phoc/9/4/1520-0485_1979_009_0815_adtsim_2_0_co_2.xml
- Hunke, E. C., and Dukowicz, J. K. (1997). An Elastic-Viscous-Plastic Model for Sea Ice Dynamics. *J. Phys. Oceanogr.* 27, 1849-1867. https://journals.ametsoc.org/view/journals/phoc/27/9/1520-0485_1997_027_1849_aevpmf_2.0.co_2.xml
- Bouillon, S., Fichefet, T., Legat, V., and Madec, G. (2013). The elastic-viscous-plastic method revisited. *Ocean Modelling* 71, 2-12. https://www.sciencedirect.com/science/article/abs/pii/S146350031300098X
- Kimmritz, M., Danilov, S., and Losch, M. (2015). On the convergence of the modified elastic-viscous-plastic method for solving the sea ice momentum equation. *J. Comput. Phys.* 296, 90-100. https://www.sciencedirect.com/science/article/abs/pii/S0021999115003083
- Kimmritz, M., Losch, M., and Danilov, S. (2017). A comparison of viscous-plastic sea ice solvers with and without replacement pressure. *Ocean Modelling* 115, 59-69. https://www.sciencedirect.com/science/article/pii/S1463500317300690

JanStreffing:
ice%work%ice_strength was only populated by EVP, so the strength_ice output stream was silently zero or stale whenever whichEVP=1 (mEVP) or 2 (aEVP). EVP itself stored P/2 rather than the canonical Hibler (1979) strength P = P*·h·exp(-C(1-A)), because the 0.5 factor was folded into storage as an inlining optimisation for the stress_tensor iteration; this gave a factor-2 mismatch against CICE/icepack and the literature.

EVP: move the 0.5 from storage to the use sites in stress_tensor (zeta and r1). Mathematically identical; FP-reassociation only.

mEVP: populate ice_strength in the existing pressure_fac precompute, with pressure_fac = det2 * ice_strength.

aEVP: add a once-per-timestep precompute pass at the top of EVPdynamics_a; under __icepack, use the per-node icepack strength averaged to the element, matching the rheology branch in stress_tensor_a.

All three solvers now store the canonical P.

### DIFF
--- a/src/MOD_ICE.F90
+++ b/src/MOD_ICE.F90
@@ -34,6 +34,11 @@ TYPE T_ICE_WORK
     real(kind=WP), allocatable, dimension(:)    :: sigma11, sigma12, sigma22
     real(kind=WP), allocatable, dimension(:)    :: eps11, eps12, eps22
     real(kind=WP), allocatable, dimension(:)    :: ice_strength, inv_areamass, inv_mass
+    ! strength_ice: canonical Hibler (1979) ice strength P = P*·h·exp(-C(1-A))
+    ! exposed to def_stream as the 'strength_ice' output. Distinct from
+    ! ice_strength, which stores P/2 as an inlining optimisation for the
+    ! standard EVP rheology iteration.
+    real(kind=WP), allocatable, dimension(:)    :: strength_ice
     !___________________________________________________________________________
     contains
         procedure WRITE_T_ICE_WORK
@@ -825,9 +830,11 @@ subroutine ice_init(ice, partit, mesh)
     allocate(ice%work%ice_strength(    elem_size))
     allocate(ice%work%inv_areamass(    node_size))
     allocate(ice%work%inv_mass(        node_size))
+    allocate(ice%work%strength_ice(    elem_size))
     ice%work%ice_strength= 0.0_WP
     ice%work%inv_areamass= 0.0_WP
     ice%work%inv_mass    = 0.0_WP
+    ice%work%strength_ice= 0.0_WP
 
     !___________________________________________________________________________
     ! initialse thermo array of ice derived type
--- a/src/ice_EVP.F90
+++ b/src/ice_EVP.F90
@@ -370,6 +370,7 @@ subroutine EVPdynamics(ice, partit, mesh)
     real(kind=WP), dimension(:), pointer  :: u_w, v_w, elevation
     real(kind=WP), dimension(:), pointer  :: stress_atmice_x, stress_atmice_y
     real(kind=WP), dimension(:), pointer  :: inv_areamass, inv_mass, ice_strength
+    real(kind=WP), dimension(:), pointer  :: strength_ice
 #if defined (__icepack)
     real(kind=WP), dimension(:), pointer  :: a_ice_old, m_ice_old, m_snow_old
 #endif
@@ -406,6 +407,7 @@ subroutine EVPdynamics(ice, partit, mesh)
     inv_areamass    => ice%work%inv_areamass(:)
     inv_mass        => ice%work%inv_mass(:)
     ice_strength    => ice%work%ice_strength(:)
+    strength_ice    => ice%work%strength_ice(:)
 
     !___________________________________________________________________________
     ! If Icepack is used, always update the tracers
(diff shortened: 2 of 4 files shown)


JanStreffing:
Old EVP did (0.5·P) · (1/Δ) (0.5 pre-baked at storage), new EVP does 0.5 · P · (1/Δ). Mathematically identical, IEEE-754 not bit-identical because operand ordering and rounding differ. Magnitudes match what you'd expect from a single rearranged multiplication on a typical Arctic field.

Should be easy to put in a bit identical conserving version of the fix though.

JanStreffing:
I will compute a year of ice strength output for standalone core2.

JanStreffing:
Half a year after init on a core2 mesh standalone run. Looks ok to me.

strength_ice_1958-05-31

dsidoren (approved):

PR #909, opened 2026-05-08
ice dyn: move strength_ice diagnostic to nodes

JanStreffing:
Follow-up to #908. I had developed and tested this nodal refactor while validating the previous PR but forgot to push it before merging — the smooth, coherent `strength_ice` plot I shared in that PR was produced by *this* version of the diagnostic, not the per-element version that actually got merged. So the field on `main` right now still has the per-element fragmentation issue we discussed.

This PR moves `strength_ice` from per-element storage to per-node storage so the diagnostic lines up with every other ice state variable (`a_ice`, `m_ice`, `h_ice`, `h_snow`, `thdgr*`, `dyngr*`, `apnd`, `hpnd`, `ipnd`, `uice`, `vice` — all nodal; only the genuinely elemental stress tensor components `sgm11/12/22` stay on elements).


src/ice_EVP.F90
```diff
@@ -422,19 +422,18 @@ subroutine EVPdynamics(ice, partit, mesh)
...
+    ! Computed pointwise from m_ice and a_ice so it lines up with every other
+    ! ice state diagnostic (a_ice, m_ice, ...) which are also nodal. The
+    ! rheology storage ice_strength is per-element and stays untouched, so
+    ! EVP results are bit-identical to before this fix.
+!$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(n)
+    do n = 1, myDim_nod2D
+        strength_ice(n) = 0.0_WP
+        if (ulevels_nod2D(n) > 1) cycle
+        if (m_ice(n) <= 0._WP .or. a_ice(n) <= 0._WP) cycle
+        strength_ice(n) = ice%pstar*m_ice(n)*exp(-ice%c_pressure*(1.0_WP-a_ice(n)))
```
patrickscholz: could we agree on a common notation between ice_strength(node) and strength_ice(el) (later in the code), otherwise that might lead to some confusion. Im already confused!
maybe something like ice_strength_nod2 and ice_strength ...

On the other hand do we really need to allocate effectively twice the same variable, for the sake of RAM efficiency, could we not outsource the 0.5 factor to somewhere else (write some comment story about it) and stay with one variable  ?!

JanStreffing: Which one should we go for?

Variant A: rename and relocate the 0.5 factor, retaining both arrays. ice_strength is renamed to ice_strength_elem and continues to serve as a per-element cache for the EVP iteration. strength_ice is renamed to ice_strength_nod and provides the per-node diagnostic. Both arrays then store the canonical Hibler strength P, and the 0.5 factor is applied at the use sites in stress_tensor. The arrays differ only by their grid location.

Variant B: a single nodal array. ice_strength is removed. Only ice_strength_nod remains, storing the canonical P at each node. The EVP stress_tensor evaluates the per-element strength inline as the mean of the three nodal values and applies the 0.5 at the use site. The mEVP and aEVP solvers already evaluate strength inline and are unaffected.

In both variants the EVP path is no longer bit-identical to current main, owing to floating-point reassociation in the rheology iteration, so a refresh of the CI baseline is required.

JanStreffing:
Ready to merge from my side.
Screenshot 2026-05-17 at 17 23 57

patrickscholz (approved):

PR #911, opened 2026-05-11
REcoM as a submodule in FESOM

wiltonloch:
This is a new updated version of the work for having REcoM as a submodule in FESOM.

It uses the most updated version of REcoM in https://github.com/RECOM-Regulated-Ecosystem-Model/REcoM and will separate the concerns and developments of both models.

It will also be very valuable to facilitate rework and refactoring which are now taking place in an active natESM sprint.


JanStreffing:
Nice! Differences seem to be numerical in nature. There is a script to update the truth values in https://github.com/FESOM/fesom2/blob/recom_submodule_update/setups/update_truth_values.sh

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/recom"]
+	path = src/recom
+	url = https://github.com/RECOM-Regulated-Ecosystem-Model/REcoM_Standalone
```
ogurses: Check the URL!!

ogurses (approved):

ogurses (approved):

ogurses (approved):

wiltonloch:
Hello, the differences are due to the compilation flags which are not the same in the REcoM standalone. I've updated that in the latest version of REcoM and will update it soon here also.

JanStreffing:
> Hello, the differences are due to the compilation flags which are not the same in the REcoM standalone. I've updated that in the latest version of REcoM and will update it soon here also.

If you think you can recover bit-identicalness, all the better.

wiltonloch:
@ogurses I have updated the url. We talked about another commit of yours which was in `recom_submodule` that updated the config files, but apparently the changes there are already in the main branch of FESOM. As there were no differences between the tip of this branch and that commit for the config files.

@JanStreffing The  bit-identicalness has been restored for the CI tests. The differences were, as I mentioned, due to different compilation flags. I have fixed that, but note that on the new [CMakeLists.txt of REcoM](https://github.com/RECOM-Regulated-Ecosystem-Model/REcoM/blob/master/CMakeLists.txt#L18) these options are added just for when the build is happening inside of the CI. When running on any other system I think it will inherit the default compilation flags set by FESOM.

Issue #350, opened 2022-06-09
Using restarts instead of continuous run leads to differences


alxthor:
I am doing a (cold start) simulation with zstar on master branch. One time in one go from 1958 to 1960. The other time one year after another with in total two restarts. Subtracting both runs leads to identical results in 1958 but differences building up in the first half year of 1959 and persisting after then.
The differences of upto approx. +-2 Kelvin / +-1 psu in temp / salt are most prominent in top layers near the equator and travel westwards in waves.
@dsidoren asked me to post an issue here. I hope I am doing it correctly.

![image](https://user-images.githubusercontent.com/107220804/172959296-da6b8fd7-3b0a-409f-ad59-44153e133c89.png)
![image](https://user-images.githubusercontent.com/107220804/172959319-ce3a982b-174c-41e3-9df2-68fe33597029.png)

hegish:
@alxthor, thank you for reporting this. Could you tell us which branch and commit hash you are using? I want to try to reproduce the problem.

alxthor:
Thanks, sorry for the missing info. I am using master branch, commit e7562dcb1754181492f41e4f664886fba6160562

koldunovn:
I think it's the consequence of forcing treatment during the restart. In continuous run you interpolate forcing between timesteps, in the case of the restart, one just takes the forcing that is closest to the time step. So you end up with runs, that have a bit different forcings.

alxthor:
I see. Can that explain, why the effects are so large in magnitude and mainly occur in the equatorial region (especially mid Pacific and Malaysia) and sometimes near Antarctica?

patrickscholz:
Could you say which mixing scheme you using ?

alxthor:
I'm using KPP

patrickscholz:
"KPP" ... than it is OK i just discovered this week a small mistake in the restart of TKE+IDEMIX mixing.

alxthor:
Thank you. So you recon that this mistake could be causing this effect and fixing this could remove the large descrepancy between the runs?

patrickscholz:
No fixing this bug in TKE+IDEMIX will not affect KPP, So it will not fix your problem.

dsidoren:
> I think it's the consequence of forcing treatment during the restart. In continuous run you interpolate forcing between timesteps, in the case of the restart, one just takes the forcing that is closest to the time step. So you end up with runs, that have a bit different forcings.

I guess you are right. We need to setup a test for this.

[...]

PR #919, opened 2026-05-25
sbc: rotate initial wind/stress  coefficients on rotated grids

koldunovn:
## Summary

On a rotated grid (`rotated_grid = .true.`, e.g. the standard CORE2 setup), the **initial** surface-forcing wind/stress interpolation coefficients are never rotated into the rotated-grid frame. As a result, for the **first forcing window** of a run the wind stress applied to **both the ocean and the sea ice** stays in the geographic frame and is mis-directed by the local geo→rotated (`g2r`) angle.

## Mechanism

The `g2r` rotation of the wind/stress is applied **only** in `sbc_do` (`src/gen_surface_forcing.F90`), and only when the time-interpolation coefficients are **refreshed**:

```fortran
! sbc_do: do_rotation_wind is set ONLY inside the getcoeffld (refresh) branch
if ( ((rdate > nc_time(t_indx_p1)) .and. ...) .or. force_newcoeff) then
   call getcoeffld(fld_idx, rdate, partit, mesh)
   if ((l_xwind .and. (fld_idx==i_xwind)) .and. rotated_grid) do_rotation_wind=.true.
endif
...
if (do_rotation_wind) then        ! rotate coef_a/coef_b for wind
   ... call vector_g2r(coef_a(i_xwind,i), coef_a(i_ywind,i), ...) ...
end if
```

The cold-start initialization `nc_sbc_ini` loads the very first set of coefficients but **never rotates them**:

```fortran
! nc_sbc_ini
do fld_idx = 1, i_totfl
   call getcoeffld(fld_idx, rdate, partit, mesh)   ! load first coefficients
end do
call data_timeinterp(rdate, partit)                ! build atmdata -> NO rotation
```

At cold start, `force_newcoeff` is `.false.` (no year change) and `rdate` lies inside the first forcing interval, so the refresh condition in `sbc_do` is **not** met on the first steps → `do_rotation_wind` stays `.false.` → the rotation block is skipped → the wind used (`atmdata = rdate*coef_a + coef_b`) is the **unrotated** geographic wind. The rotation only kicks in once the run crosses the first forcing-window boundary and `sbc_do` calls `getcoeffld` again.

## Impact

- **Transient**: it self-corrects after the first forcing-window crossing (for 3-hourly JRA55 at `dt = 1800 s`, ≈ 6 steps / ≈ 3 model hours).
- **Magnitude-preserving**: a rotation is orthogonal, so `|stress|` is unchanged. This is why the bug is invisible to every magnitude-based diagnostic — `u*`, turbulent fluxes, SST/SSS/SSH, ice area/volume, ice/ocean speed *ratios* — and to multi-year climate validation. Only the wind-stress **direction** is wrong, so it perturbs the **initial ice drift and surface-current direction** for the first few steps.
- **Scope**: only rotated grids (`rotated_grid = .true.`). Non-rotated setups are unaffected.

## How it was found

Discovered while bit-comparing a C re-implementation of the FESOM2 forcing/EVP path against this Fortran code at ocean step 1 (CORE2 mesh, JRA55, dt=1800, cold start). Every EVP input was bit-identical **except** the wind-on-ice stress `stress_atmice`, which differed at **all** surface nodes. The difference was a **pure rotation**: per-node `|stress|` matched to 1.0000, while the direction differed by exactly the local `g2r` vector angle (`stress_C = g2r(stress_Fortran)`, residual 0.0 at every node). Tracing it showed the C code rotated the wind from step 1 while this Fortran code did not, isolating the missing rotation to `nc_sbc_ini`.

## Fix

In `nc_sbc_ini`, after loading the initial coefficients, apply the same `g2r` rotation that `sbc_do` already applies on refresh (guarded by `l_xwind`/`l_xstre` and `rotated_grid`). It is a faithful mirror of the existing `sbc_do` rotation block, so steady-state behaviour is unchanged; it only fixes the cold-start window.

## Validation

Compared step-1 wind-on-ice stress (CORE2, JRA55, dt=1800, cold start) against a rotated-frame reference, before vs after the fix:

| quantity | before fix | after fix |
|---|---|---|
| `stress_atmice` C-vs-ref, per node | pure rotation: `|·|` ratio 1.0000, direction off by the local `g2r` angle (median ~15°, up to 180°); max abs diff ~0.43 N/m² | `|diff|` ≤ 5e-14 (bit-match) at all 126858 nodes |
| resulting first-EVP-subcycle ice velocity | ~70% relative diff | ~4e-16 |

Multi-year climate is unchanged (the bug was a bounded transient that self-corrects after the first forcing window), confirming the fix only touches the initial-window forcing direction.


patrickscholz (approved):

koldunovn:
@patrickscholz @JanStreffing @dsidoren  - this is actually the reason for our restarts not being bit identical, guys.

JanStreffing (approved):

koldunovn:
Verified this against a restart test on Levante — it removes the wind discontinuity completely.

CORE2 (rotated) + JRA55-do + `zstar`/`cvmix_TKE`/mEVP, 512 ranks, 2 days straight vs 1 day + restart + 1 day, with `tair`/`uwind`/`shum`/`sst` written every time step:

| | wind @ first step after restart | SST @ that step | SST max over the following day |
|---|---|---|---|
| without this PR | 2.097e+01 m/s (differs at **all** 126858 nodes, mean 1.32 m/s) | 2.87e-02 K | 4.16e-01 K |
| **with this PR** | **0.000e+00** | **2.68e-04 K** | **6.92e-02 K** |

`tair` and `shum` were bit-identical in both cases — only the vector field was affected, which is what pointed at the missing rotation rather than at time interpolation in general. Without the PR the wind is wrong at every node for exactly the 2 steps between the restart and the next forcing-record boundary, then recovers on its own.

This is also the dominant term in #350 (restart vs continuous differences), where the cause had been hypothesised but never measured — details and the full test description there: https://github.com/FESOM/fesom2/issues/350#issuecomment-5209031168

JanStreffing:
Any idea why the tests dont run?

koldunovn:
The change changes model result, so tests will be failing. I will update the numbers.

sebastianbeyer:
But this does not explain the difference when running IFS-FESOM, since there you checked the flux that comes from the IFS, correct, @dsidoren ?

JanStreffing (approved):

PR #925, opened 2026-05-31
Output along ship tracks / mooring curtains

JanStreffing:
I wrote FESOM 2D curtain output. It works by providing a list of lat lon in a csv file. FESOM the computes the greatcircle connections. It desifies those lines and for each subsection picks the nearest neighbour fesom nodal center. Here is a first example of an track / curtained i picked with 7 coordiantes:

    /home/a/a270092/esm_tools/namelists/fesom2/tracks/test_track_core2_transect.csv
    # Multi-segment transect for the FESOM-2.7 ship-track / mooring curtain
    # feature (io_tracks.F90, feature/tracks branch). Same 7 vertices used by
    # track_densify_demo.py to visualize the densify-and-snap pipeline.
    #
    # Path: subpolar N. Atlantic -> Iceland -> Nordic Seas -> Fram Strait.
    # At track_resolution_km = 5.0 (default), this densifies to ~639 dense
    # samples that resolve to ~152 unique CORE2 mesh nodes.
    #
    # Format: lon, lat
    -35.0,  55.0
    -30.0,  57.5
    -25.0,  60.0
    -20.0,  62.5
    -15.0,  65.0
     -8.0,  70.0
      5.0,  78.0

grafik
grafik

One year of hourly test outputs at: /work/bb1469/a270092/runtime/awiesm3-develop/test_track_11/outdata/fesom

I have only tested the output via XIOS. Not without it.


JanStreffing:
Putting this on draft while reworking to be in line with tripyview logic.

JanStreffing:
Nodal works with new logic:
grafik

JanStreffing:
Old method vs new one:
grafikgrafik
Limitation of tripyview is that we don't do great circles. Needs more points in csv to get close. Can be improved later if desired.

JanStreffing:
Mass transport from elements look alright.
grafik

JanStreffing:
## Status of the geometry-aware redesign

**What works**

- Pure-Fortran geometry kernel `mod_tracks_geometry.F90`, ported from `tripyview/sub_transect.py` (bbox + polar widening, signed-distance line-edge intersection, alternating triangle path, cumulative great-circle distance). Uses FESOM's in-memory `mesh%edge_cross_dxdy(4, edge2D)` directly as tripyview's `edge_dxdy_l`/`_r`, no reconstruction.
- Rank-0 gather of global topology via `gather_edge`/`gather_elem`/`gather_nod`, broadcast of the small `transect_t` to all ranks. Per-rank strict-own lookups (`global → 1..myDim_*`) for nodes (`myList_nod2D(1:myDim_nod2D)`) and elements (`myInd_elem2D_shrinked`).
- **Node-based sampling** at edge crossings (`temp`, `salt`): per-edge `V·(1−t)+V·t` with weighted Allreduce; lone-endpoint wet/dry edges fall back to the surviving value instead of a biased partial sum; per-level `nlevels_nod2D` mask; NaN where neither endpoint contributes.
- **Element-based sampling** at path triangles (`u`, `v`, `vflx`, `hflx`, `sflx`): velocities rotated to geographic via `vector_r2g` at each centroid; `vflx = (u·|dy|·nvec_x + v·|dx|·nvec_y)·dz` matches tripyview's `calc_transect_Xtransp` `do_nveccs=True` formula bit-for-bit. `hflx`/`sflx` multiply `vflx` by element-averaged T/S and the standard `ρ₀·cp` / `ρ₀/1000` constants.
- **XIOS register** dispatches on var kind: node-based → `track_<name>` domain at M cells; element-based → `path_<name>` at P cells. One `fesom_tracks` file_group, one file per track.
- **Cross-check vs tripyview** on identical monthly-mean u/v: io_tracks 1-year mean = **+150.4 Sv**, tripyview = **+149.3 Sv** at the same Drake CSV. 2-year mean = **+151.0 Sv** (cold-start CORE2 LR). Spatial structure (deep-jet core, weak westward returns) matches expectations.
- 7-track smoke configs work cleanly: 4 node-based `temp` (CORE2 Atlantic→Fram transect + 3 MOSAiC ice-camp windows) + 2 element-based `u`/`v` on the transect + 1 `vflx` on Drake.
- All cleanup from the round-2 review folded in: weight-rescaled wet/dry, per-level `nz_wet` mask, MPI-safe `par_ex` instead of `abort`, `vec_autorotate`-gated rotation at the path centroid, polar bbox widening per tripyview.

**What doesn't (yet)**

- **Variable list is hardcoded**: `temp, salt, u, v, vflx, hflx, sflx`. Adding any other field (e.g., `ssh`, `w`, `bolus_*`, passive tracers) needs an explicit `case` branch + the right Fortran array reference. A generic `tracer:N` mode would unlock any FESOM tracer index in ~20 LOC.
- **Sub-segments are straight in lat-lon**, not great-circle (inherited from tripyview's planar line equation `a·lon + b·lat + c = 0`). Identical to the rhumb line for constant-lat sections, identical to the geodesic for constant-lon, but diverges meaningfully for long two-waypoint high-latitude sections. Densely sampled CSVs (e.g. MOSAiC ~220 waypoints) are unaffected. Fix is a spherical line-equation swap (~60 LOC) gated by a `track_geodesic` flag — open item, not blocking.
- `track_resolution_km` was removed — sampling cadence is mesh-driven now.
- `insert_landpts` for inserting NaN slots between consecutive boundary cuts is currently a stub; the within-segment ghost-slot logic in `add_upsection`/`add_downsection` covers the common case, but a transect that exits and re-enters the ocean repeatedly might miss a NaN gap. Easy to add when needed.
- No standalone Fortran unit harness — validation today is via running the real model and diffing against tripyview on the same NetCDF u/v output. The plan called for a 1e-10 geometry diff against tripyview; not yet wired up.

JanStreffing:
@patrickscholz Ready for review.

patrickscholz (approved):

PR #929, opened 2026-06-03
mask netcdf output by wet-cell topography instead of value==0

### LINKED ISSUE #821: fill values in 2D fields
I noticed the fill values were changed (which is awesome for the 3D fields that can run into bathymetry at deeper levels), but it has also been changed for 2D fields (I noticed this for a_ice/m_ice).

Was this changed on purpose for the 2D fields? sea ice concentration = 0 is a physically reasonable value, same for sea ice thickness. On the other hand, I see the point in using different values for ice free points depending on whether there has ever been ice at there in the past (0) or there has never been any ice there (NaN/fill_value).

However, for plotting and computations I had to do a small workaround like this in python:
```
a_ice[:] = a_ice.fillna(0) -> set all NaN values to 0 (to make open ocean show as 0)
```
then grid the data for plotting and then apply land sea mask (mask_ne from pyfesom2) to get the land values back to NaN
```
mask = pf.mask_ne(xx, yy)
a_ice_gridded[mask] = np.nan
```

without the workaround:
Image

with the workaround:
Image

### LINKED ISSUE #939: runoff output from fesom2.7-recom3.1 with AWICM forcing substantially higher than expected
When running FESOM 2.7 with AWICM forcing, the runoff output is unexpectedly high compared to (1) a run with FESOM2.1 using the same AWICM forcing and (2) the original AWICM runoff forcing. Figure attached shows the difference.

Image

What could be a possible reason for this and places in the code to inspect the same?

JanStreffing:
PR #799 set any accumulated output value below 1e-30 to _FillValue. That also masks legitimate ocean zeros: ice-free sea-ice concentration (a_ice and all other 2D ice variables) and vanishing 3D fields such as IDEMIX energy were written as NaN instead of 0, contrary to the CMIP/CMOR convention. The masking was applied uniformly to 2D and 3D fields.

Replace the value-based test with a proper wet-cell mask derived from the mesh level arrays (ulevels/nlevels for elements, ulevels_nod2D/ nlevels_nod2D for nodes). Only cells above a cavity or below the bottom topography are written as _FillValue; every wet cell keeps its averaged value, including a genuine zero. 2D fields and non-spatial vertical axes (density / ice classes) are never masked.

@chrisdane Can you check if this fixes what you flagged in #799 ?


src/io_meandata.F90
```diff
@@ -2923,46 +2178,43 @@ subroutine output(istep, ice, dynamics, tracers, partit, mesh)
...
-            !
-            ! NetCDF fill-value substitution is skipped under __MULTIO: multio
-            ! ships the array straight to GRIB and a 9.97e36 sentinel in the
-            ! data range wrecks GRIB packing precision (real ice values get
-            ! quantized to ~0). Multio gets the clean mean (zeros stay zeros).
             if (entry%accuracy == i_real8) then
-!$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(I,J)
+                nlev_loc = size(entry%local_values_r8,dim=1)
+!$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(I,J,ul_loc,kmax_loc)
                 DO J=1, size(entry%local_values_r8,dim=2)
```
patrickscholz, line 2987: Are there any performance issues when exchanging the outer and inner loops, get it better or worser?

JanStreffing:
Adding a data point in favor of this fix: #939 is another symptom of the value-based masking from #799, this time on river runoff. Same pattern as #821: a sparse 2D field where `_FillValue` lands at every node that should hold a legitimate zero, so `xr.mean()` over nodes divides by the nonzero count and inflates the apparent magnitude by `N_total / N_nonzero ≈ 2.6×`. Worth prioritizing for 2.7.1.

src/io_meandata.F90
JanStreffing: I have not seen any, but worth a check

patrickscholz (approved):

src/io_meandata.F90
```diff
@@ -2960,40 +3000,59 @@ subroutine output(istep, ice, dynamics, tracers, partit, mesh)
...
-                    DO I=1, size(entry%local_values_r8,dim=1)
 #if defined(__MULTIO)
+                    ! multio ships the clean mean to GRIB; no _FillValue substitution
+                    ! (a 9.97e36 sentinel in the data range wrecks GRIB packing).
+                    DO I=1, nlev_loc
                         entry%local_values_r8_copy(I,J) = entry%local_values_r8(I,J) /real(entry%addcounter,real64)  ! compute_means
+                        entry%local_values_r8(I,J) = 0._real64 ! clean_meanarrays - reset to 0 for next accumulation
```
suvarchal: I am bit worried about this line, why reset after we just assign?

src/io_meandata.F90
```diff
@@ -2960,40 +3000,59 @@ subroutine output(istep, ice, dynamics, tracers, partit, mesh)
...
-!$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(I,J)
+                nlev_loc = size(entry%local_values_r4,dim=1)
+!$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(I,J,ul_loc,kmax_loc)
                 DO J=1, size(entry%local_values_r4,dim=2)
-                    DO I=1, size(entry%local_values_r4,dim=1)
 #if defined(__MULTIO)
+                    ! see comment in the double precision branch above
+                    DO I=1, nlev_loc
                         entry%local_values_r4_copy(I,J) = entry%local_values_r4(I,J) /real(entry%addcounter,real32)  ! compute_means
+                        entry%local_values_r4(I,J) = 0._real32 ! clean_meanarrays - reset to 0 for next accumulation
```
suvarchal: just make it zero immediately after computing mean?

suvarchal: my bad i retract

src/io_meandata.F90
suvarchal: my bad i retract

suvarchal:
there is issue if we mask cells then global mean computation changes w.r to netcdf reading and fcheck used in CI.  this needs a little careful look once on what it means for analysis.

JanStreffing:
> there is issue if we mask cells then global mean computation changes w.r to netcdf reading and fcheck used in CI. this needs a little careful look once on what it means for analysis.

Its the opposite of what we had for #799 https://github.com/FESOM/fesom2/pull/799/changes/4efe31bd821dd93bad71a4d619f5012fd858e18f
Looks ok.

JanStreffing:
I'll simulate a test year with this.

JanStreffing:
One year eval showed the 0s are back in Sea ice over none covered regions. Carefull, there are also zeros in fields now that should perhaps but truely undefined. Like sea ice surface temp where concentration is 0.

PR #934, opened 2026-06-11
Account for additional fields from Lars iceberg fix in oifs

FinnHeu:
This fixes an oifs crash right after the first coupling in awiesm3-v3.4.2. Lars added two fields and oifs was receiving 0K for the ice temperature which caused a floating point exception.

### DIFF
--- a/src/MOD_ICE.F90
+++ b/src/MOD_ICE.F90
@@ -161,7 +161,7 @@ TYPE T_ICE
     !___________________________________________________________________________
     ! total number of ice tracers (default=3, 1=area, 2=mice, 3=msnow, (4=ice_temp)
 #if defined (__oifs) || defined (__ifsinterface)
-    integer                                     :: num_itracers=4
+    integer                                     :: num_itracers=6
 #else
 !    integer                                     :: num_itracers=3
     !------------------------------


src/MOD_ICE.F90
```diff
@@ -161,7 +161,7 @@ MODULE MOD_ICE
     !___________________________________________________________________________
     ! total number of ice tracers (default=3, 1=area, 2=mice, 3=msnow, (4=ice_temp)
 #if defined (__oifs) || defined (__ifsinterface)
-    integer                                     :: num_itracers=4
+    integer                                     :: num_itracers=6
 #else
 !    integer                                     :: num_itracers=3
     !------------------------------
```
ackerlar, line 167: you now always allocate 6 tracers regardless of whether icebergs are included or not. (For the ECHAM case, this is actually also true because __async_icebergs is always TRUE, but maybe we would like somehow account for this?)

FinnHeu: Yes, 6 tracers are always allocated when compiling with oIFS. But to me, this is the safest solution because:

- `use_icebergs` is a runtime setting, not a compile-time flag. The tracer array size is determined at compile time, so it cannot depend on `use_icebergs` without recompilation.

- We're now always allocating 2 extra tracer slots which are simply not used when the iceberg code is not executed, they remain empty. This will cause increased memory usage but nothing significant I guess.  

For now, "always allocate 6" to me is the simplest, safe fix that lets users toggle icebergs on/off via namelist without recompiling.

ackerlar (approved):

PR #937, opened 2026-06-17
Feat/move grounded icebergs

FinnHeu:
Allow grounded icebergs to slowly drift instead of forcing their velocity to 0. The velocity of grounded icebergs is scaled by a factor in the range [0, 1] depending on how deep the iceberg penetrates the seafloor, in accordance with Marsh et al. (https://doi.org/10.5194/gmd-8-1547-2015). 

```
fact=1.0-((draft_scale(ib)*abs(depth_ib)-Zdepth) / (draft_scale(ib)*abs(depth_ib)))
u_ib = u_ib*fact
v_ib = v_ib*fact
```

Icebergs which hit the seafloor are, in reality, not necessarily stuck there until the ice draft becomes shallow enough for the iceberg to float again but are e.g. lifted by tides or can "plough" through the sediments. Additionally, this prevents icebergs from getting stuck and accumulating in certain locations for a long time. 

The plot shows two iceberg trajectories of a permanently grounded and a floating iceberg seeded at the same location, same time. The grounded one is moving way more slowly but still changing its position.

download-5


ackerlar (approved):

FinnHeu:
download-1
Seems to work fine :)

FinnHeu:
I think the test fails due to me fixing a calving day bug (setting the default calving day from 1 to 0 in https://github.com/FESOM/fesom2/pull/937/commits/69741acee81edeb47da5f14be32e9eb2b3452dc4). Before, the icebergs calculations were skipped at the first day of a run. The ground truth has to be updated. @JanStreffing, would you - again - be so kind :)

ackerlar (approved):

JanStreffing (approved):

PR #940, opened 2026-06-23
Support Single Precision mode across the model

suvarchal:
Single precision capability, 
adds speed up about 1.4x  on core2 and ng5 and needs use of sss_anomaly =true to get drift free.

 
  Adds opt-in single-precision (-DUSE_SINGLE_PRECISION=ON) for standalone fesom.x. Default build stays double precision and unchanged. Mesh/restart/forcing files remain double on disk; WP arrays convert on I/O so existing features. CVMIX is in DP.  

## CORE2
Benchmark: core2 mesh, core2 forcing, 1 year, on levante 64-tasks-1-compute node, used KPP as CVMIX needs bit more work to be used in SP mode. Compiled with intel compiler. 
SP:    SYPD :                       22.0574 years/day
DP:   SYPD :                      13.3427 years/day
speedup : ~ 1.69x 
 
spatial plots are just last month (december)
a_ice_north_SP_DP
a_ice_south_SP_DP
a_ice_timeseries_SP_DP
sst_SP_DP
ssh_SP_DP
salt_timeseries_SP_DP
sst_timeseries_SP_DP
ssh_timeseries_SP_DP

## PI   January 1948, 1 month)
  Benchmark: bundled pi mesh (~3.1k nodes), CORE2 forcing, 2 976 steps (Δt = 900 s), lean monthly I/O (sst, sss, ssh, a_ice). Two builds: build-sp (SP) and build-dp (DP). Local ubuntu, GNU + OpenMPI.

  Global metric — SYPD (Simulated Years Per Day, from fesom.stats): higher = more simulated time per wall-clock day.

  SP vs DP: ~1.21× faster wall time at both rank counts (+17% at mpi2, +19% at mpi8). SYPD: +21% (mpi2) and +23% (mpi8).

  MPI scaling (same precision): mpi2→mpi8 ≈ 2.7× wall-time speedup and 2.7× SYPD (sub-linear vs 4× cores on this small mesh).

  Largest mpi2 component gains (SP/DP): oce_mix_pres ~1.6×, oce_ssh_solve ~1.55×, oce_timestep_ale ~1.27×. ice_timestep ~neutral (~1.01×).

  Field sanity (month 1, mpi2): SST max|Δ| ≈ 0.24 °C, SSS ≈ 0.05 psu, SSH ≈ 2.3 mm — SP and DP stay close.
mpi2_a_ice_north_sp_dp_diff
mpi2_a_ice_south_sp_dp_diff
mpi2_ssh_sp_dp_diff
mpi2_sst_sp_dp_diff


koldunovn:
We used this PR as the design reference for an FP32 working-precision mode in our C++/Kokkos port of FESOM2, and hit two failures in algorithms that the Fortran code shares. Both are latent in `support_sp`. We have not run your branch: the failures are measured in our port, and the Fortran-side reasoning is from reading the source at `636a60a3`, so please sanity-check it.

## 1. The forcing time axis cannot be FP32 (crash with 3-hourly forcing, silent with 6-hourly)

`flf%nc_time` is `real(WP)` (`gen_surface_forcing.F90:203`) and this PR moves its broadcast to `MPI_WP` (`:390`). After `:474`–`:504` it holds an **absolute Julian day**, ~2.44e6 for 1958.

float32 ulp at 2.44e6 is exactly **0.25 d = 6 h**:

```
ulp(float32) at JD 2436205 = 0.250000 d = 6.00 h
```

The representable time grid in an SP build is therefore 6-hourly, whatever the forcing dataset is.

**Crash path.** JRA55-do is 3-hourly. Adjacent record times collide after rounding, so `delta_t = nc_time(t_indx_p1) - nc_time(t_indx)` (`:784`) is 0 for many pairs, `coef_a = (data2-data1)/delta_t` (`:996`) is ±Inf, `coef_b` (`:997`) is ∓Inf, and `atmdata = rdate*coef_a + coef_b` (`:1022`) is Inf−Inf = NaN over the whole surface field. Neither `delta_t = 0` nor a non-finite `coef_a` is guarded; the `delta_t = 1.0_wp` assignments at `:828`/`:838` cover only the out-of-range-date branches.

Rounding 16 record times at JD 2436205 to float32:

| forcing | spacing | distinct times | zero-`delta_t` pairs (of 15) |
|---|---|---|---|
| CORE-II | 6 h | 16/16 | 0 |
| JRA55-do | 3 h | 9/16 | 7 |
| hourly | 1 h | 3/16 | 13 |

In our port every SP run died at ~9.4 model hours, all tracers NaN at once, because the global salt/heat balance correction turns one bad node into a whole-field NaN in a single step. The trigger is a fixed point in **model time**, not step count: dt=1800 s died at step 19, dt=900 s at step 37.

**Why the PR's validation could not see it.** The benchmark used CORE-II, whose 6-hourly spacing is exactly one float32 ulp at that magnitude and is grid-aligned, so `delta_t` survives as an exact 0.25. CORE-II therefore passes for a reason that does not carry over to any finer dataset.

**The silent half, which does affect CORE-II.** `rdate` is `real(WP)` too and is built as `julday(...) + (daynew-1) + timenew/86400` (`:644`–`:645`), so it is snapped to the same 6-hourly grid. In an SP build the model clock that the forcing sees advances in 6-hour jumps:

```
hour-of-day  ->  rdate(f32) time-of-day
 0,1,2,3     ->   0.00 h
 4,5,6,7,8   ->   6.00 h
 9..15       ->  12.00 h
16..20       ->  18.00 h
21,22,23     ->  24.00 h
```

Every `atmdata` sample therefore lands on a record node, and the linear interpolation in `data_timeinterp` degenerates to a staircase: nearest record, ties to the earlier one. For a field changing by Δ between records the mid-interval error is Δ/2, so a 1 K/record air-temperature ramp gives −0.5 K and a 5 K ramp gives −2.5 K. The 1-year CORE-II SP run in the PR body ran on piecewise-constant forcing, which bears on its SST/SSS deltas.

**Fix we used:** keep the whole time-affine chain in FP64 — `nc_time`, `rdate`, `delta_t`, `coef_a`, `coef_b`, and the `binarysearch` over `nc_time` — and narrow only the interpolated value to WP at the end. In Fortran that means declaring these `real(real64)` explicitly rather than letting them ride the default-real promotion, and reverting the `nc_time` broadcast at `:390` to `MPI_DOUBLE_PRECISION`. The cost is the time axis itself plus `coef_a`/`coef_b`, which are `i_totfl × (myDim_nod2D+eDim_nod2D)` (`:1186`), and the once-per-step surface loop at `:1022` then reads doubles instead of floats.

If that array cost matters, the offset form avoids it: store `data1` and the record time `t0`, and evaluate `atmdata = data1 + real(rdate - t0, WP) * coef_a`. Only the time variables need to be FP64 — the difference `rdate - t0` is bounded by one record interval — so both node-sized arrays stay WP. It also removes the ~1e7-magnitude `coef_b`, which is a cancellation hazard in its own right.

The general criterion is that an absolute-calendar quantity is FP32-safe only when the record spacing is ≥ ulp(magnitude) *and* grid-aligned, which is luck rather than design. Keeping time in double is the safer default.

## 2. `update_stiff_mat_ale` accumulates open-loop in WP — zstar/zlevel diverges after months

`init_stiff_mat_ale` is called once, from `oce_setup_step.F90:138`. `update_stiff_mat_ale` then runs every step (`oce_ale.F90:3648`, non-`linfs` ALE with the implicit SSH solver) and does

```fortran
SSH_stiff%values(npos) = SSH_stiff%values(npos) + fy*factor     ! oce_ale.F90:1769
```

The CSR `values` array is `real(kind=WP)` (`MOD_MESH.F90:14`) and is **never rebuilt**, so it accumulates over the entire run.

The two `fy` expressions differ only in what multiplies the geometric factor. In `init` (`:1535`) it is the column depth `(zbar_e_bot - zbar_e_srf)`, O(1e2–1e4 m); in `update` (`:1754`) it is `dhe`, the per-step SSH increment `sum(hbar - hbar_old)/3` (`:2087`), O(1e-4–1e-3 m). The per-step relative increment is therefore ~1e-7 … 1e-6, against float32 eps = 1.19e-7. A large share of increments land at or below one ulp of the value they are added to and are dropped, and the dropping is systematic per CSR row rather than random, because it depends on that row's magnitude. The diagonal is worse: `init` adds `areasvol/dt` there (`:1581`), making the value larger and the absorption more complete.

The invariant `sum(dhe)` = `hbar - hbar(0)` is what makes this open-loop scheme correct in FP64. Under FP32 absorption it stops holding row by row, the operator de-tunes, and eventually an eigenmode of the SSH system crosses |λ| = 1.

What we measured in our port, on the same scheme, with zstar on the CORE2 mesh at dt 1800 s:

- Two 63-year SP hindcast arms and two cold-start replays, on different backends and different rank counts, all died at the same model date: 1964-10-04/05, step 118503 ± 1. Deterministic and date-locked, which is what a slowly de-tuning operator crossing a threshold under a seasonally varying state looks like.
- Global max|eta| sat at 1.85–1.97 m for the whole run, then went exponential from ~1964-09-22, doubling roughly every 20 steps and reaching 48.3 m at step 118500. The mode was decoupled from the layer thicknesses.
- Death chain: runaway `d_eta` shear drives tracer advection, T reaches −7e7 while still finite, the EOS returns −inf in `pressure_bv`, and the PGF then puts a NaN into the SSH solver's first iteration. The first NaN is several stages downstream of the cause.
- The FP64 twin runs through the same period with no signature.

**Fix we used:** keep an FP64 shadow of `values`, seeded from the base matrix at init. Increments accumulate there, and the WP working copy the solver uses is refreshed from it at each update, so the result is rounded once per step instead of accumulating rounding over the whole run. The CG SpMV keeps float bandwidth. The cost is one `nnz`-sized FP64 array, ~7 MB on CORE2, plus a per-step copy, and only for non-`linfs` ALE. In the DP build the shadow is bit-identical to the current code.

After the fix the same replay completed all 120k steps with no NaN, and the culprit node's `d_eta` stayed within ±0.003 m throughout.

Rebuilding the matrix from scratch each step would also work and needs no extra storage, at the price of an edge loop per step.

## Scope

Bug 1 affects any SP run with forcing finer than 6-hourly, which includes JRA55-do. Bug 2 needs `which_ale` ≠ `linfs` and the implicit SSH solver (`use_ssh_se_subcycl = .false.`), plus a run long enough for the drift to bite; ours took just under seven model years. Neither is reachable by a 1-month or 1-year CORE-II test, which is why the validation in the PR body is green.

We can share the diagnostics if that helps. What found both was a per-phase non-finite scan, enabled by an environment variable, that aborts at the first phase where a non-finite value appears, plus a single-node end-of-step series for `eta`, `d_eta` and `ssh_rhs`. `feenableexcept` was no help, because by the time the NaN is visible it is being loaded rather than computed.

suvarchal:
Rebased onto current `main` and brought up to date with everything that has landed since.
**Now mergeable** (was CONFLICTING).

### What came in

`main` had moved 150 commits, including five pieces of this stack that are now upstream —
#984 (symmetric CG preconditioner), #991 (generic FCT min/max), #992 (`WP`/`WP_full` from
`iso_fortran_env`), #993 (`WP_full` global integrals), #994 (mixed-precision ssh CG), #996
(meandata flip-flag init). Those drop out of this PR's diff. #995 and #997 also merged into
`support_sp` in the meantime and are folded in.

### Conflict resolutions worth a look

| file | resolution |
|---|---|
| `oce_modules.F90` | kept main's `iso_fortran_env` + `WP_full` block and restored the `USE_SINGLE_PRECISION` switch on top of it — main hardcodes `WP=real64` |
| `oce_ale_mixing_kpp.F90` | took main. It turned the double-diffusion `alphaDT`/`betaDS` into vertical **differences**, in which a constant `S_ref_anomaly` cancels exactly — so main's form is already anomaly-correct and needs no offset. The genuine absolute-S use at the surface is untouched |
| `oce_ale_pressure_bv.F90` | took main's new neverworld2 cabbeling/thermobaricity/haline block, with `s` → `s_abs` in the two `(S-35)` terms, matching every other absolute-S reference in `density_linear` |
| `write_step_info.F90` | took main's new `ldiag_dMOC` density diagnostics, reduced with `MPI_WP` — `min_dens`/`max_dens`/`loc` are `real(kind=WP)` |
| `namelist.dyn`, `MOD_DYN.F90` | took #995's `precond_variant = -1` auto default, which subsumes the hardcoded value |
| `gen_modules_cvmix_kpp.F90` | main's change there was trailing whitespace only; kept the `WP ↔ cvmix_r8` shim |

### One class of bug the merge kept introducing

Three times now, an auto-merge has paired a `WP_full` buffer with `MPI_WP`, or a
`real(kind=WP)` buffer with an 8-byte MPI datatype. Both are invisible in DP, because
`WP_full == WP` there, and both corrupt memory in SP:

- `gen_support.F90` / `oce_mesh.F90` — fixed here, and separately proposed against `main` in
  #1026, where the same mismatch is currently live.
- `src/solver.F90` — #994 makes `sprod`/`s_old`/`s_aux` `WP_full` and reduces all four with
  `MPI_DOUBLE`; the merge resolved one site by conflict and silently kept `MPI_WP` at the
  other three. All four now match main exactly.

This was verified rather than assumed: reverting only the fix commit and rebuilding SP gives

```
WARNING: ssh CG breakdown at iter 1 : p.Ap=NaN is not positive
WARNING: ssh CG did not converge in 2000 iters; target rtol=NaN
```

because the corrupted global integrals feed `water_flux` through the freshwater and
virtual-salt balancing and drive `eta_n` out of bounds.

A named `MPI_WP_FULL` constant beside `MPI_WP` in `MOD_PARTIT.F90` would make this structural
instead of something every merge has to remember. There are further latent sites on `main`
(`gen_modules_cmor_diag.F90` has 9, plus `oce_ale_tracer.F90`, `io_restart.F90:614`,
`io_tracks.F90`, `icb_coupling.F90`, `toy_channel_neverworld2.F90`) — worth a separate issue.

### Validation

Local CTest suite (`-E '^(remote_|ecbundle_)'`), full config:

- **DP: 10/10**
- **SP: 10/10**

including `integration_pi_jra_mpi2` in **both** precisions.

config/namelist.dyn
```diff
@@ -73,11 +73,21 @@ soltol             = 1e-5        ! ssh CG relative tolerance (optional; default
                                  ! a ratio and cannot see the float32 residual floor, so
                                  ! the solver will report success it has not achieved
 maxiter            = 2000        ! ssh CG iteration cap (optional; default 2000 if omitted)
-                                 ! hitting the cap does not abort -- it is reported as
-                                 ! 'non-convergences' in the end-of-run solver summary
-precond_variant    = 1           ! ssh CG preconditioner: 0 = pre version 2.8.0 asymetric, 
-                                 !                        1 = symmetric 2D^-1 - D^-1*A*D^-1
-
+precond_variant    = -1          ! ssh CG preconditioner formula.
+                                 !  -1 = auto (default): 1 in a single-precision build,
```
JanStreffing: @patrickscholz and I already set the default to 1. I guess we want to keep that?

config/namelist.oce
```diff
@@ -19,6 +19,9 @@ C_d                = 0.0025  ! bottom drag coefficient (dimensionless, typical:
 A_ver              = 1.e-4   ! background vertical viscosity [m²/s]
 scale_area         = 5.8e9   ! reference element area for viscosity/diffusivity scaling [m²]
 
+! --- Salinity state representation ---
+use_salt_anomaly   = .false. ! store salinity as the anomaly S-35 (finer float32 spacing; helps single precision). .false. = absolute salinity, bit-identical to before
```
JanStreffing, line 23: Needs scientific signoff. I know we have the PSU vs absolute salinity discussion every once in a while.

src/cvmix_driver/gen_modules_cvmix_idemix.F90
```diff
@@ -369,46 +386,72 @@ subroutine calc_cvmix_idemix(partit, mesh)
...
                 ! FIXME: nils: better output IDEMIX Ri directly
-                alpha_c         = iwe_alpha_c(uln:nln+1, elem),       & ! out (for Ri IMIX)
-                ! only for Osborn shortcut 
+                alpha_c         = alpha_c_r8(uln:nln+1),             & ! out (for Ri IMIX)
+                ! only for Osborn shortcut
                 ! FIXME: nils: put this to cvmix_tke
-                KappaM_out      = iwe_Av(  uln:nln+1, elem),          & ! out
-                KappaH_out      = iwe_Kv(  uln:nln+1, elem),          & ! out
-                Nsqr            = bvfreq2( uln:nln+1),                & ! in
+                KappaM_out      = KappaM_r8(uln:nln+1),              & ! inout
```
JanStreffing, line 417: Changes look larger than between SP and DP. Intended?

src/cvmix_driver/gen_modules_cvmix_kpp.F90
```diff
@@ -620,29 +643,45 @@ subroutine calc_cvmix_kpp(ice, dynamics, tracers, partit, mesh)
...
+                        vE_r8(nun:nln)     = real(UVnode(2, nun:nln,  node), cvmix_r8)
+                        uS_r8(nun:nln)     = real(kpp_uS_t( nun:nln),        cvmix_r8) ! inout
+                        vS_r8(nun:nln)     = real(kpp_VS_t( nun:nln),        cvmix_r8) ! inout
+                        uSbar_r8(nun:nln)  = real(kpp_uS_m( nun:nln),        cvmix_r8) ! inout
+                        vSbar_r8(nun:nln)  = real(kpp_VS_m( nun:nln),        cvmix_r8) ! inout
+                        stokesXi_r8        = real(kpp_stokesXi_z(nz),        cvmix_r8) ! inout (out)
                         call cvmix_kpp_compute_StokesXi (&
-                            zbar_3d_n(nun:nln+1,node),   & ! full depth levels
-                            Z_3d_n(nun:nln,node),        & ! mid depth levels
+                            zbar_r8(nun:nln+1),          & ! full depth levels
```
JanStreffing, line 657: See above

src/io_meandata.F90
```diff
@@ -2178,7 +2189,22 @@ subroutine ini_mean_io(ice, dynamics, tracers, partit, mesh)
             write(*,*) '    Override via io_list entries in namelist.io'
         end if
     end if
-    
+
```
JanStreffing, line 2197: Does that make sense? 
Should we not truncate at what the model internals provide and write a warning of that?

src/oce_ale.F90
```diff
@@ -1705,6 +1714,31 @@ subroutine update_stiff_mat_ale(partit, mesh)
 #include "associate_mesh_ass.h"
 
     !___________________________________________________________________________
+#if defined(USE_SINGLE_PRECISION)
```
JanStreffing, line 1717: This reads like a workaround. Are we doing something funny a-prioi here?

src/oce_ale_tracer.F90
```diff
@@ -1878,8 +1883,26 @@ FUNCTION bc_surface(n, id, sval, nzmin, partit, mesh, sst, sss, aice)
     CASE (2)
         ! --> real_salt_flux(:): salt flux due to containment/releasing of salt
         !     by forming/melting of sea ice
+        ! use_salt_anomaly background-dilution term (S_ref=0 unless enabled, so
```
JanStreffing, line 1886: Reading this, I wonder once more, if this is not a bug in DP already?

JanStreffing (changes requested):
Fantastic work, @suvarchal!! Just a few questions and changes needed from my side.

We should not merge ~100mb of forcing files to main. And I think we should purge them from history. These should come via container.

I also dont see any changes to the coupling interface files. I wonder if we default keep DP there, or if this slipped through the cracks and we would get precision mismatches between what we send there now, and what the receivers expect.

Some more detail questions below:

build-dp.build.log
```diff
@@ -0,0 +1,806 @@
+[fesom]
```
JanStreffing: Should be cleaned up.

JanStreffing (changes requested):

config/namelist.dyn
suvarchal: yeah we get rid of option and just use symmetric preconditioner, it is almost a bug fix actually, i don't see a reason now to keep old one.

config/namelist.oce
suvarchal: can you explain what is  scientific signoff? , it is off by default. nothing changes in DP builds.

JanStreffing: Meaning that with all these conversions we don't mess anything up. I'm not an expert on ocean salinity though.

src/cvmix_driver/gen_modules_cvmix_idemix.F90
suvarchal: CVmix is always in DP, so if we are running in sp and use cvmix we need take care a bit about boundaries so many are intended but i also note that i see some dimensions may be changed. i will check if we have test coverage and see if everything is intact as before.

config/namelist.dyn
suvarchal: give a thumbs up @patrickscholz if you agree too.

suvarchal:
> Fantastic work, @suvarchal!! Just a few questions and changes needed from my side.
> 
> We should not merge ~100mb of forcing files to main. And I think we should purge them from history. These should come via container.
> 
> I also dont see any changes to the coupling interface files. I wonder if we default keep DP there, or if this slipped through the cracks and we would get precision mismatches between what we send there now, and what the receivers expect.
> 
> Some more detail questions below:

I thought about forcing and made subset really small 2 day set, i think it is important i thought as we extend sp further because JRA is forcing that we often use in production too and then the higher time freqency is also important check. we can reduce the forcing set to smaller after compression but need to check input reading routines once before that. 
the problem keeping it in container is it is a different workflow and very suitable for code development tests....

config/namelist.oce
suvarchal: that part is checked noting changes for dp. and values in output are in same units as before if we use this option or not.

src/io_meandata.F90
suvarchal: totally! this was to do, we give a warning and use what variable is represented in.

src/oce_ale_tracer.F90
suvarchal: right instinct! Storing `S = S' + S_ref` should be invisible in double precision: advection is linear, and by
continuity `adv(S_ref) = -S_ref*dh/dt` exactly cancels the `-S_ref*dh/dt` the anomaly transform
introduces, leaving the same equation in `S'`. That argument holds only if FESOM's free-surface
advection preserves a constant exactly — and it doesn't. With no correction, the DP
anomaly-vs-absolute salt discrepancy is ~1.3e-3, and it's localised exactly where the surface
freshwater flux enters, which is the fingerprint of a constancy error rather than roundoff.

this is a general model thing it only becomes visible when we use sss anomaly used. we can call it general model fix and be seperate but as sss anomaly comes in this PR it may sit here.

src/oce_ale.F90
suvarchal: this is the fix for a genuine bug, and it was independently reproduced by
the Kokkos port (see @koldunovn's comment) ; hit a deterministic date locked failure 1964-10-04. 

issue is that The problem: `update_stiff_mat_ale` runs every step and does
`SSH_stiff%values(npos) = SSH_stiff%values(npos) + fy*factor` on a CSR array that is never
rebuilt, so it accumulates over the whole run and after a point crashes. In future we should do this elegently by building matrix from scratch or periodically. the performance trade off we need to explore in another issue...

src/io_meandata.F90
suvarchal: aah i see we leave the option to the user because the output is not instantaneous it is accumulated over output period then this has an issue if we accumululate longer we may cross precision then the variables represent. I already hit this issue and that lead to constant bias. after averaging we can then truncate that is a possibility for future. I added a check checking all precisions and show a notice for now.

src/cvmix_driver/gen_modules_cvmix_idemix.F90
suvarchal: cv mix ctests are working in both sp and dp so interface is okay but that said all mixing schemes that are part of cvmix are not tested. only TKE is tested.

config/namelist.dyn
suvarchal: for now defaulting to 1.

JanStreffing (approved):

JanStreffing:
Thanks a lot @suvarchal :D

PR #945, opened 2026-06-26
Fix cvmix_KPP ice scaling: apply (1-Aice) to ustar and Stokes wind velocity

patrickscholz:
## Summary

**Affects only the cvmix_KPP mixing scheme** (`kpp_reduce_tauuice = .true.`). All other mixing schemes (PP, TKE, etc.) are unaffected.

When `kpp_reduce_tauuice = .true.`, the friction velocity `ustar` was scaled by `(1-Aice)^2` to suppress wind-driven boundary layer turbulence under sea ice, but the wind speed inputs to the Stokes drift / Langmuir turbulence parameterisation (`wind_norm`, `u_wind`, `v_wind`) were left at their full open-ocean values. This created two problems:

**1. EFactor blow-up under high ice concentration**
The Langmuir enhancement factor is `EFactor = sqrt(1 + c/LaSL^4)` where `LaSL = sqrt(ustar / uv_SLmean)` and `uv_SLmean ∝ 0.0162 * wind_norm`. As `a_ice → 1`, `ustar → 0` while `uv_SLmean` remained finite, driving `LaSL → 0` and `EFactor → ∞` — spurious infinite Langmuir turbulence enhancement directly under sea ice.

**2. Unphysical Stokes drift under ice**
The Stokes drift parameterisation uses a Pierson-Moskowitz fully-developed sea approximation, which assumes waves are locally generated by the local wind. Sea ice suppresses local wave growth, so applying the full open-ocean wind speed in ice-covered cells is physically inconsistent.

## Changes (`src/cvmix_driver/gen_modules_cvmix_kpp.F90`)

- **`ustar` scaling corrected**: `(1-Aice)^2` → `(1-Aice)`. Physical basis: effective wind speed at the ocean surface is reduced by `(1-Aice)` due to ice sheltering → `tau ∝ (1-Aice)^2` → `ustar = sqrt(tau/rho) ∝ (1-Aice)`. The original `^2` line is retained as a comment (it was ported from MPIOM without prior testing in FESOM).
- **New `u_wind_eff` / `v_wind_eff` variables**: scaled by `(1-Aice)` when `kpp_reduce_tauuice = .true.`, otherwise equal to `u_wind` / `v_wind`. Using the same factor as `ustar` preserves `LaSL = sqrt(ustar/us)` — both numerator and denominator scale identically, keeping EFactor finite.
- **All Stokes wind inputs updated**:
  - `wind_norm` (feeds `cvmix_kpp_ustokes_SL_model` and `compute_Efactor`, called twice)
  - Three calls to `compute_stokes_velocities_MOM6style` (two in the StokesMOST `do nz` loop, one after OBL depth is known)

## Notes

- No behavioural change when `kpp_reduce_tauuice = .false.` (default); `u_wind_eff = u_wind` exactly.
- The TKE Langmuir scheme (`tke_dolangmuir`) is unaffected — it derives its Stokes proxy from the wind stress, which is inherently reduced under ice.
- For reference: MOM6 handles this differently, suppressing the Langmuir enhancement multiplier (`lamult = 1`) above a 5% ice threshold at the coupler level rather than scaling wind inputs inside KPP.


JanStreffing:
Verified the bug and fix end-to-end. Should we have a test run of this?

patrickscholz:
Xuejing Chen found that i think she will make some test, i dont think anybody else except her is using this in moment. Also our default option for kpp_reduce_tauuice=.false. so far!

JanStreffing (approved):

PR #950, opened 2026-06-26
Fix mEVP/aEVP: zero stress tensor before thin-ice cycle to prevent stale stress application

patrickscholz:
## Summary

Affects **both mEVP and aEVP** rheology variants. When an element's ice mass falls below the skip threshold, the stress tensor update is skipped but the old stress values are left in memory. A mismatched skip condition in `stress2rhs_m` then applies those stale stresses to the momentum equation, spuriously accelerating nearly-vanished ice. The fix explicitly zeros the stress components before `cycle`-ing.

---

## The mismatch

The stress **update** and stress **application** use different skip conditions:

```fortran
! stress_tensor_m / stress_tensor_a — stress UPDATE (skips on low MASS)
msum = sum(m_ice(elnodes)) * val3
if (msum <= 0.01) cycle        ! sigma11/22/12 keep old values

! stress2rhs_m — stress APPLICATION (skips on low CONCENTRATION)
if (sum(a_ice(elnodes)) < 0.01) cycle
```

Ice **mass** and **concentration** do not go to zero simultaneously. During summer melt, an element can have:
- Very low mass (`msum <= 0.01 kg/m²`) — stress update is **skipped**
- Moderate concentration (`sum(a_ice) > 0.01`) — stress application is **not skipped**

The result: `stress2rhs_m` picks up `sigma11`, `sigma22`, `sigma12` from a previous timestep when the ice was thicker, and applies them to the momentum RHS. The nearly-vanished ice is driven by stresses that no longer reflect its physical state.

---

## Impact by EVP variant

**mEVP** (`stress_tensor_m`): `alpha_evp` is fixed, so stale stresses decay geometrically over sub-cycling iterations (`det1 * sigma` each step). The bug is present but self-damps over the EVP sub-cycle. Impact is localised and most visible during rapid summer decay.

**aEVP** (`stress_tensor_a`): `alpha` is recomputed each sub-cycle from the current `pressure`, which itself depends on ice conditions. As ice thins, `alpha` changes, making the stale stresses increasingly inconsistent with the current state. The bug is more persistent in aEVP because the relaxation rate is not fixed.

Both variants are most affected during **summer ice decay** — thin, patchy ice where the mass and concentration thresholds diverge most strongly.

---

## The fix

Zero `sigma11`, `sigma22`, `sigma12` explicitly before `cycle`-ing in both subroutines (`src/ice_maEVP.F90`):

```fortran
! Before (wrong — stale stresses survive the skip)
if (msum <= 0.01_WP) cycle

! After (correct — stress zeroed before skipping)
if (msum <= 0.01_WP) then
    sigma11(elem) = 0.0_WP
    sigma22(elem) = 0.0_WP
    sigma12(elem) = 0.0_WP
    cycle
end if
```

Applied identically to `stress_tensor_m` (mEVP, line 141) and `stress_tensor_a` (aEVP, line 1064). `stress2rhs_m` then contributes zero momentum tendency for those elements regardless of which concentration-based skip it uses — physically correct for nearly-vanished ice.

## Scope

- Affects all runs using mEVP or aEVP rheology (i.e. essentially all standard FESOM2 configurations)
- No namelist changes required
- Effect is most visible in summer Arctic/Antarctic marginal ice zones during decay season


patrickscholz:
i will make a test run to this !

patrickscholz:
Done for aEVP:

- Difference in Summer/Winter NH/SH sea ice concentration: 
a_ice_NH_nps_y2000-2019_mDJF
a_ice_NH_nps_y2000-2019_mJJA
a_ice_SH_sps_y2000-2019_mDJF
a_ice_SH_sps_y2000-2019_mJJA

- maximum NH/SH MLD2
MLD2_NH_nps_y2000-2019_mMar
MLD2_SH_sps_y2000-2019_mSep

- temperature/salinity biaS
salt_rob_y2000-2019_dep10 0m
temp_rob_y2000-2019_dep10 0m

Fazit: 
 this bux fix has very minor effect !!!

JanStreffing (approved):

PR #957, opened 2026-07-23
cpl(__oifs): coupled-slab ice surface temperature + sea-ice-thickness exchange

### LINKED ISSUE #877: ice_thermo_cpl check very cold ice_temp
https://github.com/FESOM/fesom2/blob/0a429531274dc94e4e5854497662034e3ac456a5/src/ice_thermo_oce.F90#L130-L131

> Why 173.15K @JanStreffing? You meant 273.15? It's still in main. @a270091 

 _Originally posted by @chrisdane in [5a19d91](https://github.com/FESOM/fesom2/commit/5a19d9132#r181199934)_

JanStreffing:
## What

Adds the FESOM side of **OIFS coupled-slab sea-ice** coupling. Under coupled-slab mode (`LNEMOLIMTEMP=.false.`, `LNEMOLIMTHK=.true.`) the atmosphere solves the ice-tile skin temperature itself, using slab conduction through the **coupled** ice/snow thickness rather than a fixed nominal slab. This wires FESOM to feed that scheme and to book the growth flux consistently.

## Changes (all `#ifdef __oifs` / `__ifsinterface`)

- **Send effective sea-ice thickness as a new coupled field `sit_feom`** (`cpl_send` slot 8, `nsend` 7→8 under `__oifs`; `gen_forcing_couple`). Sent grid-mean (per-ice × concentration), weighted convention like `ist`/`alb`, so the atmosphere's slab conduction uses the real ice thickness.
- **`ist_ref` anchor** — remember the ice surface temperature as *actually transmitted* at the OASIS send (the temperature the atmosphere evaluated its ice-tile flux at), captured only on real transmissions.
- **`ice_surftemp`: `dQ/dT`-linearized (implicit) skin solve** anchored at `ist_ref`, `zlam = 4·ε·σ·tref³ + turbulent sensitivity`. Unconditionally stable regardless of snow/ice insulation; leaves the fixed point unchanged where the surface tracks the coupling temperature. It sets the internal skin temperature used for **albedo / melt-pond state only**.
- **Growth budget:** the atmosphere owns the surface energy balance in coupled-slab mode, so `Qatmice = -a2ihf` (received atmosphere flux) directly.

## Compatibility

Inert for non-OIFS builds. For OIFS builds, `sit_feom` is defined via `oasis_def_var` as an extra send; OASIS leaves it **uncoupled** (a no-op `put`) wherever the namcouple does not reference it, so this does **not** require a coordinated namcouple change to remain runnable — only a namcouple that *adds* `sit_feom` needs a FESOM that provides it.

## Validation

Multi-year AWI-ESM3 (FESOM2-CORE3-cavity + OIFS-48r1) coupled spin-ups run stably with this + the OIFS receive-side; the implicit solve holds through the polar-night temperature range with no cold runaway.


patrickscholz (approved):

JanStreffing:
We still need the ice temp because we pick albedo from that.

PR #958, opened 2026-07-23
cavity: reset restart hnode to the mesh nominal in restart_thickness_ale (moving cavity)

JanStreffing:
## What

Fixes a persistent tracer-content drain when restarting a run whose ice-shelf **cavity geometry has changed between legs** — i.e. a **moving-cavity** coupled setup (FESOM ↔ an ice-sheet model such as PISM, where the ice-shelf draft is updated and the mesh regenerated/remapped each coupling leg).

## Why it only bites a moving cavity

In `restart_thickness_ale`, cavity columns (and their open-ocean edge ring — any node touching a cavity element, `ulevels_nod2D_max > 1`) keep their layer thicknesses fixed at the mesh nominal (`hnode_new` from `init_thickness_ale`, the same exclusion class as the per-step `hnode`↔`hnode_new` sync loop). The previous code `cycle`d these columns on the restart read.

With a **static** mesh that is harmless: the restart `hnode` equals the nominal, so nothing drifts. But when the cavity **moves**, a column that was open-ocean zstar in the leg that *wrote* the restart can be cavity (or edge-ring) in the leg that *reads* it, so its remapped restart `hnode` no longer matches the new mesh's nominal. Because the column is `cycle`d, that mismatch **persists forever**, and the tracer `T*` update then applies it as a multiplicative content drain every step.

## Fix

For cavity/edge-ring columns, force restart `hnode` to the nominal (`hnode_new`) and rebuild the depth arrays (`zbar_3d_n`, `Z_3d_n`) before `cycle`. Open-ocean columns are unchanged, so static-mesh and non-cavity runs are entirely unaffected.

### DIFF
--- a/src/oce_ale.F90
+++ b/src/oce_ale.F90
@@ -1293,15 +1293,30 @@ subroutine restart_thickness_ale(partit, mesh)
         do n=1, myDim_nod2D+eDim_nod2D
             nzmin = ulevels_nod2D(n)
             nzmax = nlevels_nod2D(n)-1
-            
+
             !___________________________________________________________________
-            ! if there is a cavity layer thickness is not updated, its 
-            ! kept fixed 
-            if (nzmin > 1) cycle
-            
+            ! Cavity columns AND their open-ocean edge ring (any node touching
+            ! a cavity element, ulevels_nod2D_max>1 -- the same exclusion class
+            ! as the per-step hnode<->hnode_new sync loop) keep layer
+            ! thicknesses fixed at the mesh nominal (hnode_new from
+            ! init_thickness_ale, untouched by the restart read). A restart
+            ! hnode deviating from that nominal (e.g. remapped from a formerly
+            ! ice-free zstar column) would persist forever, and the tracer T*
+            ! update applies the mismatch as a multiplicative content drain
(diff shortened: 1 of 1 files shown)


JanStreffing:
Marking as **draft** — not merge-ready yet.

In practice this `hnode`-to-nominal reset only *partly* addresses the moving-cavity problem under **zstar**. It removes the multiplicative tracer-content drain from a remapped restart, but it does **not** recover cavity velocities to the quality we get under **linfs**: zstar moving-cavity columns still show elevated, persistent barotropic currents relative to a linfs run on the same mesh (the excess appears injected across the mesh increment and is cavity-wide and persistent). So the fix is necessary but not sufficient.

Reaching linfs-like velocities under zstar needs further work on the momentum/thickness handling **across a mesh increment**, not just the restart `hnode` seeding. Keeping this open as a partial correctness fix; it should stay in draft until that gap is closed.

(Note: it is inert for linfs — the whole block is guarded by `which_ale=='zstar' .or. 'zlevel'`, and the linfs restart path rebuilds `hnode` from the reference depth levels regardless.)

cavity_velocity_zstar34_vs_linfs37

PR #960, opened 2026-07-31
Fix uninitialised GM streamfunction at degenerate cavity level ranges, and two out-of-bounds reads

### LINKED ISSUE #1023: fesom_meshpart can produce partitions that blow FESOM up at mstep=1, and the METIS seed is compile-time only
On the CORE3 mesh (220509 nodes, 426403 elements, 48 levels, cavities on), partitions generated by `fesom_meshpart` sometimes make FESOM go non-finite on the first timestep, while the `dist_*` shipped with the mesh run cleanly. Whether a given partition is usable depends on the METIS seed, and the seed can only be changed by rebuilding the partitioner.

Same mesh, same `fesom.x` built from `main` at `f4150b1d`, same namelists. Standalone, cold start from JRA55 1958-01-01, `use_cavity=.true.`, 300 s, one day:

| npes | partition | METIS seed | result |
|---|---|---|---|
| 512 | shipped with mesh | | clean day |
| 1024 | shipped with mesh | | clean day |
| 128 | generated, `n_levels=1` | 35243 | clean day |
| 1024 | generated, `n_part = 8, 128` | 35243 | blow-up at `mstep=1` |
| 1024 | generated, `n_part = 8, 128` | 17 | clean day |
| 1792 | generated, `n_levels=1` | 35243 | blow-up at `mstep=1` |
| 1792 | generated, `n_part = 14, 128` | 35243 | blow-up at `mstep=1` |
| 1792 | generated, `n_part = 14, 128` | 17 | blow-up at `mstep=1` |
| 1792 | generated, `METIS_PartGraphKway` with `CONTIG=1` | 35243 | blow-up at `mstep=1` |

The two 1024 rows differ only in the seed. Every failure is temperature going non-finite at `mstep=1` at Antarctic cavity nodes, on the ranks holding those nodes.

The partitions are not obviously malformed. Owned nodes per rank are 116 to 128 with no empty ranks, and `check_partitioning`'s isolated-node repair runs and reports moves. The 1-ring halo in `my_list` is exact against the mesh connectivity for both generated and shipped partitions. Disconnected subdomains are not the discriminator either: `METIS_PartGraphKway` with `CONTIG=1` gave 0 of 1792 ranks with multi-component owned sets against 1458 for the default recursive bisection, and blew up anyway, while the shipped `dist_512` that runs cleanly has multi-component owned sets on several hundred ranks.

The behaviour is anticipated in the code. [fort_part.c:126-130](https://github.com/FESOM/fesom2/blob/f4150b1dbe55aa20a9d8d87f6ad361d1bc1339c2/src/fort_part.c#L126-L130) sets the seed and says to try different seeds until the partition is ok, and [fort_part.c:103-105](https://github.com/FESOM/fesom2/blob/f4150b1dbe55aa20a9d8d87f6ad361d1bc1339c2/src/fort_part.c#L103-L105) notes that `METIS_OPTION_CONTIG` is ignored by the recursive bisection actually called at line 233. But [mesh_part/CMakeLists.txt:82](https://github.com/FESOM/fesom2/blob/f4150b1dbe55aa20a9d8d87f6ad361d1bc1339c2/mesh_part/CMakeLists.txt#L82) hardcodes `METISRANDOMSEED=35243`, so following that advice means editing a build file and rebuilding for each attempt.

Two things would help. Exposing the seed as a cache variable or namelist entry would make the documented workaround usable without a rebuild. More importantly, seed roulette does not terminate here: at 1792 every partition we produced failed, across two seeds, flat and hierarchical decomposition, and the contiguous Kway path, so a mesh can be left with no usable partition at a task count it needs.

This reports a reproducible symptom rather than a diagnosis. We did not establish what makes a partition invalid.

JanStreffing:
Three independent fixes found while tracking down a reproducible leg-start blow-up in a coupled AWI-ESM3 run with a moving (PISM-driven) ice-shelf cavity. The first is the crash; the other two are latent out-of-bounds reads found with a bounds-checked build along the way.

### The bug

`fer_solve_Gamma` (`src/oce_fer_gm.F90`) solves the GM streamfunction over

```fortran
nzmax = nlevels_nod2D_min(n)   ! min over the node's surrounding elements
nzmin = ulevels_nod2D_max(n)   ! max over the node's surrounding elements
```

A node that touches **both a deep cavity element and a shallow one** can end up with `nzmin >= nzmax`, i.e. no interior range to solve on. Both sweeps then execute zero times:

```fortran
DO nz = nzmin+1, nzmax          ! e.g. 8..5 -> zero iterations
   ...
END DO
tr(:,nzmax) = tp(:,nzmax)       ! <-- still runs; tp(:,nzmax) never assigned
do nz = nzmax-1, nzmin, -1      ! e.g. 4..7 step -1 -> zero iterations
```

but `tr(:,nzmax) = tp(:,nzmax)` is unconditional, so an **unassigned element of the local automatic array `tp`** is written straight into `fer_gamma`.

### Why it is destructive

`fer_gamma2vel` turns that into a bolus velocity,

```fortran
fer_uv(1,nz,el) = sum(fer_gamma(1,nz,elnod) - fer_gamma(1,nz+1,elnod))/(3*helem(nz,el))
```

and `solve_tracers_ale` adds `fer_uv` to `UV`, **advects the tracers with it**, then subtracts it back.

Measured in the failing run (one bad gamma at the level-5 interface, node shared by three elements):

```
fer_uv(1, nz=4) = -1.63611672e+251     fer_uv(2, nz=4) = +1.8e-04    helem = 10.0
fer_uv(1, nz=5) = +1.63611672e+251     fer_uv(2, nz=5) = +3.6e-06    helem = 10.0
```

The antisymmetric pair is the signature of a single corrupted `fer_gamma` at that interface; `helem` and the meridional component are healthy. Consequences in one timestep:

* tracers advected at ~1e251 m/s: `temp` reaches 1e250, `salt` pins at its 3/45 bounds;
* the subtract-back leaves **exactly** `0.0` in `u` wherever `|fer_uv| >> |UV|`, because `(x + y) - y` rounds `x` away entirely;
* `Wvel`/`Wvel_e` get the same treatment via `fer_Wvel`, so the vertical velocity acquires exact-zero holes too.

### Why it is intermittent

`tp(:,nzmax)` is whatever the stack holds, so the outcome depends on memory layout. In our case the same source crashed at `mstep 1` under `-O3` and ran fine under `-check bounds -check pointers -check uninit -g`. Degenerate nodes were present in most of our meshes (0–2 per mesh) without being fatal, which is why this can sit latent for a long time.

`-check bounds` does **not** catch it: `fer_gamma` is written in bounds; the defect is a read of an unassigned value.

### Fix

A node with no interior range carries no GM transport, so zero its column and skip:

```fortran
if (nzmin >= nzmax) then
    tr(:,:) = 0.0_WP
    cycle
end if
```

Requires `Fer_GM = .true.`; the degenerate range requires `use_cavity = .true.`.

## 2 & 3. Two non-short-circuit `.and.` expressions reading index 0

Fortran does not guarantee short-circuit evaluation of `.and.`, so the compiler may evaluate the right operand even when the left is false.

**`src/ice_maEVP.F90`** (both `EVPdynamics_m` and `EVPdynamics_a`), applying the sea-ice velocity boundary condition at the cavity–ocean edge:

```fortran
if ( (ulevels(edge_tri(1,ed))>1) .or. &
    ( edge_tri(2,ed)>0 .and. ulevels(edge_tri(2,ed))>1) ) then
```

At a boundary edge `edge_tri(2,ed) == 0`, so `ulevels(0)` is read. The value read then decides whether the ice velocity at that edge is zeroed, i.e. a physics branch taken on out-of-bounds memory. Only reachable with `use_cavity = .true.`.

**`src/io_xios.F90`**, insertion sort of cell-corner angles:

```fortran
do while (k >= 1 .and. angles(k) > tmp_ang)
```

reads `angles(0)` at the end of every sweep. Harmless in practice, but it aborts any bounds-checked build during initialisation.

Both are rewritten so the guard is evaluated separately. No behavioural change other than removing the out-of-bounds access.

## Testing

* The failing leg previously blew up at `mstep 1` on every attempt, bit-identically; with fix 1 it ran past the failure point (day 27, step 1920) with no blow-up.
* Fixes 2 and 3 were verified against a bounds-checked build, which traps on them without the change and passes with it.
* Compile-checked against `main` (uncoupled Release build, Intel).


JanStreffing:
Reading back through. Not 100% clear yet why this bites with the moving cavity, and wether we should be able to avoid configurations where A node that touches both a deep cavity element and a shallow one.

JanStreffing:
> Reading back through. Not 100% clear yet why this bites with the moving cavity, and wether we should be able to avoid configurations where A node that touches both a deep cavity element and a shallow one.

Digging a bit more. The cavity is fixed inside a run, since `ulevels` is only ever assigned in `fvom_init.F90`. It changes between cycles when the submesh is regenerated, which is why this fails at leg start on `mstep` 1.

The geometry itself is not new. `max(ulevels) > min(nlevels)` over a node's elements finds 2 such nodes in the base mesh, and 1/2/1/2/0/2 across submeshes 1901 to 1906. So the question is why a decade of static cavities never tripped it.

The answer is that [`tp`](https://github.com/FESOM/fesom2/blob/e3c3d9d54ba916fa0cc03640a0f62d5c354994f9/src/oce_fer_gm.F90#L64) is thread-private and uninitialised at region entry, but it is reused across every node in a thread's chunk. A healthy node writes it over `[nzmin, nzmax]` at [:140](https://github.com/FESOM/fesom2/blob/e3c3d9d54ba916fa0cc03640a0f62d5c354994f9/src/oce_fer_gm.F90#L140) and [:147](https://github.com/FESOM/fesom2/blob/e3c3d9d54ba916fa0cc03640a0f62d5c354994f9/src/oce_fer_gm.F90#L147), and open-ocean nodes span the shallow levels, so by the time a thread reaches the degenerate node, `tp(:,nzmax)` normally holds the previous node's gamma at that level. That is an ocean-sized number, so the node gets a wrong GM transport rather than a blow-up, and nobody notices. Raw stack is only read when the degenerate node comes before any node in that chunk has written that level, which means at or near a chunk start.

With static scheduling that placement is deterministic, so a fixed mesh makes the draw once and keeps it for the whole run. Regenerating the submesh every cycle changes node numbering and partitioning, so it re-rolls where the degenerate node sits relative to a chunk boundary. Two nodes across eighteen-plus cycles is enough draws to land on one. The leg dying bit-identically on every retry fits the same picture, since the submesh, partition and schedule were all unchanged.

On whether we should forbid such nodes, I do not think so. [`oce_dyn.F90:973-976`](https://github.com/FESOM/fesom2/blob/e3c3d9d54ba916fa0cc03640a0f62d5c354994f9/src/oce_dyn.F90#L973-L976) computes the same range and survives it, because `c1=0.0_WP` is set before the loop so an empty range falls through to the `c_min` floor. [`oce_muscl_adv.F90:383-384`](https://github.com/FESOM/fesom2/blob/e3c3d9d54ba916fa0cc03640a0f62d5c354994f9/src/oce_muscl_adv.F90#L383-L384) computes it across both edge nodes and only ever writes inside its loops. `fer_solve_Gamma` is the only place that publishes a value derived from the empty range, so the defect is in the routine rather than the mesh.

JanStreffing:
Visualization of the area that ends up crashing:
peninsula_26533
ross_2244
range_bars_peninsula_ross


Thinking about whether the zeroing is ok for nodes without overlapping element depths:

`fer_gamma2vel` loops each element over its own levels `ulevels(el) .. nlevels(el)-1` and builds the bolus velocity in layer `nz` from `Γ(nz) − Γ(nz+1)` at the element's three nodes. Summing that over the element's column, everything telescopes: the depth-integrated bolus transport of element `el` is `⅓ Σ_nodes [Γ(ulevels(el)) − Γ(nlevels(el))]`. A GM streamfunction is only a streamfunction if that vanishes, otherwise the bolus field carries net horizontal mass and the continuity-derived `fer_Wvel` pushes flow through the ice base or the bed. So Γ must be zero at every node at the draft level and the bottom level of every element that node belongs to. A node whose elements have drafts at levels 2, 5, 14 and 15 therefore needs Γ = 0 from the top down to level 15, and with bottoms at 17 and 18 it needs Γ = 0 from level 17 down. The solve range `[max elem draft, min elem bottom]` is exactly that constraint, not a conservative shortcut. The same logic already governed the bottom in the old code (`minval(nlevels)` in the commented line 70); the cavity work extended it to the top.

So when the range is empty, Γ ≡ 0 is the only answer consistent with no normal flow. Zeroing is correct. The physical cost is that GM transport is switched off at nodes where the ice front steps by more than the local water depth, and tapered to a few levels nearby. That is the same suppression FESOM already applies at steep bathymetry, and GM is meant to vanish at boundaries anyway. 796 nodes out of 220509, all on the cavity front, is not a transport you would miss.

patrickscholz (approved):

PR #964, opened 2026-08-06
Fix/ib heatcap

ackerlar:
- capping iceberg related heat fluxes (can be turned on/off via namelist entry l_cap_ibhf_n)
- "call trajectory" has been omitted entirely for "l_allowgrounding == 2" - fixed this now only for grounded icebergs
- avoid by zero division in subroutine trajectory for high northern latitude icebergs


JanStreffing:
Do you have some statistics on how often this fires, and what percentage of ICB heat flux are lost on one of your runs this way?

ackerlar:
discarded_hf

This is from a 60 year test run. The discarded heat is written out in the FESOM log as discarded heat per step. Accumulated over one year it is less than 1% of the total iceberg related heat flux. 

EDIT: However, this is the global integral. It might be larger locally.

src/gen_modules_config.F90
```diff
@@ -132,6 +132,7 @@ module g_config
   logical                       :: lmin_latent_hf=.true.
   logical                       :: lverbose_icb=.false.  
   integer                       :: l_allowgrounding=1    ! 0=free drift, 1=slow drift, 2=stationary
+  logical                       :: l_cap_ibhf_n=.true.   ! cap iceberg-driven interior cooling (ibhf_n) at a safe temperature floor per cell/step
```
JanStreffing: Should the default be true, or perhaps this can be a namelist change on automatic retry with esm_tools:
https://github.com/esm-tools/esm_tools/pull/1460

JanStreffing (changes requested):

src/gen_modules_config.F90
JanStreffing: The fact that this seems to trigger consistently is a sign that every run of this simulation would crash otherwise?

ackerlar: > The fact that this seems to trigger consistently is a sign that every run of this simulation would crash otherwise?

No, not necessarily. It just checks whether iceberg melting would lead to unphysical low temperatures, currently set to `ibhf_n_t_threshold = -1.85_WP`. The FESOM internal check is at `-5`. So the model would still continue but might run into a blowup over time.

But I will set the default to `.false.` and then trigger it via `esm_tools`

JanStreffing (approved):

PR #965, opened 2026-08-07
restart: make a restarted run reproduce a continuous one

### LINKED ISSUE #350: Using restarts instead of continuous run leads to differences

koldunovn:
Closes #350. Based on #919 — merge that first.

A CORE2 / zstar / cvmix_TKE / mEVP / Fer_GM+Redi run stopped and restarted after one day did not reproduce the same run done in one go. Two groups of causes.

**Prognostic state missing from the netCDF restart.** All four are already written by the binary (derived-type) restart path.

| field | why it is prognostic |
|---|---|
| `sigma11/12/22` | mEVP carries the internal stress across steps (`ice_maEVP.F90`, `sigma12 = det1*sigma12 + det2*r3`); a restart re-spun it from zero |
| `t_skin` | seeds the Newton solve for the ice surface temperature (`ice_thermo_oce.F90:280`), fixed `imax=5`, no convergence test (`:713`, `:741`) |
| `d_eta` | passed to `ssh_solve_cg` as `X` — the CG initial guess as well as the solution; the CG stops on a relative residual, not at machine precision |
| `hnode_new` | `fer_solve_Gamma` and `init_Redi_GM` run before `vert_vel_ale` and read it (`oce_fer_gm.F90:82`, `:266`), so with `Fer_GM` they consume the previous step's value |

**SSH preconditioner built from a moving matrix.** It was built under `logical, save :: lfirst` on the first solve of the run. `SSH_stiff%values` is cumulative — `update_stiff_mat_ale` adds the elevation increment `dhe` once per step and it is never rebuilt — so a cold start built the preconditioner from the unperturbed matrix (`dhe==0` because `hbar==0`) while a restart built it after `restart_thickness_ale` had fed the absolute sea surface height in. Now built in `ocean_setup` after `init_stiff_mat_ale`. The lazy call is kept as a fallback but keyed on `allocated(pr_values)`, which also fixes the binary restart path aborting with `forrtl: severe (151): allocatable array is already allocated` in `ssh_solve_precond`.

### Result — first step after the restart, against a continuous run

| field | before | after |
|---|---|---|
| `sst` | 2.681e-04 K | 4.97e-14 K |
| `ssh` | 1.07e-05 rel | 8.86e-14 rel |
| `d_eta` | 4.24e-06 rel | 6.42e-15 rel |
| `fer_C` / `bolus_w` | 1.2e-03 / 1.5e-03 rel | bit-identical |
| `a_ice`, `uice`, `fh`, `fw`, `unod`, `Kv` | round-off | round-off |

Cold start unchanged: `max|diff| = 0` over a whole run.

### Notes

- Backward compatible: `def_*_var_optional`, so restarts written before this change still load and reproduce the old behaviour.
- Numbers measured on 4ccdd35a + #919, CORE2 mesh, 128 ranks. This branch compiles on current main; the runs were not repeated there.
- Round-off still amplifies afterwards: one day after the restart the SST difference is ~1e-4 K area-weighted RMS (max 7e-2 K at a handful of sea-ice nodes). That is freeze/melt branch flips, not remaining state loss — tightening `soltol` from 1e-5 to 1e-8 leaves the divergence curves unchanged at 3x the runtime.


sebastianbeyer:
Nice that it's backwards compatible, but should we emit some kind of WARNING when the variables are not present? But maybe that's more of an edge-case and not too interesting for most people.

koldunovn:
It already does — the read path prints one line per skipped variable (`io_restart.F90:874`):

```fortran
write(*,*) 'skipping reading restart for ', filegroup%files(i)%varname, ' at ', filegroup%files(i)%path
```

The netCDF restart is one file per variable, so a missing optional variable is a missing file and that branch fires with the name. Restarting from a pre-fix restart gives `skipping reading restart for t_skin`, and the same for `d_eta`, `hnode_new`, `sigma11/12/22`.

I didn't add anything on top of that, for two reasons. The other optional variables (`tke`, `iwe`, `uke`, `ice_albedo`, `ice_temp`, tracer `_M1`/`_M2`) go through the identical path, so special-casing these six would be inconsistent. And it only fires on the first restart after upgrading — every restart written afterwards contains all six.

If the wording should be stronger than "skipping" (it reads as informational, not "this restart is no longer reproducible"), that's a one-line change at that spot and it would cover all optional variables at once. Happy to do it as a separate PR if you think it's worth it.

sebastianbeyer:
Ah, then that's fine :)

koldunovn:
OK, all green - I've updated the numbers in tests. Better to merge #965 first probably (which ia also green now).

sebastianbeyer (approved):

PR #966, opened 2026-08-07
awiesm3: coupled-slab ice surface temperature, cavity hnode reset, iceberg bounds-check option

JanStreffing:
The branch AWI-ESM3 `develop-is` is built from. Opening it for visibility and review, **not** to be merged yet: see the ordering note at the bottom.

## What is in it that is not already on main

- **`cpl(__oifs)` coupled-slab ice surface temperature.** The implicit, `dQ/dT`-linearised skin solve, `sit_feom` sent as effective ice thickness, and the atmosphere owning the growth flux in coupled-slab mode (`Qatmice = -a2ihf`). This is the same work as #957, carried here because the AWI-ESM3 build needs it: a binary without `sit_feom` aborts every component at startup with `(oasis_coupler_setup) ERROR: namcouple variable not used: sit_feom`.
- **`cavity`: reset restart `hnode` to the fixed nominal in `restart_thickness_ale`**, which is #958.
- **Two cherry-picks that are still open as their own PRs**, #960 (uninitialised GM streamfunction at degenerate cavity level ranges, plus two out-of-bounds reads) and #961 (mask ice albedo on ice-free cells in the XIOS sender).
- **`FESOM_ICB_BOUNDS_CHECK`**, an opt-in CMake option that builds only the iceberg sources with `-check bounds -check pointers -traceback -g -O0`, leaving the rest of the model optimised. Off by default. It is here because it earned its keep: it caught the `ulevels_nod2d(0)` read of #963 within 100 timesteps of a coupled run, on code that had been reading out of bounds silently.

## Already on main, listed so reviewers are not surprised to see the commits

#962, #963 and #964 all merged while this branch was in use, so the branch's own copies of them are redundant. The merge commit for #964 is still in the history here.

## Do not merge this before the others

This must go after #960 and #961, which are the same changes as standalone PRs and should be reviewed on their own terms rather than swept in as part of a large branch. If #957 is the preferred home for the coupled-slab work, that should merge first too and this branch should be rebased onto it, leaving only the cavity `hnode` reset and the CMake option here.

Merging this first would land four separable changes in one commit and make the individual PRs harder to review or revert.

## Validation

Runs coupled in AWI-ESM3 (FESOM2 CORE3 cavity + OIFS 48r1 + PISM), including across a mesh-change leg with icebergs: a full year with `ib_num = 1883`, 948 carried and 935 calved, completing normally.


JanStreffing:
This leftover is not needed. The moving cavity uses linfs for now. zstar is TODO for later. This "fix" was not it.

Issue #279, opened 2022-01-25
Only can restart from the last record in restart file


koldunovn:
I am on `refactoring` branch.

Running simulation with monthly restarts, that, say, last for 3 months with monthly restarts. I have 3 restart records with time stamps:
`time = 2675700, 5094900, 7773300`

I would like to restart from the month 2, and chance the `fesom.clock` to:

```
   83700.0000000000               59        1958
   86400.0000000000               59        1958
```

I get the error:

```
raw restart time    7773300.00000000      does not match current clock time   5094900.00000000
```

So there is no way to start from anything but last record in the restart? If so, it's not a good behaviour, since in high resolution setups we restarts from couple of steps before the last one quite often.

@hegish @dsidoren

hegish:
Looks like there is a fesom_raw_restart directory with an*.info file available (i.e. memory dump restart). If so, the memory dump is currently always preferred over the regular netcdf restart. You can remove the fesom_raw_restart directory and it should work as always.

helgegoessling:
Speaking of FESOM restart files: It has always been a stumbling block that FESOM restart files contain multiple times rather than on file (or set of files) per restart date, where it would be straight forward to also name the exact restart date in the file name. Would also make the life for ESM tools easier; Miguel is working on an issues related to exactly this right now, if I understand correctly. Can we maybe consider changing this sometime?

koldunovn:
@hegish I tried to remove the binary restarts, result is the same. Does this code imply we only look for the last record? https://github.com/FESOM/fesom2/blob/092a79439387fcbd645087388f2d6e0e649a6ac7/src/io_restart.F90#L479

koldunovn:
> Speaking of FESOM restart files: It has always been a stumbling block that FESOM restart files contain multiple times rather than on file (or set of files) per restart date, where it would be straight forward to also name the exact restart date in the file name. Would also make the life for ESM tools easier; Miguel is working on an issues related to exactly this right now, if I understand correctly. Can we maybe consider changing this sometime?

Moving around individual files would be easier of course. @hegish implemented the new scheme, where each year get it's folder and then each variable. 

Before, when the grass was greener and the files much smaller having everything in one file was reasonable, now, when I have restarts of several TB per year, having them splitted to individual time steps make sense. @dsidoren ?

koldunovn:
[...]

Issue #617, opened 2024-08-15
output files should be separate from input files


sebastianbeyer:
Hey,
as we started discussing during the retreat, I would like if the output files of FESOM were different from the inputs. 
This would make handling the runs easier in many cases, e.g. we would not need to copy the initial/restart files for every new run that we do. Also for runs that failed at some point we could simply delete the outputs and start from the same inputs again, so the model is always behaving the same for the same way of configuration/namelist. 
We would need to set the input and output folders via namelist parameters then, I guess.

JanStreffing:
We already had a more extensive discussion on this here: https://github.com/FESOM/fesom2/issues/279. IMO we should continue there.

Issue #485, opened 2023-08-01
Do not add timesteps to existing restart files


JanStreffing:
**The issue:**
Currently when FESOM2 finds preexisting restart files in a restart folder for the given year, e.g.
`fesom.1849.oce.restart/ssh.nc`
and we ask it to write monthly restarts during year 1849 (e.g. during first year of spinup from PHC3), it will keep adding restart timesteps to the netcdf files therein:
```
cdo sinfo temp.nc_18490501-18490531
   File format : NetCDF4 classic
    -1 : Institut Source   T Steptype Levels Num    Points Num Dtype : Parameter ID
     1 : unknown  unknown  v instant       1   1         1   1  I32  : -1            
     2 : unknown  unknown  v instant       1   1 249666860   2  F64  : -2            
   Grid coordinates :
     1 : generic                  : points=1
     2 : generic                  : points=249666860 (3160340x79)
   Vertical coordinates :
     1 : surface                  : levels=1
   Time coordinate :
                             time : 5 steps
  YYYY-MM-DD hh:mm:ss  YYYY-MM-DD hh:mm:ss  YYYY-MM-DD hh:mm:ss  YYYY-MM-DD hh:mm:ss
  0000-00-00 00:00:00  0000-00-00 00:00:00  0000-00-00 00:00:00  0000-00-00 00:00:00
  0000-00-00 00:00:00
cdo    sinfo: Processed 2 variables over 5 timesteps [0.16s 33MB]
```

Upon restarting from such a folder/files FESOM2 will use the latest step found and double check that it matches with the fesom.clock. If the timesteps don't match, the model exits.

I find this behavior unsafe. Especially when one accidentally links instead of copying a restart from a pool_dir to a work folder, fesom will try to modify to original restart files in the pool_dir and start adding timesteps there. In the worst case scenario, the user will have write permissions on the pool dir, and the restart file will actually be modified. This recently happened to @mzapponi.

**Proposed solution:**
I think a better solution, would be to have a more detailed timestamp on the restart folder name. e.g. `YYYY-MM-DD-HH-MM-SS`, or at least  `YYYY-MM-DD`. Instead of checking if the folder/file exists, and if it does adding a timestep, we can check if it exists, and if so, exit the model. This way we never accidentally modify an existing restart file.

Unless I hear a strong no, to the suggestion, I would create a draft for such a change soon.

@patrickscholz @hegish @dsidoren

trackow:
[...]

PR #969, opened 2026-08-09
io: collective parallel writer for output and restarts

### LINKED ISSUE #967: compression_level > 0 roughly doubles wall-clock cost at high resolution
NG5 (7.4M surface nodes), 64 nodes / 8192 tasks on Levante, 31 output streams. Same run,
1958, only `compression_level` differs:

| | months 01–02 | months 03–12 |
|---|---|---|
| `compression_level` | 1 | 0 |
| `runtime output` (max over ranks), per model month | 3610 s | 577 s |
| written per model month | 76 GB | 134 GB |

Compression at level 1 costs **~50 min per model month**, against ~54 min/month for the
whole uncompressed run. It nearly doubles the run to save 43% of disk. At 64 nodes:
**660 node-hours per simulated year, ~6600 node-hours for a 10-year hindcast.**

Cause: `write_mean` (`io_meandata.F90:2038`) gathers each level to one `root_rank` and
writes it with a single `nf90_put_var`, so compression is single-threaded over the whole
global field no matter how many cores the run has. Measured with the same netCDF stack:
compressed writes run at 12 MB/s against 567 MB/s uncompressed.

Compression time is fixed by field size while everything else scales, so this worsens with
resolution and node count.

Minor: `io_meandata.F90:1943` passes `deflate = 1` unconditionally, so
`compression_level = 0` leaves compression enabled at level 0 instead of switching it off.

koldunovn:
**TL;DR**

- Output and restarts are gathered onto one CPU per file and written single-threaded. This adds a
  collective path where every writer holds a contiguous slice and writes it directly. The file is
  unchanged.
- **Off by default** (`parallel_write = .false.`), `&io_parallel` is optional, existing namelists
  need no edit and behave exactly as before.
- NG5 whole-run wall clock: 1472 → 693 s at 1024 CPUs, 950 → 128 s at 8192. Per-timer speed-ups on
  the write itself are 18–24× for output and 7.4× for restarts.
- Output is also **chunked deliberately** — chunks aligned to the writer block and `chunk_levels`
  rather than netCDF defaults — which helps profile reads and zarr conversion independently of
  speed.
- ⚠️ **Switch it on only at a file boundary** (new year or month). The collective path cannot
  append to an output or restart file the gather path created; see "Switching the flag on
  mid-period" below.
- Correctness: vertex fields bit-identical, element fields within 2 ULP by a documented
  ownership rule, 190 compatibility checks pass.

---

Output and restarts are currently gathered onto one CPU per file, which compresses and
writes single-threaded no matter how many CPUs the run has. This adds a collective path:
every writer holds a contiguous slice of the global field and writes it directly, so the
file is unchanged but the work is shared. Off by default. Follows #968.

## Speed-up

`runtime output` and `runtime restart`, maximum over CPUs, gather against the collective
path at its best writer count.

| | CPUs | gather | collective | |
|---|---|---|---|---|
| **NG5 output** | 1024 | 813.8 s | 34.3 s | **23.7×** |
| | 4096 | 821.4 s | 44.8 s | **18.3×** |
| | 8192 | 870.4 s | 47.2 s | **18.4×** |
| **NG5 restarts** | 1024 | 1037.3 s | 139.9 s | **7.4×** |
| **core2 output** | 128 | 4.04 s | 1.15 s | 3.5× |
| **core2 restarts** | 128 | 4.83 s | 2.92 s | 1.65× |

NG5 was run with 12 streams as 6-hourly means, core2 with 9 as daily means. With both output
and restarts collective, the NG5 wall clock fell from 1797 s to 1094 s.

Restarts gain less because their baseline is already partly parallel: one file per variable,
17 for the ocean, each with its own writer CPU. This shares the work inside each file, not
across them.

## Namelist

New optional group in `namelist.config`. Absent, everything behaves as before.

```fortran
&io_parallel
parallel_write = .true.    ! collective writes instead of gather-to-one
n_writers      = 256       ! CPUs that share the writing; 0 = as many as allowed
chunk_levels   = 8         ! levels per output chunk; 0 = all
/
```

For output, about a quarter of the CPUs is a good starting point: the fastest values measured
were 256 writers at 1024 CPUs, 1024 at 4096 and 512 at 8192. From an eighth of the CPUs
upward everything stayed within 20 % of the best. Below 64 writers it degrades clearly, 76.7 s
against 44.8 s at 4096 CPUs.

Restarts peak lower, at 64 writers on 1024 CPUs, and are slower again at 128 — 17 files
written at once means 64 writers are already about 1100 concurrent streams. A single knob
therefore cannot suit both paths, and one of them will be mistuned. Separating them is the
obvious next step and is not in this PR.

`chunk_levels` sets the vertical extent of a chunk. The horizontal extent is fixed to the
writer's block, `ceil(nod2D / n_writers)`, and the time extent to one record, since output is
appended one timestep at a time. Chunk size is `chunk_levels × block × bytes`.

Taller chunks suit profiles and zarr, flatter ones suit maps at a single depth. On NG5 at 256
writers, `chunk_levels` of 1, 8 and 69 gave 39.0 s, 35.0 s and 33.7 s, with volumes within
0.7 % of each other.

## Compilation

No new dependencies: `nf90_create_par`, `nf90_var_par_access` and `nf90_def_var_chunking` are
already in netcdf-fortran, and the redistribution uses `MPI_Alltoallv`.

The collective path needs netCDF built against parallel HDF5. Builds without it still compile
and link, because `nc_create_par` is present in the API either way; the failure appears at
runtime, and only if `parallel_write = .true.` is set. Default is `.false.`, so a serial
netCDF build is unaffected.

One version quirk: netcdf-fortran 4.5.3 ships `nf90_set_var_chunk_cache` in the library but
does not export it from the F90 module, so the F77 entry point is declared locally.

## Correctness

Nodal fields are bit-identical to the gather path on every mesh tested. Element fields differ
in at most 2 ULP: an element belongs to a CPU if any of its vertices does, and the two paths
resolve that duplicate differently: the gather path takes the highest-numbered CPU holding it,
this one the CPU owning the element's first vertex (`myInd_elem2D_shrinked`). On NG5 restarts,
19 of 23 variables are bit-identical and the other four are element fields differing by
1e-16 to 1e-12.

Files are unchanged in structure: same dimensions, attributes, `_FillValue` and data model,
with `nf_classic_model` kept in the parallel create. 190 checks pass against gather-written
references, covering `ncdump`, xarray both eager and lazy through dask, and CDO 2.2.2, where
`cdo diffn` between the two `timmean` results is empty.

## Implementation

The partitioner does not renumber the mesh, so the vertices of a subdomain are scattered through
the global numbering instead of forming a block. On NG5 at 1024 CPUs a CPU owns about 7200
vertices, and they sit in roughly 130 separate groups of consecutive numbers, about 58 vertices
each, spread over nearly the whole range.

A netCDF write covers one consecutive range: a start index and a count. Writing a CPU's
vertices directly would therefore need one call per group, about 130 of them. And under
collective access every CPU must make the same number of calls, while here the number of
groups differs from CPU to CPU, 116 to 134.

The field is rearranged before writing. One `MPI_Alltoallv` sends each CPU's values to the
writer responsible for that range of global vertex numbers, so every writer ends up holding one
contiguous range and writes it in a single call. The rearrangement depends only on the mesh
partition and the number of writers, so it is computed once at start-up and reused for every
write.

Writers are taken at a fixed stride through the CPU list rather than as the first N CPUs, so
they are spread across compute nodes. Chunk boundaries along the horizontal dimension coincide
with writer boundaries, so no chunk is shared between two writers and no compressed data has
to be exchanged and repacked between CPUs while writing.

Each variable is written in one call rather than one per level: worth 1.35× on NG5, and it
avoids repacking a compressed chunk once per level when a chunk spans several.

Asynchronous writing is switched off on this path. The write is a collective operation, and
issuing it from the background thread deadlocks against the main thread's own collective calls;
HDF5 here is also built without thread safety.

New file: `src/io_redistribute.F90`, the schedule and the exchange, with no dependency on
`MOD_PARTIT` so it can be unit-tested. Modified: `io_meandata`, `io_fesom_file`,
`io_netcdf_file_module`, `io_restart`, `gen_modules_config`, `gen_model_setup`.

Sorry for the size of this one. The gather path is untouched and remains the default, so the
diff is additive rather than a rewrite, but it does span both the output and the restart
writers because they share `io_fesom_file`.

### Switching the flag on mid-period

The collective path cannot append to an output or restart file that the gather path created.
Turning `parallel_write` on part-way through a period whose file already exists fails at the first
write:

```
error in line 339 io_meandata.F90              NetCDF: HDF error   (output)
error in line 614 io_netcdf_file_module.F90    NetCDF: HDF error   (restarts)
```

Switch at a file boundary instead -- a new year, or a new month with `filesplit_freq` -- so the file
is created by the collective path. A cold start is unaffected, and so is any chain that turns the
flag on between segments that begin a new file. The namelist entry added in this PR says so too.

The mechanism is not established. Compression is not the whole story: the same failure occurs with
`compression_level = 0`, so it is not only that the gather path writes one chunk spanning the whole
node dimension while the collective path writes per-writer slabs. A chunk-layout check at
`par_open_existing` would turn this into a clear message and can follow; it is left out here to keep
the diff to what has been tested.

This restriction would disappear if each restart event were written to its own file instead of
appending records to an existing one, which is #279.

## Not addressed

Restart reads still gather onto one CPU and scatter. A parallel read is a natural follow-up, but
not by inverting this schedule: a write partitions the global index space and a read does not,
because a rank needs every entry it holds locally and its neighbours need the same ones. For
elements the difference is not just the halo -- an element inside `myDim_elem2D` whose first node
is not owned is in neither the shrinked set nor the halo receive list, so an inverted schedule
would leave it unfilled. The read needs its own scatter schedule. Implemented and measured
separately (NG5/8192: 141 s -> 40 s, bit-identical), and it will come as its own PR on top of this
one.


koldunovn:
## Why not just `ASYNCHRONOUS_IO_THREADS=ON`?

**TL;DR**

- **All ×-figures below are against the same gather baseline unless it says otherwise.** On output
  at the cadence measured here: threads 1.90× at NG5/1024 and 3.42× at NG5/8192, this PR 2.12× and
  7.40×. So this PR is ahead of the threads by 1.12× at 1024 and 2.16× at 8192.
- The existing thread mechanism is not a straw man, and on restarts it beats this PR.
- **Which one wins depends on how often you write.** Threads hide a write behind the compute that
  follows it, so they win when writes are far apart and lose when they are close together. The
  crossover at NG5/8192 is roughly one output every 1900 steps. **At monthly output, threads win.**
- Threads are not free: `runtime ocean` grows 14–23 % at 8192, because the writer thread shares a
  core with a compute rank that everyone waits for at each halo exchange.
- This PR also **chunks the output deliberately** — chunks aligned to writer blocks and
  `chunk_levels` instead of whatever netCDF defaults to. On NG5 that is `[1, 8, 28918]` against
  `[1, 2, 321865]`, which matters for reading profiles and for zarr conversion, independently of
  speed.
- The two differ in what they need from the machine: a library version that can be checked, against
  runtime thread behaviour that has been switched off six times upstream and fails by hanging.

---

FESOM already has a way to take output off the critical path. `async_threads_module` puts each
output stream, each restart file and each forcing reader on its own pthread. It is `OFF` by default
for every platform in `src/CMakeLists.txt` and was OFF in every build this PR was measured against.
So the fair question is whether this PR duplicates it.

Measured. Two binaries from the same commit (`4b0aea2b`), differing only in
`-DASYNCHRONOUS_IO_THREADS`. NG5, 7.4M surface nodes, 69 levels, 360 steps, 12 streams at 6-hourly
means, `compression_level = 1`.

The headline is `Runtime for all timesteps`, not `runtime output`. With the flag ON that timer
measures a pthread dispatch, so it collapses towards zero whether or not the write was really
hidden. Only wall clock compares across the two builds.

**Output only**, wall seconds. Speed-ups in brackets are against the gather column on the same row:

| ranks | gather | async threads | this PR | async + this PR |
|---|---|---|---|---|
| 1024 | 1472.1 | 775.2 (1.90× vs gather) | **693.1 (2.12× vs gather)** | 693.3 |
| 8192 | 950.7 | 277.9 (3.42× vs gather) | **128.4 (7.40× vs gather)** | 140.9 |

This PR against the threads, which is the comparison this whole comment is about: **1.12× at 1024
and 2.16× at 8192**.

**With restarts** (2 restart events, 46 restart files):

| ranks | gather | async threads | this PR |
|---|---|---|---|
| 1024 | 1703.8 | **815.2** | 822.8 … 862.1 |
| 8192 | 1372.1 | 294.1 | **247.4** |

Controls reproduce earlier runs within 0.2 %: `off_gather` 1472.1 against 1473.1 measured before at
1024, 950.7 against 948.6 at 8192, `off_par` 128.4 against 127.7 at 8192.

## Output frequency decides this, so state your cadence before quoting a number

A thread hides a write behind the compute that comes after it, and `join` is at the next output
event of the same stream (`io_meandata.F90:3227`). So the mechanism works exactly while

> compute between two outputs > time to write one output

Both sides of that are measurable here. Compute per step is wall minus write on the collective
path; write per event is the gather writer's own timer divided by the number of events:

| | compute/step | write/event | crossover |
|---|---|---|---|
| NG5 / 1024 | 1.84 s | 406 s | ~220 steps between outputs |
| NG5 / 8192 | 0.23 s | 436 s | ~1900 steps between outputs |

This benchmark writes every 180 steps. At 1024 that is just under the crossover, which is why
threads get 1.90× — nearly the whole write is hidden. At 8192 it is an order of magnitude under,
and the ceiling shows up directly in the timer: `runtime output` max falls 811.8 → 12.5 at 1024
(dispatch only, the thread was done) but only 872.7 → 78.9 at 8192, where those 78.9 s are the main
thread blocked in `join`.

**Turn it round and threads win.** At `dt = 240 s`, monthly output is 10800 steps between events,
far past the 1900-step crossover at 8192. The write is then hidden completely and this PR's
remaining ~46 s per event is still on the critical path. If you write monthly, prefer the threads —
subject to the CPU cost below and to what they need from the machine.

The crossover moves the wrong way as machines grow: write per event is fixed by the data volume and
one writer, while compute per step falls with every core added. Whatever cadence is safe today
becomes unsafe at twice the core count. This PR removes the term instead of hiding it, which is why
its lead over the threads widens with rank count — from 1.12× at 1024 to 2.16× at 8192.

## Hiding is not free

`runtime ocean` mean, gather → async threads:

| | 1024 | 8192 | 8192 with restarts |
|---|---|---|---|
| gather | 570.5 | 64.3 | 64.4 |
| async | 573.7 (+0.6 %) | 73.5 (**+14 %**) | 79.1 (**+23 %**) |

128 ranks per node on 128 cores, so a writer thread shares a core with a compute rank, and that rank
is in every halo exchange. Noise at 1024, real at 8192, and it grows with the number of live threads
— 46 restart plus 12 output plus 12 forcing at 8192. At a sparse output cadence the threads are
active a smaller fraction of the time, so this cost shrinks too; it has not been measured there.

## Where the threads win outright

On restarts. Cost that turning restarts on adds to wall:

| | 1024 | 8192 |
|---|---|---|
| gather | +231.7 | +421.4 |
| async | **+40.0** | **+16.2** |
| this PR | +129.5 … +169.0 | +119.0 … +181.4 |

Restarts are far apart by nature, so they sit on the good side of the crossover and threads hide
them almost entirely. Part of this PR's cost is a knob problem rather than a design problem: one
`n_writers` serves output and restarts, their optima differ, and the restart numbers above were
taken at a count chosen for output. Splitting the knob is the obvious follow-up.

The collective column is a **range** because both endpoints come from pairs of code-identical runs
— `parallel_write` force-disables async, so ON and OFF take the same path — that differed only in
the write timer, by up to 20 %:

- 1024 with restarts: `runtime restart` 166.2 against 128.0, wall 862.1 against 822.8
- 8192 output only: `runtime output` 45.7 against 58.1, wall 128.4 against 140.9

Worth knowing before anyone reads a 5–10 % difference on this path out of a single run pair.

## A second benefit, independent of speed: the output is chunked on purpose

The gather path does not choose a chunk layout, so netCDF picks one. This PR sets chunks to the
writer block along the horizontal dimension and to `chunk_levels` along the vertical, because that
is what makes a compressed collective write possible at all. The layout that falls out is better
for reading. NG5/1024, 256 writers, same fields:

| | gather | this PR |
|---|---|---|
| `sst` (2D) | `[1, 925361]` | `[1, 28918]` |
| `temp` (3D) | `[1, 2, 321865]` | `[1, 8, 28918]` |

Reading one profile out of `temp` touches 643 730 values per chunk on the gather layout and 231 344
on this one, and the chunks are regular rather than a function of how netCDF felt about the
dimensions. That helps point and profile extraction, subsetting, and conversion to zarr, none of
which this PR is about. `chunk_levels` is a namelist knob (default 8) precisely because the right
trade-off between maps and profiles is the user's to make.

## What each one needs from the machine

`ASYNCHRONOUS_IO_THREADS`:

- `MPI_THREAD_MULTIPLE` actually provided (`io_meandata.F90:4004`, `io_fesom_file.F90:206`), else it
  silently turns itself off
- netCDF called from a pthread against a non-threadsafe HDF5. Of the 12 HDF5 builds in
  `/sw/spack-levante`, every one with `Parallel HDF5: ON` has `Threadsafety: OFF`, and every one
  with `Threadsafety: ON` has `Parallel HDF5: OFF`. `io_netcdf_workaround_module` exists to give
  each consumer a distinct rank so no process ever has two netCDF threads.
- no OpenMP. `b0c88285`: "Fesom jobs get stuck on levante if both I/O multithreading and openmp are
  active", seen in IFS-FESOM and standalone. Its `-qno-openmp` workaround is not in main.
- not XIOS. `b1776ae7` forces the flag OFF when `FESOM_WITH_XIOS=ON`.

This PR:

- netCDF-4 built against parallel HDF5. `nf90_create_par` is exported by netcdf-fortran 4.5.3 on
  Levante and `nc-config --has-parallel4` says yes.
- `Parallel Filtered Dataset Writes: ON`, HDF5 >= 1.10.2. Levante has 1.12.1 with it on.
- ordinary collectives from the main thread. `MPI_THREAD_SINGLE` would do.
- no pthreads, no hostname parsing, no per-rank I/O role assignment.

Both add a requirement. The difference is what happens when it is missing: a serial-only netCDF
fails at the first `nf90_create_par` with a message, the other one hangs.

## Backward compatibility

`&io_parallel` is read with `iostat` and rewound if absent (`gen_model_setup.F90:86-91`), and
`parallel_write = .false.` by default (`gen_modules_config.F90:125`). Existing `namelist.config`
files need no edit and behave exactly as before. Output and restarts from the ON build are
bit-identical to the OFF build at NG5/1024 — 12 streams, 15.928 GB, `max|diff| = 0`.

## One limitation found while measuring this

The collective path cannot append to an output or restart file the gather path created. Switching
`parallel_write` on part-way through a period whose file already exists dies at the first write with
`NetCDF: HDF error` (`io_meandata.F90:339` for output, `io_netcdf_file_module.F90:614` for
restarts). Turning compression off does not help, so it is not the chunk geometry alone. Switch at a
file boundary — a new year or month — and the file is created by the collective path with matching
layout. Details in the PR description.

This would disappear on its own if each restart event got its own file instead of appending records,
which is #279 (open since 2022, with #485 and #617 closed into it).

## A profiling trap worth knowing about

`runtime forcing` at 1024 goes from 382.1 s to 3.1 s when the writer stops being the bottleneck — by
either mechanism, and neither has anything to do with reading forcing. That time was never forcing
work. It was every other rank waiting at the forcing broadcast for the one rank still writing.
`runtime ice` max does the same, 459.4 → 63.7.

Anyone profiling FESOM at scale on the gather path is likely attributing writer wait to whichever
collective happens to come next.

JanStreffing:
Can you make a TL TL DR? Should we keep both, or just one way? Between Serial, Async, MultIO & XIOS this is the fifth output scheme in the model. Would it make sense to have this one replace Async instead?

koldunovn:
@JanStreffing We have to talk about that. I would get rid of Async, as it was constantly causing problems and off by default on many machines. But I think we need more people/opinions to make a decision. What we can't do is to run with serial on high resolution setups ever - this is enormous loss of computing time.

sebastianbeyer:
I agree, @koldunovn , async in its current form seems like a dead end, not worth keeping. 
I'd like to experiment with [PIO](https://github.com/ncar/parallelio) + NetCDF4P and maybe look at [scorpio](https://github.com/E3SM-Project/scorpio) also, but might take a bit of time.

sebastianbeyer:
Looking at this again, I think it's pretty nice already and using libraries like e.g. PIO we do not gain much (maybe some cleaner interfaces, but at the cost of additional dependencies)!
I'd like to see how close we can get the netcdf restarts to the raw restarts (ofc raw will always be faster, it requires no global 'coordination'), but if we can get somewhat close (within 2-3x of raw) I personally would then choose netcdf restarts over raw, because of the PE-independence, easier handling of files (10s not 1000s) and possibly compression. Ofc that also requires them to be fully reproducible, same as raw (no missing state in netcdf vs raw, see e.g. https://github.com/FESOM/fesom2/pull/965)
(I am also for having only a single restart per file, because it would simplify quite some things as mentioned in your comment @koldunovn )

JanStreffing:
@patrickscholz Will make some tests before we merge this.

JanStreffing:
I did a 1-year test with fairly heavy daily output. Its not as clear as I would have hoped:


leg | output | restart | model (ice+oce) | wall | node-hours | volume
-- | -- | -- | -- | -- | -- | --
serial | 5389.2 s | 283.7 s | 5090.9 s | 3:02:22 | 15.2 | 144 GB
parwrite | 1146.0 s | 173.3 s | 5092.5 s | 1:52:23 | 9.4 | 112 GB
async | 230.7 s | 14.3 s | 5152.8 s | 1:35:22 | 7.9 | 144 GB

I'm afraid removal of the old async output would cause performance degradation in such cases. It may well be, that with better parameters one can get a lot more out of parwrite. But in contrast do async it would need tuning. Maybe if we come up with a way to have the code automatically determines optimal parameters?

JanStreffing:
Well and its missing half the element output. 😬 
grafik
async for comparison
grafik

koldunovn:
Oh, shit...

koldunovn:
@JanStreffing Confirmed, thanks — real bug, fixed in af7ff29c.

The writer builds two redistribution schedules, nodal and element, and passed the same `n_writers` to both. That request is clamped to `n_global/REDIST_MIN_BLOCK` (4096), and `elem2D ≈ 2·nod2D`, so it could be cut for nodes and left alone for elements — the two schedules then sit on different ranks. The file is opened over the *nodal* writer set and the put gates on that same predicate, so every element block held outside it was never written and stayed at the fill value. Silently: each schedule is internally consistent, and every writer that did call `nf90_put_var` wrote a correct slab.

It triggers when `n_writers > nod2D/4096`. On core2 that is `> 30`, and 29 of 59 element blocks are lost — the half you saw. Nodal fields are never affected.

The fix builds the element schedule with the count the nodal one actually got (`nod2D <= elem2D`, so it always survives the element clamp unchanged), in all three places that build such a pair — output, restart write, restart read — and each now asserts the two sets agree instead of assuming it. Verified on core2/128: pre-fix at `n_writers = 64` leaves 49.2 % of `u` and `v` at the fill value, in 29 gaps of exactly one element writer block; with the fix, 64 and 30 are both bit-identical to the gather reference across all nine streams.

koldunovn:
On async vs parwrite:
- I did not state it will be always faster - writing in parwrite stops the flow, so writing alongside computations as with async or XIOS will often win - and in particular on small meshes.
- parwrite offer: in my opinion less complicated solution, that can be ON by default. And it has chunks.

JanStreffing:
I did a few tests tuning the parwrite parameters now that the element output is fixed:


leg | stripe | writers | output | restart | s/record
-- | -- | -- | -- | -- | --
serial | 1 | — | 452.9 | 682.1 | 15.10
serial | 32 | — | 456.5 | 632.0 | 15.22
async | 1 | — | 19.6 | 2.8 | 0.65
async | 32 | — | 20.0 | 2.2 | 0.67
parwrite | 1 | 51 | 106.9 | 379.5 | 3.56
parwrite | 32 | 51 | 66.8 | 111.4 | 2.23
parwrite | 32 | 32 | 65.0 | 105.0 | 2.17
parwrite | 32 | 16 | 86.1 | 106.5 | 2.87
parwrite | 32 | 8 | 118.9 | 101.4 | 3.96

I would suggest, if we want to limit the number of IO methods, perhaps serial should be the one to go? Clearly both async and parwrite have their use cases. And I doubt we would have trouble these days getting one of these two to run on any HPC.

koldunovn:
Serial is useful for testing. I would rather keep all of them until something begin to cause troubles or will be hard to maintain.

JanStreffing (approved):

PR #977, opened 2026-08-13
Salinity as an anomaly state (USE_SALT_ANOMALY): removes the float32 salinity drift and the NAC mean-state shift

koldunovn:
## What

Adds `USE_SALT_ANOMALY` (CMake option, off by default): the salinity state is
stored as the anomaly `S - S_ref_anomaly` (35 psu, named parameter in
`o_PARAM`). At |S-35| < ~2 the float32 spacing is 30-250x finer than at S~35
(ulp 1.9e-6 psu), so the slow increments that float32 accumulation absorbs
become representable. Every consumer of absolute salinity adds the offset back
at its gather point: EOS (`densityJM_components`), `sw_alpha_beta`, the
ice-ocean salinity gather, `ref_sss_local`/SSS restoring, KPP surface buoyancy
flux and double diffusion, the `dens_flux` diagnostic, and the surface
dilution term in the salinity BC (`+S_ref*water_flux*is_nonlinfs` - the exact
analogue of the `sval*water_flux` term in the temperature BC: freshwater
enters carrying S=0, i.e. anomaly -S_ref). Clip and blow-up bounds shift by
-S_ref.

Compatibility:
- **Restarts**: files keep what the state holds; after `read_initial_conditions`
  the representation is detected from the global salinity maximum (absolute
  peaks near 41 psu, anomaly near 6). An absolute-salinity restart - e.g.
  warm-starting from an existing spun-up state - is converted in place
  (including AB history); an anomaly restart chains unchanged.
- **Output**: mean streams gained a per-stream additive offset applied at
  accumulation; `salt` and `sss` register `S_ref_anomaly`, so NetCDF output
  stays absolute psu regardless of the build.
- **Guards**: configurations with absolute-salinity consumers not yet
  offset-corrected (SPP, cavities, icebergs, age tracer, transient tracers,
  KPP nonlocal fluxes, 3D climatology relaxation) refuse to start with a clear
  message.

## Why

20-year CORE2 attribution twins against the DP reference (cold start 1948,
identical namelists, single-job runs; the campaign behind the paper draft):

| run | global-mean salinity drift @ year 20 |
|---|---|
| six SP arms (reference, different rank count, FP64 global reductions, CG fix, sub-ulp-perturbed control) | **-1.87 mpsu** (all identical +-0.02) |
| SP + FP64 accumulation of the tracer state | **+1.85 mpsu** (mirror image) |
| DP + tracer state rounded to float32 each step (2-yr probe) | -0.135 mpsu in 2 years |
| **SP + this option** | **+0.064 mpsu** (= vanilla / 29) |
| DP control (512 vs 128 ranks) | +-0.001 mpsu |

The decomposition: float32 salinity error = state-accumulation absorption
(negative) + increment-arithmetic rounding at |S|~35 (positive). The anomaly
representation scales BOTH by the ulp ratio (~30x) - the residual slope
(+3.2 upsu/yr) times 31 reproduces the ledger arm's slope.

It also heals the mean state: the 1952-1962 SP-DP climatological SST
difference drops from 0.069 K rms / **1.77 K max** (the NAC/NWC shift) to
0.024 / 0.38 K, SSS from 0.019 / 0.36 psu to 0.006 / 0.12 (DP-DP floor:
0.0074 K / 0.20 K). A DP run with ONLY the tracer state rounded to float32
reproduces the full SP-DP year-2 signal - the mean-state shift and the drift
share the one mechanism this option removes.

Figures (paper repo, you have access): `figures/fig_twin_drift.pdf`,
`figures/fig_twin_maps.pdf` in koldunovn/fesom2-sp-paper.

## Validation

- 5-day cold smoke vs an unmodified build: day-5 state differences at weather
  level (global-mean salt agrees to 2.4 upsu), ranges identical, no guard hits.
- 5-day warm start from an absolute-salinity restart (the 1967 state of the
  DP hindcast): auto-conversion path.
- 2+3-day restart chain: anomaly restart read-back path.
- The 20-yr twin + climatology numbers above (run at cd60ac69..896cae28,
  reproduced by this branch's code path).

## Open design questions

1. `S_ref_anomaly = 35.0`: value and name OK? (Any constant near the ocean
   mode works; 35 keeps the mental arithmetic easy.)
2. Compile-time option (mirrors `USE_SINGLE_PRECISION`, zero runtime cost) vs
   a namelist switch - preference?
3. Restart auto-detection by global max: acceptable, or would you rather have
   an explicit netCDF attribute on the restart files?
4. CI: worth adding an SP+USE_SALT_ANOMALY leg to the precision test matrix
   from #940?
5. The temperature state has the same absorption mechanism (ulp(20 C) ~ 2e-6);
   a T_ref anomaly would be the natural follow-up if we see it matter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ Makefile.in
 mesh_part/build
 test/output_pi
 .claude/
+build_*/
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -81,6 +81,8 @@ message(STATUS "USE_ICEPACK: ${USE_ICEPACK}")
 
 option(USE_SINGLE_PRECISION "Build FESOM in single precision (WP=4)" OFF)
 message(STATUS "USE_SINGLE_PRECISION: ${USE_SINGLE_PRECISION}")
+option(USE_SALT_ANOMALY "Store the salinity state as the anomaly S - S_ref (finer float32 spacing; see o_PARAM::S_ref_anomaly)" OFF)
+message(STATUS "USE_SALT_ANOMALY: ${USE_SALT_ANOMALY}")
 
 
 # option to trigger building a library version of FESOM
@@ -297,6 +299,10 @@ endif()
 
 if(USE_SINGLE_PRECISION)
    target_compile_definitions(${PROJECT_NAME} PRIVATE USE_SINGLE_PRECISION)
+endif()
+
+if(USE_SALT_ANOMALY)
+   target_compile_definitions(${PROJECT_NAME} PRIVATE USE_SALT_ANOMALY)
 endif()
 
 if(CVMIX)
--- a/src/fesom_module.F90
+++ b/src/fesom_module.F90
@@ -139,6 +139,9 @@ contains
       ! EO parameters
       logical mpi_is_initialized
       integer              :: tr_num, n
+#ifdef USE_SALT_ANOMALY
+      real(kind=WP)        :: salt_max_loc, salt_max_glob
+#endif
 
 #if defined (__recom)
       type(tracers_info_type)               :: tracers_info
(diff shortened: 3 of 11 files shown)


suvarchal:
Superseded by #986, which reworks this into a **runtime** namelist toggle (`use_salt_anomaly`, default off) on a single code path — no separate build, bit-identical when off, and bit-identical to this PR when on. It builds on the 4 commits here (authorship preserved). #986 also documents where the small residual DP change comes from (a free-surface advection constancy error, not a missed conversion). Suggest we continue review on #986 and close this one.

PR #978, opened 2026-08-14
Omp reproducable fix

### LINKED ISSUE #444: using OPENMP_REPRODUCIBLE doesn't compile due to missing clause in openmp directives
cannot compile the model with -DOPENMP_REPRODUCIBLE=ON (must be used together with -DENABLE_OPENMP).

error is 
```
The DO directive to which the ordered section binds does not have the ORDERED clause specified.
!$OMP ORDERED
```
for multiple files that use openmp directives. 

Not sure if this is because of openmp version.

and also note for future:

- OPENMP_REPRODUCIBLE compile option is currently independent of ENABLE_OPENMP, it should be only relevant when OPENMP is used.

### LINKED ISSUE #528: broken -DOPENMP_REPRODUCIBLE=ON
compilation is broken with flag OPENMP_REPRODUCIBLE flag on.
core issue is errors like this when reproducible is on:
```
  790 | !$OMP ORDERED
      |             ^
Error: 'ordered' region must be closely nested inside a loop region with an 'ordered' clause
```
fix for instance in code block example would be 
https://github.com/FESOM/fesom2/blob/231c416488007aa622045ab0b43c2c6f9f94a7d5/src/ice_fct.F90#L864-L899

to add ordered to all clauses where we may use OMP_ORDERED like: `!$OMP DO ordered`.

suvarchal:
Fixes #444
  Fixes #528
  
  something popped during SP work, worth a fix. 

  `-DOPENMP_REPRODUCIBLE=ON` has never compiled: 42 worksharing loops contain
  `!$OMP ORDERED` regions but none carried the required `ORDERED` clause, so
  every affected file failed with

      Error: 'ordered' region must be closely nested inside a loop region with
             an 'ordered' clause

  Nothing in the tree exercised this build dimension, which is how it survived.

  ## Changes

  * Add the `ORDERED` clause to the 42 enclosing loop directives, guarded so the
    default build is unaffected (13 files). In `gen_support.F90` (3 sites)
    `REDUCTION(+: lval)` is also dropped on that path — it makes the accumulator
    thread-private, so ordering the region alone would compile without actually
    being reproducible.
  * Fix 2 `ORDERED` regions that opened inside an `if` branch and closed after its
    `end if`, straddling the `else` (`oce_ale_ssh_splitexpl_subcycl.F90`,
    `oce_ale_vel_rhs.F90`). These only surfaced once the clause errors cleared.
  * Fix 14 `ORDERED` regions nested inside an inner `DO`, so they executed more
    than once per iteration of their worksharing loop. That is undefined
    behaviour: the ordering guarantee is lost and threads enter the region
    concurrently. Confirmed as a real data race — two 4-thread runs differed from
    each other while 1-thread runs were identical. Affected loops are now
    serialized (`SINGLE`, or omitting the combined `PARALLEL DO`), matching the
    pattern already used in `gen_support.F90` and `solver.F90`.
  * Serialize two `+` reductions on the reproducible path (`oce_dyn.F90`,
    `oce_ale_pressure_bv.F90`).
  * Reject `OPENMP_REPRODUCIBLE=ON` without `ENABLE_OPENMP=ON` at configure time
    (second point of #444).

  ## Tests

  `tests/openmp/check_omp_ordered.py` + 3 ctests (label `openmp`). It drives the
  real Fortran preprocessor, so it sees exactly the directive stream the compiler
  sees, and needs only a configured tree — no built `fesom.x`. Invariants:

  1. every active `ORDERED` region binds to a loop carrying the clause (47
     violations on the unfixed tree)
  2. no `ORDERED` region is nested in an inner `DO` (13 violations)
  3. neither regions nor clauses leak into the default build, where they would
     serialize loops that should use the OpenMP locks

  New CI job `fesom2_openmp_reproducible.yml` runs the guards, builds with
  `-DENABLE_OPENMP=ON -DOPENMP_REPRODUCIBLE=ON`, and runs the pi integration
  tests with 4 threads. The runtime leg matters: it catches the `sprod(2)` class
  of bug (af17e849), which compiles fine either way — verified by re-introducing
  that bug, which fails the test with `eta_n become NaN`.
  

  Verified default, OpenMP, and OpenMP+reproducible builds all compile clean and
  pass the integration tests.


JanStreffing:
Good bit of progress. Does the OMP actually reproduce as well? Or still not?

suvarchal:
> Good bit of progress. Does the OMP actually reproduce as well? Or still not?

not yet, OMP reproducible option/branch runs and few races are fixed which is necessary for further investigation. but for true reproducibility (it comes from ice parts, rest is reproducable) there is more to come after this in another PR later.

JanStreffing (approved):

PR #981, opened 2026-08-17
solver: always-on SSH CG diagnostics, breakdown guard, allocate guards

suvarchal:
> Stacked on #980 (the `EXTRA_SUCCESS_MARKERS` test hook) — please review/merge that first.

## Why

The SSH conjugate-gradient solver is entirely unobservable, depending on mesh when convergence 
is not hit it wastage is not reported. this is needed for single precision port performance too.
 `FESOM_PROFILING` is **OFF** by default and `fesom.stats` is
only written by the profiler, so in a normal production run there is **no** record of how
many CG iterations a step took, whether the solver converged, or whether it broke down.

Three consequences, all silent today:

* **A solver that stalls at `maxiter` prints nothing.** The loop simply ends and the
  unconverged `d_eta` feeds the barotropic and baroclinic dynamics. Unlike `cfl_z`, which
  has `check_blowup` as a hard backstop, non-convergence has **no** backstop at all.
* **A breakdown poisons the solution with no trace.** The guard on `al = s_old/s_aux` was
  removed in `4eb2f21d` ("Removed debug output in solver."). If `p·Ap` reaches zero or NaN,
  `al` fills the entire `d_eta` vector with Inf/NaN, which surfaces several stages
  downstream — typically as a NaN in `Wvel` — with nothing pointing back at the solver.
  Debugging that from the symptom is expensive; we did exactly that.
* **Cost is invisible.** Iteration count per solve is the quantity that determines how
  expensive the barotropic solve is, and it cannot currently be seen without a special build.

This matters most during **spin-up**, which is precisely when the state is furthest from
equilibrium, SSH gradients are largest, and the solver is most likely to need many
iterations or to hit the cap. A spin-up that is quietly running under-converged solves is
indistinguishable, in today's logs, from a healthy one.

## What this adds

* `src/fesom_rtdiagnostics.F90` — a small registry for per-call scalar quantities
  (iterations, residuals, event counts), MPI-reduced and reported in `fesom.stats`. Mirrors
  the `fesom_profiler` patterns; stores in double precision regardless of the model `WP` so
  SP and DP numbers are directly comparable. Reusable for future counters, not solver-specific.
  The `rt-` prefix is deliberate: this is **runtime** diagnostics (how the model *ran*), kept
  clearly apart from the physical diagnostics in `gen_modules_diag.F90` (`module diagnostics`,
  the `ldiag_*` family in `namelist.io`) which compute geophysical quantities.
* **Always-on counters** on `T_SOLVERINFO`: iterations (last/max/sum), solves,
  non-convergences, breakdowns, last residual and target tolerance.
* **A per-step line in `write_step_info`**, beside the existing `cfl_z` line and reusing its
  `logfile_outfreq` gate — so iteration count is a named quantity in the standard diagnostic
  block, not something that only appears when it misbehaves.
* **An end-of-run summary** in the always-on runtime table:

  ```
   ___SSH CG SOLVER_____________________________________
      solves :                                  96
      iterations mean :                    12.2708
      iterations max :                          16
      non-convergences (maxiter) :               0
      breakdowns (p.Ap<=0) :                     0
      iters with r.z<0 (not SPD) :               0
  ```

* **A rate-limited `WARNING:` at the maxiter hit itself**, carrying `rms(resid)` and the
  target `rtol`. The per-step line only prints every `logfile_outfreq` steps, so a stall in
  between would otherwise still be invisible. Rate-limited (first 5, then every 100th)
  because a stalled solver usually stalls every step.
* **The breakdown guard restored**, counted and logged. It keeps the current `X` and leaves
  the loop — deliberately **no abort**, since a stall on step 1 is legitimate in some
  configurations and an abort would turn a diagnostic case red. Never aborting is exactly
  why the logging is obligatory rather than optional.
* **Two unguarded allocates fixed** in `ssh_solve_preconditioner`, which abort with
  "allocatable array is already allocated" if it is ever built twice.

* **Test coverage, so this cannot rot.** Using the `EXTRA_SUCCESS_MARKERS` hook from #PR0,
  `integration_pi_mpi2` pins five labels
  (`___SSH CG SOLVER`, `non-convergences (maxiter)`, `breakdowns (p.Ap<=0)`, `ssh_cg iters=`,
  `ssh_cg rms(r)=`). These are plain writes with no artifact behind them, so without this a
  diagnostic that stopped being printed would fail nothing — which is exactly how the
  `sprod(2)` defect in this same routine survived. Labels, not numbers, so it stays stable
  across meshes, rank counts and precisions.

  Verified the assertion has teeth: removing the summary block makes the test fail with
  `missing success marker`, and pass again once restored.

## Cost and safety

**Results are unchanged.** Verified bitwise-identical to `main` on `integration_pi_mpi2`
(`nc_diff.py`: `WORST_REL=0.000e+00`). The counters are integer increments on values the
solver already computes; no new MPI. The exit test is on a globally reduced `sprod`, so
every rank exits on the same iteration and the count needs no reduction of its own.

The `FESOM_PROFILING` path costs nothing when compiled out. Full local CTest suite passes.

## A note on the SPD sensor

The `r·z < 0` counter is included because `r·z` cannot be negative for an SPD
preconditioner, so a non-zero count would be direct evidence of a defective `M⁻¹`. It reads
**zero** on both pi and core2, in both precisions — so it currently confirms nothing. It is
kept because it is two lines on an already-reduced value and makes a real failure mode
visible, not because it has found anything.

### DIFF
--- a/src/MOD_DYN.F90
+++ b/src/MOD_DYN.F90
@@ -19,6 +19,20 @@ TYPE T_SOLVERINFO
     real(kind=WP) :: droptol = 1.e-8
     real(kind=WP)  :: soltol  = 1e-5
     real(kind=WP), allocatable   :: rr(:), zz(:), pp(:), App(:)
+    !___________________________________________________________________________
+    ! Per-run solver diagnostics. Deliberately NOT in WRITE/READ_T_SOLVERINFO
+    ! below: these are transients, and adding them to the dump would change the
+    ! restart record layout and silently break every existing restart file (the
+    ! read would just consume the wrong bytes).
+    ! The iteration count is identical on every rank -- the CG exit test is on a
+    ! globally reduced sprod -- so none of this needs an MPI reduction.
+    integer       :: iters_last = 0     ! iterations taken by the last solve
+    integer       :: iters_max  = 0     ! worst solve so far
+    integer       :: iters_sum  = 0     ! for the running mean
+    integer       :: nsolves    = 0
+    integer       :: nonconv    = 0     ! solves that hit maxiter
+    real(kind=WP) :: resid_last = 0.0_WP ! rms residual at exit
+    real(kind=WP) :: rtol_last  = 0.0_WP ! target it was compared against
     contains
     procedure WRITE_T_SOLVERINFO
     procedure READ_T_SOLVERINFO
--- a/src/MOD_DYN.F90
+++ b/src/MOD_DYN.F90
@@ -31,6 +31,8 @@ TYPE T_SOLVERINFO
     integer       :: iters_sum  = 0     ! for the running mean
     integer       :: nsolves    = 0
     integer       :: nonconv    = 0     ! solves that hit maxiter
+    integer       :: nbreakdown = 0     ! solves where p.Ap was not positive
+    integer       :: nnegrz     = 0     ! iterations with r.z < 0 (M^-1 not SPD)
     real(kind=WP) :: resid_last = 0.0_WP ! rms residual at exit
     real(kind=WP) :: rtol_last  = 0.0_WP ! target it was compared against
     contains
--- a/src/fesom_module.F90
+++ b/src/fesom_module.F90
@@ -1461,6 +1461,22 @@ contains
         44 format (a33,i15)        !Format OMP threads
         45 format (a33,f15.4,a4)   !Format runtime
 
+        !_______________________________________________________________________
+        ! SSH CG solver summary. Always on -- fesom.stats only exists when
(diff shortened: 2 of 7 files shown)


suvarchal (commented):
checked again

JanStreffing (approved):
Looks good. Have you done some testing on the performance impact of the profiler? If its not measurable, maybe we should default to on?
If it's more than 1-2% probably not.

JanStreffing:
I'd also be interested in some sample output here, so that we can get a feel for what numbers one would expect from a normal run.

suvarchal:
It is very light weight, we should make it default.

PR #984, opened 2026-08-18
solver: symmetric Jacobi preconditioner variant for the SSH CG (opt-out)

suvarchal:
> Stacked on #983 → #981 → #980. Please merge those first.

## Motivation

The SSH free-surface system is solved with **preconditioned Conjugate Gradient**
(`ssh_solve_cg`, `src/solver.F90`). CG is not a general-purpose Krylov method: it is derived
for **symmetric positive definite** systems, and preconditioned CG additionally requires the
**preconditioner `M⁻¹` to be SPD**, because the method works in the `M`-inner product and its
conjugacy and minimisation properties only hold if that is a genuine inner product.

- Hestenes & Stiefel (1952), *Methods of Conjugate Gradients for Solving Linear Systems*,
  J. Res. NBS **49**, 409–436.
- Saad (2003), *Iterative Methods for Sparse Linear Systems*, 2nd ed., SIAM — §9.2 on PCG.
- Golub & Van Loan (2013), *Matrix Computations*, 4th ed., §11.5.
- The routine's own reference: Marshall et al. (1997), JGR **102**, 5753–5766 (MITgcm).

`ssh_solve_preconditioner` says so explicitly in its header: *"We need symmetrization to be
able to use the conjugate gradient method."*

**The coded preconditioner is not symmetric.** Its off-diagonal is

```
M⁻¹(r,i) = −0.5·a_ri / (a_rr·(a_rr + a_ii))
```

whose transpose entry is `−0.5·a_ir / (a_ii·(a_ii + a_rr))`. With `A` symmetric these are equal
**only where `a_rr == a_ii`** — that is, only on a uniform mesh. On any variable-resolution
mesh, which is the reason to use FESOM at all, the requirement the header comment states is not
met. (The coded expression also differs from the formula in that same comment by a factor of
four.)

`precond_variant = 1` implements the textbook symmetrised Jacobi
`M⁻¹ = 2D⁻¹ − D⁻¹AD⁻¹`, whose off-diagonal `−a_ri/(a_rr·a_ii)` is symmetric by construction and
whose diagonal reduces to `1/a_rr` — exactly what the code already uses on the diagonal. So only
the off-diagonal changes.

## How to enable it

A runtime switch in the **existing** `&dynamics_general` group of `namelist.dyn` — no new
group, no new file, no rebuild:

```fortran
&dynamics_general
  ...
  precond_variant    = 1   ! 0 = as coded today (default), 1 = symmetrised Jacobi
/
```

| value | formula | |
|---|---|---|
| `0` **(default)** | `−0.5·a_ri / (a_rr·(a_rr + a_ii))` | as coded since `60b46bdc`; results unchanged |
| `1` | `−a_ri / (a_rr·a_ii)` | textbook `M⁻¹ = 2D⁻¹ − D⁻¹AD⁻¹` |

It is read in `oce_setup_step.F90` into `dynamics%solverinfo%precond_variant` and echoed at
startup next to the other solver settings, so a run's log records which preconditioner it used:

```
     ssh CG soltol  =    1.0000000000000001E-005
     ssh CG maxiter =         2000
     ssh CG precond =            1
```

Note it takes effect on a **cold start**; see #983 for why `soltol`/`maxiter` are restored from
the restart dump on a continuation. The switch itself is per-run, not per-solve — the
preconditioner is built once at the first solve.

## Results

All iteration counts are exact: runs are bitwise reproducible, so no averaging over repeats is
involved.

### pi (NP=2, 96 solves, `step_per_day=96`, DP)

| | mean iterations | max | `runtime oce. solve ssh` |
|---|---|---|---|
| `precond_variant=0` | 12.271 | 16 | 0.1856 s |
| `precond_variant=1` | **8.604** | **11** | **0.1344 s** |
| | **−29.9%** | | **−27.6%** |

### core2 (126,858 nodes, NP=16, 32 solves, `step_per_day=32`)

| | mean iterations | max | `runtime oce. solve ssh` |
|---|---|---|---|
| DP `precond_variant=0` | 138.19 | 164 | 6.264 s |
| DP `precond_variant=1` | **90.75** | **108** | **4.221 s** |
| | **−34.3%** | | **−32.6%** |
| SP `precond_variant=0` | 138.25 | 164 | |
| SP `precond_variant=1` | **91.00** | **108** | |
| | **−34.2%** | | |

The worst solve falls 164 → 108 as well, so this is not an averaging artefact — the peak
per-step cost improves as much as the mean, which also tightens load balance. SP tracks DP
throughout.

Solver wall-clock tracks the iteration count closely, as it must: the preconditioner is built
once (`oce_ale.F90`, `if (lfirst) call ssh_solve_preconditioner`), `precond_variant` is branched
on only inside that build, and the CG loop is the identical code path for both variants —
variant 1 adds no per-iteration cost.

**Total model runtime is unchanged** on these configurations (+0.4% pi, +0.2% core2, both inside
run-to-run spread), because the SSH solve is only 0.7% (pi) and 5.2% (core2) of the total. The
solve's share grows with mesh size, which is one reason the default question is deferred below.

### Why iteration count matters more than the arithmetic suggests

Each CG iteration contains **two global `MPI_Allreduce`** — one for `p·Ap` (`solver.F90:248`)
and one for the `r·z` / `r·r` pair (`:304`) — plus two halo exchanges. Only two further
reduces happen once, at setup, outside the loop.

Those collectives are synchronisation points: every rank must arrive before any can proceed,
and their cost is latency-bound rather than flop-bound, so it grows with rank count instead of
shrinking. That makes the iteration count, not the arithmetic per iteration, the quantity that
limits how the SSH solve scales — an established characteristic of Krylov solvers at scale, and
the reason communication-avoiding CG variants exist at all.

So **a third fewer iterations is a third fewer global synchronisations per timestep.** On the
16-rank runs above that saves ~2 s and disappears into the total; at production rank counts,
where the collectives dominate the solve, the same relative reduction is worth considerably
more. This is why the unchanged total runtime measured locally should not be read as "no
benefit" — it is the regime, not the change, that makes it invisible here.

### Dependence on the time step

The preconditioner's job depends on how diagonally dominant the system is, and that is set by
`dt`: the assembly puts `areasvol/dt` on the diagonal (`oce_ale.F90:1591`) and scales the
off-diagonal by `g·dt·α·θ` (`:1521`), so dominance scales as `dt²`. pi, DP, both rank counts:

| `step_per_day` | dt | NP=2 v0 → v1 | | NP=8 v0 → v1 | |
|---|---|---|---|---|---|
| 24 | 60.0 min | 46.708 → 30.625 | −34.4% | 47.208 → 31.125 | −34.1% |
| 48 | 30.0 min | 24.438 → 16.396 | −32.9% | 25.000 → 16.812 | −32.8% |
| 96 | 15.0 min | 12.271 → 8.604 | −29.9% | 12.781 → 8.812 | −31.1% |
| 192 | 7.5 min | 6.281 → 4.375 | −30.3% | 6.448 → 4.792 | −25.7% |
| 384 | 3.8 min | 3.643 → 2.206 | −39.5% | 3.700 → 2.412 | −34.8% |

Absolute iteration counts are strongly `dt`-dependent, as expected — a 16× smaller time step
costs 13× fewer iterations. **The relative gain, however, is roughly scale-invariant at 30–35%
across the whole range**, so the improvement does not evaporate at the fine time steps
production configurations use. (The `spd=384` row is the least reliable: at 3.6 vs 2.2
iterations, small-integer discretisation dominates the mean.)

Rank count is a second-order effect: NP=8 needs 1–4% more iterations than NP=2 in every
configuration, consistently signed, which is the expected consequence of more partial sums in
the global reductions. It is far below the `dt` (13×) and variant (30%) effects, but it does mean
iteration counts are not exactly reproducible across decompositions.

## The solutions differ, and that is expected

In exact arithmetic a preconditioner cannot change the answer — CG with any SPD `M⁻¹` converges
to the same `x*`. It changes the path, not the destination. But CG stops at the first iterate
satisfying `rms(r) < rtol`, and two preconditioners cross that threshold at **different
iterates**, each legitimately within tolerance.

After a **single timestep** on pi (NP=2, `step_per_day=96`, DP, `soltol=1e-5`; 16 iterations at
variant 0, 11 at variant 1), comparing `ssh` across all 3140 surface nodes:

| | |
|---|---|
| `ssh` range (variant 0) | -0.039568 ... +0.029651 m (span 69.2 mm) |
| max abs(v1-v0) | **5.34e-06 m** |
| rms abs(v1-v0) | 2.08e-07 m |
| mean (v1-v0) | **-1.12e-10 m** |
| rms relative difference | 6.0e-05 |
| max abs difference / range | 7.7e-05 |

Micron-scale differences on a field spanning 69 mm, and a mean difference of 1e-10 m -- i.e.
**no systematic bias**, just a different rounding of the same answer.

The relative difference comes out a few times `soltol` rather than below it, which is what the
criterion actually promises: `soltol` bounds the **residual**, and the corresponding error in
the *solution* is that residual amplified by the operator's conditioning. Both iterates satisfy
the same residual test; they simply stop at different points inside it. The quantities to judge
are the absolute difference and the absence of bias, both shown above.

Worth stating plainly for anyone assessing CI impact: this difference is real and persists — a
separate tolerance sweep showed that tightening `soltol` does **not** buy bitwise agreement over
a model run, because the intervening nonlinear dynamics erase the tolerance dependence. Any
future default change therefore requires regenerated ground-truth data, not a tighter tolerance.

## This was measurable only because of the stacked PRs

None of the numbers above could have been obtained on `main`. Iteration counts, non-convergence
and breakdown counters and the per-step residual-vs-target line come from **#981**; the
`soltol` / `maxiter` knobs that made the tolerance and time-step sweeps possible come from
**#983**; the test-marker hook that pins the diagnostics comes from **#980**.

That instrumentation also earned its keep during this work: an experimental third formula —
the one the header comment specifies — was found to *diverge* on core2, running 2000 iterations
with the residual growing and blowing the model up at step 2. The solver diagnostics named the
cause immediately:

```
WARNING: ssh CG did not converge in 2000 iters; rms(resid)= 30888.95  target rtol= 19.57
```

Without them this presents as a step-2 blow-up whose built-in advice is "increase
step_per_day" — pointing at the time step rather than the solver. That formula is **not** part
of this PR; it is mentioned only because it shows the symmetry, rather than the magnitude, is
the load-bearing part of the fix.

## Default stays `0`

This ships **opt-in**, and the default is unchanged:

- `precond_variant=0` is **bitwise-identical** to `main` (`nc_diff.py`: `WORST_REL=0.000e+00`).
- Full local CTest suite passes.

### Namelist compatibility, verified in both directions

The three new keys (`soltol`, `maxiter`, `precond_variant`) are added to an existing group as
locals with defaults matching the previously hard-coded values, so neither direction of
version skew regresses an existing setup. Both were tested rather than assumed:

One caveat worth stating: that tolerance is compiler behaviour, not a language guarantee, and
this was gfortran.

Making variant 1 the default is a genuine question, but it should be settled on
high-resolution meshes at production rank counts — ng5-class, well beyond core2 — where the
solve is a much larger share of runtime and where the CG's two `MPI_Allreduce` per iteration
make a third fewer iterations a third fewer global reductions. We would rather take that
decision on that evidence than on pi and core2.

Meanwhile this PR is worth merging as it stands: it is a no-op by default, it makes the SPD
requirement explicit in code rather than only in a comment, and landing it unblocks the
single-precision work in #940 that sits on top of this stack.


koldunovn:
**TL;DR** — `precond_variant=1` holds at production scale. **−33 to −39 % iterations** on four
meshes up to NG5 (7.4 M nodes), and in-model **−10.5 % of the whole step on CORE2/GPU at 16
nodes**, −3.1 % on NG5/CPU at 16 384 ranks. `ms/iter` is unchanged, so the gain is entirely
iteration count. A one-year CORE2 twin shows no drift and no bias. The **4× magnitude** buys the
speed; the **symmetry** is what keeps it convergent.

We ran into the same asymmetry in early August, in our C++/Kokkos port of FESOM2 (its SSH solver
is a term-for-term port of `solver.F90`). We patched only the symmetry, measured −7 % iterations
on CORE2, and left it there. This PR prompted us to re-check on the meshes we actually run, which
changed the conclusion.

Method: we dumped the SSH matrix, right-hand side and starting guess from a running model at one
timestep on each mesh, then re-solved that same system offline with each preconditioner. The
matrix is identical across the comparison; only the preconditioner changes. The solve is
deterministic, so these iteration counts are exact rather than averaged.

Four forms of the off-diagonal, all sharing the diagonal `1/d_i`:

- **as-coded** `−0.5·a_ij/(d_i(d_i+d_j))` — asymmetric, and 4× smaller than the header comment
- **var 1** `−a_ij/(d_i·d_j)` — this PR. Symmetric *and* at the header's magnitude: both defects
  fixed
- **sym-only** `−0.5·a_ij/(√(d_i d_j)(d_i+d_j))` — symmetric, same spectrum of `M⁻¹` as as-coded,
  keeps the 4× under-scaling. Fixes symmetry and nothing else
- **header** `−2·a_ij/(d_i(d_i+d_j))` — the formula in the routine's own comment. Right
  magnitude, still asymmetric

| mesh | 2-D nodes | np | as-coded | **var 1** | sym-only | header |
|---|--:|--:|--:|--:|--:|--:|
| CORE2 | 126,858 | 1 | 126 | **77 (−38.9 %)** | 117 (−7.1 %) | diverges |
| farc | 638,387 | 32 | 238 | **153 (−35.7 %)** | 230 (−3.4 %) | diverges |
| dars | 3,160,340 | 64 | 52 | **35 (−32.7 %)** | 51 (−1.9 %) | diverges |
| NG5 | 7,402,886 | 256 | 86 | **57 (−33.7 %)** | 85 (−1.2 %) | diverges |

Symmetry alone buys 7 % on CORE2, decaying to 1 % on NG5. The right magnitude alone diverges on
every mesh, residual stalling ~4 orders above rtol. Both — var 1 — give −33…−39 %. (The
symmetrised header form, `−2·a_ij/(√(d_i d_j)(d_i+d_j))`, gives 79 / 155 / 35 / 58, i.e. var 1's
numbers; the two agree exactly where `d_i == d_j`.)

So our August patch was the wrong half: reasonable-looking on CORE2, close to worthless at NG5,
exactly where the solve costs most.

In-model, 300 steps, min-of-2, same allocation:

| config | nodes | ranks | s/step | **d_total** | SSH ms/step | d_SSH | SSH % of step |
|---|--:|--:|--:|--:|--:|--:|--:|
| CORE2 GPU | 4 | 16 | 0.0733 → 0.0665 | **−9.28 %** | 20.01 → 13.03 | −34.9 % | 27.3 → 19.6 |
| CORE2 GPU | 16 | 64 | 0.0890 → 0.0797 | **−10.45 %** | 27.14 → 17.69 | −34.8 % | 30.5 → 22.2 |
| NG5 CPU | 32 | 4 096 | 0.6257 → 0.6141 | −1.85 % | 11.26 → 7.37 | −34.6 % | 1.8 → 1.2 |
| NG5 CPU | 64 | 8 192 | 0.3067 → 0.3028 | −1.27 % | 11.04 → 7.27 | −34.2 % | 3.6 → 2.4 |
| NG5 CPU | 128 | 16 384 | 0.1615 → 0.1565 | −3.10 % | 14.37 → 9.23 | −35.8 % | 8.9 → 5.9 |

Iterations in-model match the replay and are the same on both backends and at every rank count:
CORE2 106.41 → 68.89, NG5 77.6 → 50.4 with the worst solve 99 → 67. `d_SSH` is likewise flat at
−34…−36 %; what differs between rows is only the SSH share it multiplies.

Accuracy — one-year CORE2 twin (512 ranks, monthly means), var 0 vs var 1. Signed bias: `sst`
+2.9e-06 K on an 11.2 K mean, `salt` −3.3e-07 on 21.4, `ssh` +6.0e-07 m. RMS differences are
2.5e-05 (`salt`) to 1.5e-02 (`v`) of each field's own spatial standard deviation. No systematic
drift after a full year.

One transient worth naming, from the shorter 20-step gate we run: `Kv` reaches 3.8e-01 at the
intermediate snapshot against our 2e-01 bound, settling to 1e-01 by step 20, where two other
solvers run as controls sit on the 1e-01 floor at both snapshots. `Kv` is a threshold-switch
field (the KPP mixing branch), so this is a switch flipping at a node rather than a continuous
error — but it is variant-1 specific, not generic to changing solver, and it is invisible in the
one-year means above.

Worth watching: `M⁻¹ = D⁻¹(2D − A)D⁻¹` is SPD only while `2D − A` is, i.e. under diagonal
dominance, which falls as `dt²`. Symmetry is unconditional, positive definiteness is not — #981's
`nnegrz` sensor covers it.

suvarchal:
question now is should we make variant 1 the default? now that HR meshed are tested?  or keep it for a while but not forget eventually to change it. any suggestions @koldunovn @JanStreffing @patrickscholz

koldunovn:
I would wait for a 60 year run that is now in flight.

JanStreffing:
This one comes next. Let me know when you have the 60 years through @koldunovn

koldunovn:
**TL;DR** — 60-year CORE2 hindcast twins in Fortran FESOM2, identical except
`precond_variant`. **Variant 1 reproduces the model's climate.** Every integrated
diagnostic ends within 0.3–1.2 % of the model's own 60-year drift, the sea-ice
trends agree to four significant figures, and there is no bias in any of them.


Setup: fesom27 `4ccdd35a` with this PR, one binary, CORE2, same forcing and same
initial state, 1958–2017, both arms cold-started. The two run directories differ
in that one namelist integer and nothing else. Statistics over 720 matched
monthly means.

| | variant 0 | variant 1 | \|Δ\| at 2017 | as % of v0's 60-yr drift |
|---|--:|--:|--:|--:|
| volume-mean T [°C] | 3.614046 | 3.614100 | 5.5e-05 | 0.25 % |
| ocean heat content [ZJ] | 19531.70 | 19532.00 | 0.29 | 0.25 % |
| volume-mean S [psu] | 34.727142 | 34.727142 | 3.3e-08 | ~0 |
| SST [°C] | 18.097369 | 18.096548 | 8.2e-04 | 0.83 % |
| SSS [psu] | 34.555080 | 34.553780 | 1.3e-03 | 1.2 % |
| NH sea-ice extent [10⁶ km²] | 13.4919 | 13.4923 | 4e-04 | — |
| SH sea-ice extent [10⁶ km²] | 12.3498 | 12.3499 | 1e-04 | — |

60-year sea-ice extent trend, NH **−0.3086 vs −0.3085**, SH **−0.0963 vs
−0.0963** 10⁶ km²/decade.

No bias: the signed mean differences are +2.8e-05 °C (volume-mean T), −2.1e-05 °C
(SST), −3.8e-09 psu (volume-mean S) — sign-indefinite and orders below the drift.

Pointwise, the two solutions separate the way any perturbed pair of a chaotic
model does, and then stop: `temp` rms 7.0e-03 °C (1958) → 6.8e-02 (1962) →
6.5e-02 (1980) → 8.4e-02 (2017). Saturated by 1962, no trend after. Panel (d)
shows where that lives — Gulf Stream, Kuroshio, equatorial Pacific, i.e. eddies —
area-weighted rms 6.7e-03 °C over the last decade. The deepest imprint on the
layer means is 1.1e-03 °C, at 55 m.

So the change is a different trajectory, not a different climate.
fig_precond_nodrift
fig_drift_precond

koldunovn (approved):

JanStreffing:
The T(z) profile is 1K colder at 50m depth. Is that ok? What is our climatological bias there? Is this in improvement or a degradation.

koldunovn:
fig_precond_meanstate
fig_precond_diff
More figures tu support the point.

koldunovn:
> The T(z) profile is 1K colder tat 100m depth. Is that ok? What is our climatological bias there? Is this in improvement or a degradation.

This is our model's bias - it just reproduce what we already have.

JanStreffing:
NVM its not new vs old. It's just that the new solver has the same cold bias as the old.

JanStreffing (approved):

JanStreffing:
I'm fixing the CTest Framework on a separate branch.

JanStreffing:
Given the obvious improvement, should this be opt-out, rather than opt-in?

sebastianbeyer:
I am in favor on making this the default or even removing the previous way if it has not benefits. Keep things simple and have best possible defaults :)

koldunovn:
Sergey thinks it should be default.

PR #985, opened 2026-08-18
Fix/avvelo cavity aware

ackerlar:
icb_dyn.F90   Make iceberg_avvelo cavity-aware consistent with iceberg_avvelo_andkeel; make interpol1D consistent with the 1.0e-6 guard; implementing "at_top" for cavity and non-cavity nodes in iceberg_avvelo instead of 'k==1'


JanStreffing:
The cavity-awareness change looks right in shape, but the `dz < 1` removal needs another look.

The `dz < 1` check this removes was guarding a divisor. `icb_dyn.F90` holds three local copies of `interpol1D` and they are not identical. The one inside `iceberg_avvelo` divides by `x1 - x0`, which is exactly `dz`, with no guard at all:

https://github.com/FESOM/fesom2/blob/0e00b4cd997b5b244428446d39de38b1a7d8439f/src/icb_dyn.F90#L1189-L1195

while the copies in `iceberg_levelwise_andkeel` and `iceberg_average_andkeel` both carry one:

https://github.com/FESOM/fesom2/blob/0e00b4cd997b5b244428446d39de38b1a7d8439f/src/icb_dyn.F90#L836-L841

The `lev_up == lev_low` exit only catches exact zero. The new cavity branch also makes small `dz` more reachable rather than less, since at `k == ulevels_nod2d` it pairs an interface depth from `zbar_3d_n` with the same layer's mid-level from `Z_3d_n_ib`, which under `use_cavity_partial_cell` can be centimetres apart. With the Intel release flags that is a SIGFPE, not a quiet NaN. Copying the `1e-6` guard into the third `interpol1D` looks like a cleaner fix than keeping the `dz < 1` abort.

The `abort=1` in `icb_elem.F90` has no effect. It sits in [`FEM_eval_old`](https://github.com/FESOM/fesom2/blob/0e00b4cd997b5b244428446d39de38b1a7d8439f/src/icb_elem.F90#L211), which has no callers anywhere in `src/` and is absent from the module's `public` list. The live [`FEM_eval`](https://github.com/FESOM/fesom2/blob/0e00b4cd997b5b244428446d39de38b1a7d8439f/src/icb_elem.F90#L149) has no equivalent error path, and [`matrix_inverse_2x2`](https://github.com/FESOM/fesom2/blob/0e00b4cd997b5b244428446d39de38b1a7d8439f/src/icb_elem.F90#L913) still ends in a bare `stop`.

Only two of the cavity guards were carried over from `iceberg_average_andkeel`, and the missing `abs(depth_ib) < abs(zbar_3d_n(k,n2))` clamp means a 150 m draft under a 300 m shelf base passes the keel test and extrapolates outside the bracket. The post-restart gate reconciliation also looks unreachable, because every rank reads the same restart path and `iceberg_out` writes a literal `0.0` into the `calving_day` field, so the [`MPI_MAX`](https://github.com/FESOM/fesom2/blob/0e00b4cd997b5b244428446d39de38b1a7d8439f/src/icb_step.F90#L104) is a no-op and the [WARNING it guards](https://github.com/FESOM/fesom2/blob/0e00b4cd997b5b244428446d39de38b1a7d8439f/src/icb_step.F90#L123) is dead code.

The MPI details check out. `-r8` is set for every compiler, so `MPI_DOUBLE_PRECISION` matches the arrays, and `par_ex(abort=1)` maps to `MPI_ABORT` in both the standalone and `__oifs` builds. The new blocking Allreduces are safe given `ib_async_mode` defaults to 0.

ackerlar:
@JanStreffing I updated the branch. Could you have another look?

JanStreffing:
Thanks, the `interpol1D` guard and the keel clamp both look right now. The PR description still says "dz is only ever a multiplicative weight, never a divisor", which the guard you added contradicts, so that paragraph wants updating before merge. Two findings from the earlier review are still open.

**The cavity fix corrects `lev_up` but not the velocity sampled there.** For a cavity node the loop skips to `k = ulevels_nod2d(n2)`, so all three `if( k==1 )` branches at [:1129](https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_dyn.F90#L1129), [:1156](https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_dyn.F90#L1156) and [:1171](https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_dyn.F90#L1171) are unreachable, and the trapezoid reads `UV_ib(:,k-1,n2)` at `k-1 = ulevels_nod2d-1`, which is the in-shelf level the bugfix comment says it avoids. `safe_lev` was introduced for the keel clamp but the integration path still indexes `k-1` directly.

The smallest fix is a top-of-column flag replacing the bare `k==1` test, since the first wet level of a cavity node needs the same piecewise-constant treatment the surface gets in the open ocean:

```fortran
logical :: at_top
...
at_top = (k == 1)
if (use_cavity .AND. mesh%cavity_depth(n2) /= 0.0) &
    at_top = (k == ulevels_nod2d(n2))
```

then `if (at_top)` at all three sites. That also makes `lev_up` and the sampled velocity consistent, since both would then refer to the shelf base rather than one to the shelf base and the other to a level above it.

**The post-restart reconciliation cannot correct `calving_day`.** `iceberg_out` writes a literal `0.0` into that field:

https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_step.F90#L1453

so every rank restarts with `calving_day = 0.0` and the `MPI_MAX` at [:104](https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_step.F90#L104) has nothing to reconcile, leaving the warning at [:123](https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_step.F90#L123) unreachable. That `0.0` is deliberate, a workaround for the uninitialised `istep` noted at [:1435](https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_step.F90#L1435), so it should stay as it is. The same argument covers `melted` and `find_iceberg_elem`, since all ranks read the same file and agree on input.

Either drop the `calving_day` reduction with its warning, or move the reconciliation to where per-rank divergence actually arises during the run. The `reject_tmp` cavity rejection at [:478-490](https://github.com/FESOM/fesom2/blob/1452f11273818fd9173bbea6c19570b7e75fbfeb/src/icb_step.F90#L478-L490) is the more likely source, and it is untouched here, so the timeouts turn a hang into a faster abort rather than removing its cause.

JanStreffing (approved):

PR #986, opened 2026-08-18
salt anomaly: runtime namelist toggle (rework of #977)

suvarchal:
## Summary

Runtime-namelist rework of #977 (salinity anomaly `S-35` for single-precision).
Builds on @koldunovn's 4 commits from #977 (authorship preserved) + one rework commit.
**Supersedes #977** (compile-time version). Targets `support_sp`.

---

# Salinity anomaly as a runtime option — rework of PR #977 and evaluation

## What this is
PR #977 (koldunovn, `sp-salt-anomaly`) stores the salinity state as the anomaly `S − 35`
(finer float32 spacing where the ocean lives) to attack single-precision salinity drift,
as a **compile-time** feature (`#ifdef USE_SALT_ANOMALY`). This branch reworks it as a
**runtime namelist toggle** on a single code path.

- Namelist: `use_salt_anomaly` (logical, default `.false.`) in `&oce_dyn` (`namelist.oce`).
- Derived: `S_ref_anomaly` is a module variable (`o_PARAM`), set to `35.0` when the toggle is
  on and `0.0` otherwise. Every consumer is unconditional (`salt + S_ref_anomaly`,
  `45 - S_ref_anomaly`, `a < -S_ref_anomaly`, `values - S_ref_anomaly`); with the toggle off
  `S_ref_anomaly = 0` and `x + 0.0 == x` bit-identically.
- All `#ifdef USE_SALT_ANOMALY` blocks removed; the CMake `USE_SALT_ANOMALY` option removed.
  Net −70 lines vs the PR (the duplicated `#ifdef/#else` pairs collapse to one path).
- Guard (`ocean_setup`) aborts, only when the toggle is on, for configs whose absolute-S
  consumers are not corrected: SPP, cavities, icebergs, age/transient tracers, KPP nonlocal
  fluxes, 3D clim relaxation, and the salinity diagnostics (`ldiag_extflds/trflx/salt3D/DVD/cmor`).

## Verification (pi/CORE2, GNU, `PATH=/usr/bin`)
- **Toggle OFF = bit-identical to `support_sp`** (SST `WORST_REL=0`; the file md5 differs only
  by the embedded `FESOM_git_SHA`). Full local CTest suite **7/7 green** with the toggle off.
- **Toggle ON = bit-identical to the pristine compile-time PR #977** (salt/sss/temp
  `WORST_REL=0` vs a `-DUSE_SALT_ANOMALY=ON` build). The rework faithfully reproduces the method.

## SP payoff (1-day pi/CORE2, salt error in psu)
| | mean | rms | max |
|---|---|---|---|
| SP precision error, absolute mode (SP_off vs DP_off) | 1.55e-5 | 6.35e-5 | 7.8e-3 |
| SP precision error, anomaly mode  (SP_on  vs DP_on)  | 0.96e-5 | 5.86e-5 | 7.1e-3 |
| Method bias in DP (DP_on vs DP_off), surface-only     | 4.2e-6 | 9.9e-5 | 1.1e-2 |

Anomaly storage cuts the **mean** SP salt error ~38% and the **interior is clean** (interior
bias 1.7e-6 vs surface 1.2e-4, a 70× split). But the surface method bias (below) is comparable
to the SP benefit on a 1-day run — so the net is roughly neutral at CI timescales. The drift
the method targets is a long-timescale effect; real validation needs a months–years run.

## The DP method bias — root cause (why does an anomaly change DP at all?)
Storing `S = S' + S_ref` should be invisible in DP: advection is linear and, by continuity,
`adv(S_ref) = −S_ref·dh/dt`, which exactly cancels the `−S_ref·dh/dt` from the anomaly
transform, leaving `d(h·S')/dt = −adv(S') − vdiff(S') + surface_flux` — the same equation with
`S'`. **So an exactly constancy-preserving advection would give zero DP change.**

FESOM's free-surface advection is **not** exactly constancy-preserving: with no correction the
DP-on vs DP-off salt error is **1.27e-3** (a constancy error for the constant offset 35,
concentrated at the surface where the freshwater volume flux enters). PR #977's
`S_ref·water_flux` surface term empirically cancels 99.4% of it, leaving a **8.2e-6/step**
residual (the part not proportional to `water_flux` — the grid-divergence / FCT-limiter part).

Empirical evidence:
- Residual is **exactly zero where the freshwater flux is zero** → 100% freshwater-driven.
- `surf_relax_S = 0` leaves the residual unchanged → `relax_salt` is consistent, not the source.
- `real_salt_flux` is built from the corrected absolute ice gather (`S_oc_array = salt+S_ref`),
  so it is consistent and needs no anomaly term of its own.
- Correction-term matrix (1-step salt `WORST_REL`, DP on-vs-off):
  - A — surface `S_ref·water_flux` only (= PR #977): **8.2e-6** (best)
  - B — no correction: 1.27e-3
  - C — per-layer thickness term `S_ref·(hnode−hnode_new)` only: 1.26e-3 (no help)
  - D — surface + thickness: 2.85e-4 (worse than A)

So the PR's surface term is the best simple correction; naive per-layer/thickness fixes fail
because they don't address the constancy error at the **advective** flux.

## The exact fix (not yet done)
Make the surface advective flux carry `S' + S_ref` (the scheme sees the true absolute value at
the surface face), so the constancy error never arises. That means editing the surface term in
**every** advection scheme (FCT low+high order in `oce_adv_tra_driver.F90`, plus
`adv_tra_ver_upw1`/QR4C in `oce_adv_tra_ver.F90`) for `tr_num` with `ID==2` only — higher risk,
and it must not reintroduce the constancy imbalance (experiments C/D show naive attempts do).
Until then the 8.2e-6/step residual is an inherent, small (0.6% of the freshwater-background
term) limitation of representing salinity as an anomaly in this scheme.

## Recommendation
1. The runtime rework is solid and no-op-safe — ready to keep.
2. The method's interior SP benefit is real; the surface bias is small but inherent without the
   exact advective-flux fix.
3. Decide adoption on a **long (months–years) SP-anomaly vs SP-absolute vs DP** run (Levante),
   not the 1-day pi test.

## Key files
`src/oce_modules.F90` (toggle + `S_ref_anomaly`), `src/oce_setup_step.F90` (set 35, subtract,
guard), `src/fesom_module.F90` (restart auto-convert), `src/ice_oce_coupling.F90`,
`src/oce_ale_mixing_kpp.F90`, `src/oce_ale_pressure_bv.F90` (EOS incl. `density_linear`),
`src/oce_ale_tracer.F90` (clip, surface BC), `src/io_meandata.F90` (output offset),
`src/write_step_info.F90` (blowup bounds), `config/namelist.oce`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ Makefile.in
 mesh_part/build
 test/output_pi
 .claude/
+build_*/
--- a/config/namelist.oce
+++ b/config/namelist.oce
@@ -19,6 +19,9 @@ C_d                = 0.0025  ! bottom drag coefficient (dimensionless, typical: 
 A_ver              = 1.e-4   ! background vertical viscosity [m²/s]
 scale_area         = 5.8e9   ! reference element area for viscosity/diffusivity scaling [m²]
 
+! --- Salinity state representation ---
+use_salt_anomaly   = .false. ! store salinity as the anomaly S-35 (finer float32 spacing; helps single precision). .false. = absolute salinity, bit-identical to before
+
 ! --- Salt Plume Parameterization ---
 SPP                = .false. ! enable Salt Plume Parameterization (for brine rejection under sea ice)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -81,8 +81,6 @@ message(STATUS "USE_ICEPACK: ${USE_ICEPACK}")
 
 option(USE_SINGLE_PRECISION "Build FESOM in single precision (WP=4)" OFF)
 message(STATUS "USE_SINGLE_PRECISION: ${USE_SINGLE_PRECISION}")
-option(USE_SALT_ANOMALY "Store the salinity state as the anomaly S - S_ref (finer float32 spacing; see o_PARAM::S_ref_anomaly)" OFF)
-message(STATUS "USE_SALT_ANOMALY: ${USE_SALT_ANOMALY}")
 
 
 # option to trigger building a library version of FESOM
@@ -299,10 +297,6 @@ endif()
 
 if(USE_SINGLE_PRECISION)
    target_compile_definitions(${PROJECT_NAME} PRIVATE USE_SINGLE_PRECISION)
-endif()
-
-if(USE_SALT_ANOMALY)
-   target_compile_definitions(${PROJECT_NAME} PRIVATE USE_SALT_ANOMALY)
 endif()
 
 if(CVMIX)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -81,6 +81,8 @@ message(STATUS "USE_ICEPACK: ${USE_ICEPACK}")
 
 option(USE_SINGLE_PRECISION "Build FESOM in single precision (WP=4)" OFF)
 message(STATUS "USE_SINGLE_PRECISION: ${USE_SINGLE_PRECISION}")
+option(USE_SALT_ANOMALY "Store the salinity state as the anomaly S - S_ref (finer float32 spacing; see o_PARAM::S_ref_anomaly)" OFF)
+message(STATUS "USE_SALT_ANOMALY: ${USE_SALT_ANOMALY}")
 
 
 # option to trigger building a library version of FESOM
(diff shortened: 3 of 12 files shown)


sebastianbeyer:
Would only SP benefit from this or also DP? Why not make it the default (or only) option then? are there disadvantages? I'd prefer to not give people too many options to keep thing simpler...

suvarchal:
It doesn't flow yet into DP, it changes conservation, that needs deeper investigation. But the direction is a good one to support mixed precisions.

PR #988, opened 2026-08-18
Modini method; Partial wind coupling

JanStreffing:


JanStreffing:
Had a chat with @qiangclimate, and he will make a similar implementation with anomaly method.

PR #989, opened 2026-08-18
modini: partial wind coupling (Thoma et al., 2015)

JanStreffing:
Cherry-pick of 31529f72 ("Modini technique is on progress") from the closed #988, replayed onto current `main`.

Opened as a **draft**: the code builds and is off by default, but it has not been run, and there are open questions listed at the bottom that the original author should settle.

## What it does

An opt-in `use_modini` path. A second pair of wind fields is read from the standalone forcing files (`nm_xmodini_file` / `nm_ymodini_file`), and in a coupled run the wind stresses received from the atmosphere are discarded and recomputed from those prescribed winds, so the ocean and ice see a wind field carrying no coupled feedback.

`use_modini = .false.` by default, so every existing configuration is unchanged.

| file | change |
|---|---|
| `gen_modules_config.F90` | `use_modini`, `check_modini`, `check_fesom` in `&run_config` |
| `gen_surface_forcing.F90` | `i_xmodini`/`i_ymodini` fields, namelist entries, rotation, runoff/SSS skip |
| `gen_forcing_couple.F90` | recompute ocean/ice stress from modini winds (OASIS path) |
| `gen_bulk_formulae.F90` | use modini winds in `nemo_ocean_fluxes_mode` |
| `gen_forcing_init.F90` | call `sbc_ini` when modini is on |
| `namelist.config`, `namelist.forcing` | the new switches and file/variable names |

## Deviations from 31529f72

The original was written in Aug 2025 against a much older tree and did not apply cleanly — 22 conflict hunks across 7 files. It is a replay, not a clean pick. What differs, and why:

**Dropped as unrelated or accidental**
1. **Whitespace-only reindentation** (tab→space) in `cpl_driver.F90` and `gen_forcing_couple.F90` — roughly a third of the original diff, and the source of most conflicts. `cpl_driver.F90` contained nothing else and is untouched here.
2. **`do_rotate_ice_wind` → `do_rotate_oce_wind`** at `cpl_recv` index 3 (`taux_ico`), plus two comments changed from `tauy_oce`/`taux_ice` to `taux_oce`. This looks like reformatting damage rather than intent, so it is not carried over.
3. **Debug scaffolding** — the commented-out `write(*,*)` blocks throughout `nc_readTimeGrid`, and one *uncommented* leftover, `write(*,*) 'time should be it ', id_time`, which would print on every forcing file open.

**Changed because the original would not work**

4. **Flux-count gating is `use_modini`, not `use_ice`.** The original allocated the two modini slots under `if (use_ice)`, but `nc_sbc_ini_fillnames` only fills their file names under `if (use_modini)`. Any standalone run with sea ice would therefore reserve two forcing fields with empty file names and fail on the first read — i.e. it would break the default configuration, not just the modini one.
5. **Rotation pairs fixed.** The original replaced `vector_g2r(coef_b(i_xwind), coef_b(i_ywind))` with `vector_g2r(coef_b(i_xmodini), coef_b(i_ymodini))`, leaving `coef_a` rotated for the wind pair and `coef_b` for the modini pair — so neither pair was fully rotated. Replaced with a separate `do_rotation_modini` block rotating both `coef_a` and `coef_b` of the modini pair, mirroring the existing wind and stress blocks.
6. **The stress block moved out of the receive loop.** In the original it sat inside `do i=1,nrecv`, so `sbc_do` and the full stress recomputation ran once per received field (13–20× per coupling step). It now runs once, *after* the loop — and after the rotation block, since `sbc_do` already rotates the modini coefficients and rotating the derived stresses again would double-rotate them.

**Ported to current `main`**

7. `partit` arguments, the `nf90_*` netCDF interface, and per-file calendar handling.
8. `main` added `i_xstre`/`i_ystre` since, so the modini indices are **13/14** and `i_totfl` is 14, not 11/12 and 12.
9. `config/namelist.config.dkrz` was deleted from `main` in 7afd2359, so the two switches went into `config/namelist.config` `&run_config` instead.
10. The runoff/SSS skip is expressed as conditions rather than the original's re-indented wrapper, because `main` rewrote `sbc_ini`/`sbc_do` with file-existence validation and hard exits. The chlorophyll block, which is new in `main`, is left enabled under modini.

## Testing

- Builds clean on Levante with Intel: `./configure.sh levante.dkrz.de`, exit 0, no errors, `fesom.x` produced.
- **The OASIS path is not covered by that build.** The stress block in `gen_forcing_couple.F90` sits inside `#if defined (__oasis)`, so the default build never compiles it. The `coupled` preset in `fesom2_build_tests.yml` will.
- **The feature has not been run.** No modini simulation was performed, and there is no test coverage.

## Open questions for the author

- `check_modini` and `check_fesom` are declared and exposed in `&run_config` but never read anywhere. Keep as placeholders, or drop until something uses them?
- In an **uncoupled** build with `use_modini = .true.`, `sbc_ini` is now called twice — once from the existing `#if !defined(__oasis) && !defined(__yac)` branch and once from the new block. Should the modini call be restricted to coupled builds?
- Is skipping SSS restoring and runoff the intended behaviour under modini in all cases, or only when the coupler supplies them?


JanStreffing:
@ufukozkan, this is my attempt to make modini a technique a permanent optional feature in FESOM2. Are you interested in testing this on a modern FESOM2.7.7 codebase, or shall I do some tests?

ufukozkan:
Yes, I would do that!

PR #990, opened 2026-08-18
fct: pad AUX above cavity elements to fix first-timestep NaNs

### LINKED ISSUE #906: NaNs in tracer fields in the first timestep of simulations using meshes with cavities
This problem has been reported every now and again, I remember it first by @JanStreffing towards the end of last year. Especially in the CORE3 development context while trying out different configurations and distributions it occurred frequently. In the very first timestep there might be NaNs occurring in temperature and/or salinity (probably also other tracers) resulting in the model to blow up. It only happens in meshes with cavities, and NaNs occur only in mesh partitions including a fraction of a cavity. It may depend on the actual partition used - running with a distribution for one node might work perfectly while using several nodes and thus partitioning for that size may fail.

Since this affects the test simulations in the CORE3 mesh development I tried to track it down in FESOM2.7.1 and in the meantime I have a suspicion it might be related to tracer advection routines. One of the first NaN occurrences I could find, happens in the routine  `oce_tra_adv_fct` in the file `oce_adv_tra_fct.F90`.

In that routine a large auxiliary field `AUX` is used and seemingly filled and evaluated inconsistently in some cases.

It is filled in a loop over elements 

```
    do elem=1, myDim_elem2D
        enodes=elem2D_nodes(:,elem)
        nu1 = ulevels(elem)
        nl1 = nlevels(elem)

        !$ACC LOOP VECTOR
        do nz=nu1, nl1-1
            AUX(1,nz,elem)=max(fct_ttf_max(nz,enodes(1)), fct_ttf_max(nz,enodes(2)), fct_ttf_max(nz,enodes(3)))
            AUX(2,nz,elem)=min(fct_ttf_min(nz,enodes(1)), fct_ttf_min(nz,enodes(2)), fct_ttf_min(nz,enodes(3)))
        end do
       . . .
    end do
```
 and later on evaluated in a loop over nodes

```
    do n=1, myDim_nod2D
       nu1 = ulevels_nod2D(n)
       nl1 = nlevels_nod2D(n)
       !___________________________________________________________________
       !$ACC LOOP VECTOR
       do nz=nu1,nl1-1
          ...
          do elem=2,nod_in_elem2D_num(n)
              tvert_max(nz, n) = dmax1(tvert_max(nz, n), AUX(1,nz, nod_in_elem2D(elem,n)))
              tvert_min(nz, n) = dmin1(tvert_min(nz, n), AUX(2,nz, nod_in_elem2D(elem,n)))
         end do
      end do

    ....
```
and in the case with NaN trouble that I investigate, the level range covered by `ulevels(elem)` and `ulevels_nod2D(n)` is quite different, so in this particular case `AUX` in the problematic mesh node is set in the level range 19 to 25 but evaluated in the range 7 to 25, so here the result in the nonoverlapping part depends on prior settings in `AUX` which may randomly put NaNs into `tvert_max` or `tvert_min` and from here the trouble spreads in my case. This could explain the dependency on the partition and the restriction of the problem to cavity regions where the level range varies.

My workaround for the moment is to fill the proper range by replacing the `ulevels(elem)` by the max level range over the nodes in the `elem`:

```
    do elem=1, myDim_elem2D
        enodes=elem2D_nodes(:,elem)
        !sh nu1 = ulevels(elem)
        !sh nl1 = nlevels(elem)
        ! ====== WORKAROUND ======
        nu1=minval(ulevels_nod2D(enodes(:)))
        nl1=maxval(nlevels_nod2D(enodes(:)))
        !=========================
       ...

    end do
```
which avoids the random entries but of course I cannot be sure if the resulting values in `AUX` are meaningful. In any case with the workaround this problem does no longer occur, the test simulations run so far or blow up for other reasons ;) Maybe I was looking here at some weird exceptional case but please take a look if this discrepancy might happen in other setups with cavities as well. My ongoing experiments are located on albedo in the folder
```
/albedo/work/user/sharig/FESOM271_CORE3_test/run.CORE3.cav.jig.dev/
```

JanStreffing:
Implements @patrickscholz's suggestion in #906.

In `oce_tra_adv_fct`, `AUX` is filled per element over `nz=ulevels(elem)..nlevels(elem)-1` and padded with `∓bignumber` below the bottom. The levels *above* a cavity element (`nz=1..ulevels(elem)-1`) were never written.

The following node loop reads `AUX` over `nz=ulevels_nod2D(n)..nlevels_nod2D(n)-1` for every element around node `n`. Where the node is shallower than a neighbouring cavity element, it reads that unwritten range, so `tvert_max`/`tvert_min` pick up stale memory (`AUX` is aliased onto `twork%edge_up_dn_grad`) — hence NaNs in the first timestep, only on cavity meshes and only for partitions containing a cavity.

Fix mirrors the existing sub-bottom padding, so those levels are neutral in the max/min search and the rest of the routine is untouched:

```fortran
if (nu1>1) then
    do nz=1,nu1-1
        AUX(1,nz,elem)=-bignumber
        AUX(2,nz,elem)= bignumber
    end do
endif
```

Alternative to `fix/cavity_surf_nan` (@PengyangSong's 7ed3c49), which fixes the same bug on the read side with a per-element branch in the inner loop. This one keeps the inner loop branch-free.

No new variables, so the OpenMP `private` list and the OpenACC clauses are unchanged.

Not yet tested on a cavity mesh — @svenharig, does this hold up in the CORE3 runs where you hit it?

Fixes #906


patrickscholz (approved):

svenharig:
Just for the sake of completeness, I tested the updated code with a CORE3 mesh with cavities for a number of partitions and everything works well as expected.

PR #993, opened 2026-08-18
gen_support: accumulate global integrals in WP_full

suvarchal:
**Stacked on #992**, which is stacked on #984. Review those first — this PR's own diff is 24 lines in one file. Retarget up the stack as each lands.

`integrate_nod_2D/3D` and `integrate_elem_2D/3D` each build a global integral as an uncompensated running sum over every node or element a rank owns — ~1e5 terms on CORE2, ~7e6 on NG5 — in a `WP` scalar, then `MPI_SUM` across ranks.

### Why these are not just diagnostics

Their results are used to **balance global fluxes to zero**. Every one of these does `net = integrate_nod(flux) / ocean_area` and then subtracts `net` from a surface field so the domain integral vanishes:

| correction | site |
|---|---|
| freshwater | `ice_oce_coupling.F90:620,632,636,640,643` |
| virtual salt | `:445,451` |
| virtual salt, cavity | `:484,488` |
| SSS restoring | `:516,517` |
| icebergs | `:599` |
| water isotopes | `:673,677,689` |

So an error in that integral is not noise that averages out. It is a systematic net source or sink applied uniformly to every surface node, every step, with a sign set by whatever the seasonal cycle of the flux happens to be doing.

### The change

Accumulate in `WP_full` (from #992) instead of `WP`, so the integral is independent of `WP`. The `MPI_SUM` already used `MPI_DOUBLE_PRECISION`, so this also makes the buffer type and the declared type agree **by construction** rather than by coincidence — today they agree only because `WP` happens to be `real64`.

### Verification

**No effect while `WP` is `real64`** — `WP_full == WP`, so this is bit-identical:

```
nc_diff.py <pr3 base> <this branch>   ->  WORST_REL=0.000e+00  NAN=0  (BITWISE-IDENTICAL)
```

(`nc_diff.py` rather than md5: FESOM stamps the git SHA into a NetCDF global attribute, so numerically identical runs from different commits have different md5s.) Full local CTest suite 7/7, clean build.

The point is that it *stays* correct if `WP` is ever narrowed — which is exactly when an uncompensated 1e5-term float32 sum, relative error ~`eps*sqrt(N)` ~ 4e-5, would start injecting a systematic imbalance into every one of the corrections above.

### Not covered here

`mesh%ocean_area` — the denominator those corrections divide by — is summed the same uncompensated way in `oce_mesh.F90` (~line 2393), but through a local `vol` that is shared with the mesh-resolution computation. Fixing it properly means deciding whether `mesh%ocean_area` itself becomes `WP_full` and rippling that to its consumers. Separate change.


suvarchal (commented):
checked!

suvarchal:
**Note for whoever merges this with single-precision support** — found while assembling an integration branch of every SP fix on top of `support_sp`.

This PR accumulates in `WP_full` and leaves the `MPI_SUM` as `MPI_DOUBLE_PRECISION`, which is correct on `main`. But `support_sp` reduces these same calls with **`MPI_WP`**, and `MPI_WP` is `MPI_REAL` in a single-precision build. Merging the two without touching the MPI type gives an 8-byte buffer reduced as 4 bytes:

```
lval : real(kind=WP_full)                     ! 8 bytes
call MPI_AllREDUCE(lval, ..., MPI_WP, ...)    ! MPI reads 4
```

The global integrals come back as garbage. In practice that NaN'd the SSH solve on its first iteration (`rtol=NaN`, `p.Ap=NaN`) and blew up **every** SP integration test — while DP stayed 10/10, because there `WP_full == WP` and the mismatch cannot show.

Nothing to change in this PR as it stands: on `main` today the buffer and the MPI type agree. The hazard is that they agree *by coincidence* — `WP` happens to be `real64` — rather than by construction, so it goes unnoticed until SP arrives.

Worth considering a named `MPI_WP_FULL` constant beside `MPI_WP` (as the original exploration branch had in `MOD_PARTIT`), so the pairing is structural instead of something each merge has to remember. The same applies to #994.

suvarchal:
**Correction to my note above — the problem is bigger than the MPI datatype, and it is in this PR's own code.**

I said the only hazard was `MPI_WP` vs `MPI_DOUBLE_PRECISION` at merge time. That was incomplete. The reduction also writes into the caller's result argument:

```fortran
real(kind=WP), intent(inout) :: int2D          ! real32 when WP is single
real(kind=WP_full)           :: lval           ! real64
call MPI_AllREDUCE(lval, int2D, 1, MPI_DOUBLE_PRECISION, MPI_SUM, ...)
```

`int2D` is `real(kind=WP)`. With an 8-byte MPI datatype, **MPI writes eight bytes into a four-byte variable** — a buffer overflow, and the returned integral is garbage. Correcting the datatype alone does not help; it is what *causes* the overflow once the accumulator is `WP_full`.

Measured on an integration branch of `support_sp` + this PR, single precision: the corrupted integrals reached `water_flux` through the freshwater and virtual-salt balancing, drove `eta_n` out of bounds (`--STOP--> found eta_n become NaN or <-10.0, >10.0`), and blew up **every** SP integration test. DP stayed 10/10 throughout, because there `WP_full == WP` and neither the datatype nor the width can disagree.

Fix is small and keeps the public interface unchanged — reduce into a local `WP_full` buffer and round once on the way out:

```fortran
real(kind=WP_full) :: lval, gval
...
call MPI_AllREDUCE(lval, gval, 1, MPI_DOUBLE_PRECISION, MPI_SUM, ...)
int2D = real(gval, WP)
```

With that, the same branch goes to **10/10 in SP** and stays 10/10 in DP.

This is invisible on `main` today, so the PR is not broken as merged there — but it is a latent memory-corruption bug that fires the moment `WP` narrows, which is precisely what this PR is preparing for. Worth fixing here rather than leaving for the SP merge to discover. Same shape as `mesh%ocean_area`, which needs the identical treatment.

JanStreffing (approved):

PR #994, opened 2026-08-18
solver: accumulate the ssh CG in WP_full

suvarchal:
**Stacked on #993 → #992 → #984.** Review down the stack first; this PR's own diff is one file.

The CG's vectors and matrix stay `WP` — they own the memory traffic. Its *accumulations* move to `WP_full`, because they steer the iteration rather than merely report on it:

- the **four global reductions** — `rhs.rhs`, `r.z`, `p.Ap`, and the `r.z`/`r.r` pair per iteration — which set `beta`, `alpha`, the stopping tolerance `rtol` and the exit test
- the **three sparse row sums** — `r0 = b - Ax`, `z = M⁻¹r`, `A.p` — each summed in `WP_full` and rounded once on the way back into its `WP` vector

### Both preprocessor branches

The plain-OpenMP path promotes each *term* before adding it to an already-`WP_full` accumulator; the `__openmp_reproducible` path promotes the operand arrays before `sum()`. Promoting the terms rather than only the accumulator is the part that matters — `a*b` between two `WP` operands is evaluated in `WP` and only then widened, so widening the accumulator alone would buy nothing.

### Type correctness

The `MPI_Allreduce` calls already passed `MPI_DOUBLE` while the buffers were declared `real(WP)` (`solver.F90:179, 217, 248, 304`). That is correct only because `WP` is currently `real64`. It is now correct by construction.

### Verification

No effect while `WP` is `real64`:

```
nc_diff.py <pr3 base> <this branch>   ->  WORST_REL=0.000e+00  NAN=0  (BITWISE-IDENTICAL)
CG iterations, 96 solves:  base 12.2708 mean / 16 max   ==   this branch 12.2708 / 16
full local suite: 7/7
```


suvarchal:
**Note for whoever merges this with single-precision support** — found while assembling an integration branch of every SP fix on top of `support_sp`.

This PR accumulates in `WP_full` and leaves the `MPI_SUM` as `MPI_DOUBLE_PRECISION`, which is correct on `main`. But `support_sp` reduces these same calls with **`MPI_WP`**, and `MPI_WP` is `MPI_REAL` in a single-precision build. Merging the two without touching the MPI type gives an 8-byte buffer reduced as 4 bytes:

```
lval : real(kind=WP_full)                     ! 8 bytes
call MPI_AllREDUCE(lval, ..., MPI_WP, ...)    ! MPI reads 4
```

The global integrals come back as garbage. In practice that NaN'd the SSH solve on its first iteration (`rtol=NaN`, `p.Ap=NaN`) and blew up **every** SP integration test — while DP stayed 10/10, because there `WP_full == WP` and the mismatch cannot show.

Nothing to change in this PR as it stands: on `main` today the buffer and the MPI type agree. The hazard is that they agree *by coincidence* — `WP` happens to be `real64` — rather than by construction, so it goes unnoticed until SP arrives.

Worth considering a named `MPI_WP_FULL` constant beside `MPI_WP` (as the original exploration branch had in `MOD_PARTIT`), so the pairing is structural instead of something each merge has to remember. The same applies to #993.

JanStreffing (approved):

JanStreffing:
Dependency stack has been merged. These make sense to me.

PR #1004, opened 2026-08-19
DARS mesh namelists (supersedes #824)

JanStreffing:
Adds the four DARS mesh namelists originally contributed by @TunaHim in #824, moved onto a branch in this repository so they can be maintained here, with a few corrections on top.

The original commits are preserved, so authorship stays with @TunaHim. A separate PR was needed because GitHub only allows editing a pull request's base branch and not its head, so #824 could not be repointed away from the fork.

Changes on top of the original:

- `visc_gamma2` was `0.0285` in `namelist.dyn_DARS`, where both the CMIP7 HR configuration and the repo default use `0.285`. This is inert while `opt_visc = 7`, since gamma2 only applies to easy backscatter, but wrong if the scheme is ever switched.
- `AB_order` was unset in both files and so fell back to 2. Now pinned to 3 for dynamics and 2 for tracers, matching CMIP7.
- `idemix_surforc_file` pointed at `/work/ollie`, which is decommissioned, and the two bottom forcing entries pointed at an albedo project directory. All three now use `/pool/data/AWICM/FESOM2/FORCING/IDEMIX`, where both datasets exist under the same filenames.
- Added `idemix_surforc_vname` and `idemix_botforc_vname`, which were missing, using the values CMIP7 uses.

Checked against the CMIP7 HR configuration running the 1pctCO2 experiment. `&param_tke` matches it exactly, which is the block that counts since CMIP7 runs `mix_scheme = 'cvmix_TKE'`, and `namelist.tra_DARS` lines up apart from the `AB_order` noted above.

Two things left for whoever knows this mesh. `idemix_jstar = 10.0` and `idemix_mu0 = 1.33333333` differ from both CMIP7 and the repo default (`5.0` and `0.33333333`, the latter by a factor of four), and I left them alone because they read as a deliberate tuning choice rather than a slip. There is also no `namelist.oce_DARS`, so nothing selects the mixing scheme and the IDEMIX and tidal blocks may never be read at all.

Supersedes #824, which can be closed once this is merged.


JanStreffing:
@patrickscholz, @sebastianbeyer: quick one: `idemix_jstar = 10.0` and `idemix_mu0 = 1.33333333` here differ from both CMIP7 and the repo default (`5.0` and `0.33333333`). Deliberate for this mesh, or should they follow the default?

JanStreffing:
Restored `namelist.oce_DARS` in 2d8ca033. It was deleted in 82f8cb76 on #824 in response to the request to keep only namelists that differ from the defaults, but it does differ. Six of its twenty settings are not the default:

| key | DARS | default |
|---|---|---|
| `Fer_GM` | `.false.` | `.true.` |
| `Redi` | `.false.` | `.true.` |
| `mix_scheme` | `'cvmix_TKE'` | `'KPP'` |
| `K_GM_max` | `2000.0` | `1000.0` |
| `K_GM_bvref` | `2` | `1` |
| `K_GM_resscalorder` | `1` | `2` |

`mix_scheme` is the one that matters most. Without it the configuration falls back to KPP, which leaves the fully specified `&param_tke` block in `namelist.cvmix_DARS` unused, so the two files only make sense together. GM and Redi being switched off also looks like a deliberate choice for a mesh at this resolution rather than something inherited by accident.

The default has been `KPP` for the whole history of `config/namelist.oce`, so this was a difference at the time of the original review too. 

On second look, I am worried about the GM values. While not used here, if anyone would turn GM on, those are values from before the 2.7 fixes, right @patrickscholz ?

JanStreffing:
I will update the descriptions with @FinnHeu 's and keep this content. I will also unify the rest of the namelists here.

patrickscholz (approved):

PR #1006, opened 2026-08-19
recom: stop double counting Benthos under tracer parallelisation

### LINKED ISSUE #1003: Benthic Si and calcite are halved when num_fesom_groups > 1

JanStreffing:
Fixes #1003.

**What.** `ver_sinking_recom_benthos` adds each sinking contribution to `Benthos` directly and also buffers the same value in `Benthos_tr`. Under `__usetp`, `oce_ale_tracer.F90` then rebuilds `Benthos` by summing `Benthos_tr` over all tracers, so every tracer a group owns lands in `Benthos` twice. The fix subtracts this group's own slice before the summation, so each tracer contributes exactly once.

**Why it matters.** This is not a multi-group problem. With one group it is a straight factor of two on all four benthic pools, and `RECOM_COUPLED=ON` sets `__usetp` unconditionally, so it affects every REcoM run. With more groups the error becomes `(1 + f) / 2` of the correct value, where `f` is the share of a pool's flux coming from that group's own tracers, which is why some pools looked exactly halved and others did not.

**Result.** Pi mesh, 1 day, 32 steps/day, base REcoM configuration. Domain mean of each pool on 2 ranks, before and after:

| pool | before | after | ratio |
|---|---|---|---|
| `BenN` | 1.259452e+08 | 6.297261e+07 | 0.500000 |
| `BenC` | 6.297274e+08 | 3.148637e+08 | 0.500000 |
| `BenSi` | 3.160345e+08 | 1.580172e+08 | 0.500000 |
| `BenCalc` | 9.811391e+01 | 4.905695e+01 | 0.500000 |

And 1 group against 2 groups, max relative deviation over all nodes:

| pool | before | after |
|---|---|---|
| `BenN` | 2.53e-01 | 1.66e-09 |
| `BenC` | 2.54e-01 | 6.07e-09 |
| `BenSi` | 5.00e-01 | 7.62e-09 |
| `BenCalc` | 5.01e-01 | 9.13e-04 |

The `BenCalc` outlier is one node where the pool is around 1e-4 in absolute terms, so it is cancellation rather than a real disagreement.

**Note.** Benthic pools feed the bottom fluxes for silicate and for DIC and alkalinity, so `test_pi_recom` truth values will move and need updating with this.

### DIFF
--- a/src/oce_ale_tracer.F90
+++ b/src/oce_ale_tracer.F90
@@ -171,6 +171,7 @@ subroutine solve_tracers_ale(ice, dynamics, tracers, partit, mesh)
 ! multi FESOM group loop parallelization
     integer             :: num_tracers
     integer             :: tr_num_start_memo
+    integer             :: tr_num_end_memo
 
     integer             :: group_i
     integer             :: tr_num_start
@@ -270,6 +271,7 @@ subroutine solve_tracers_ale(ice, dynamics, tracers, partit, mesh)
     Benthos_tr_slice_count_fix_1 = 1 * (myDim_nod2D + eDim_nod2D) * benthos_num
 
     tr_num_start_memo = tr_num_start
+    tr_num_end_memo   = tr_num_end
 
     request_count = 0
 #endif
@@ -460,6 +462,16 @@ subroutine solve_tracers_ale(ice, dynamics, tracers, partit, mesh)
 #if defined(__recom) && defined(__usetp)
 ! SinkFlx and Benthos values are buffered per tracer index in the loop above and now summed up to
 ! avoid non bit identical results regarding global sums when running the tracer loop in parallel
+! ver_sinking_recom_benthos adds each contribution to Benthos directly AND
+! buffers the same value in Benthos_tr. Summing the buffers below would
+! therefore count this group's own tracers twice. Remove that slice again
+! first, so Benthos ends up with every tracer exactly once. Where the
+! buffers are not filled (use_MEDUSA with sedflx_num /= 0) they are zero
+! and both loops are a no-op, leaving the previous behaviour untouched.
+        do tr_num = tr_num_start_memo, tr_num_end_memo
+            Benthos = Benthos - Benthos_tr(:, :, tr_num)
+        end do
+
         do tr_num = 1, num_tracers
             if(use_MEDUSA) then
                 SinkFlx = SinkFlx + SinkFlx_tr(:, :, tr_num)


JanStreffing:
@a270105 Please check. I do not know much about this piece of code yet. I hope this fixes to cause, not just symptoms.

a270105:
You are right.
I checked in my test runs with different numbers of tracer groups. Restart variables for Benthos are all identical. Those runs were done with the code in my fesom branch before merging with the recom submodule. In the submodule code (resom_sinking.F90) there is no conditional check for calculation of Sinkflx_tr and Benthos_tr. In my version before the merge, this is coded this way:
Screenshot 2026-08-20 at 00 24 15

So either Benthos or Benthos_tr can be calculated. And the line for Benthos without any conditional check is commented out.
I will update the submodule code and run some tests tomorrow.

a270105:
Updates in recom_sinking.F90 does not solve the problem.
New tests with code from the fesom branch `fesom2.6_recom_tp` confirm identical output for Benthos with and without tp (also identical to the output with this branch when num_fesom_groups = 1).
Further comparing these two branches...

a270105:
I think I found the cause. Please confirm the fix.

Since recom is now a submodule and has its only cmake configuration, without including __usetp there 
Benthos_tr buffering in `recom_sinking.F90` is compiled out, even though the corresponding parts in oce_ale_tracer.F90 is compiled. Sinking fluxes only write directly into Benthos, not through Benthos_tr. This is only correct when num_fesom_groups = 1. With >1 groups, the cross-group merge into Benthos_tr is never filled.

I added `add_compile_definitions(__recom **__usetp**)` in CMakeLists.txt in recom and the tests give identical output now.

JanStreffing:
Your fix is right, and `0d729bb` on `yings_branch` does it properly: the `Benthos_tr` buffering and the direct `Benthos` write are now mutually exclusive, with ten guards covering all eight pools.

https://github.com/RECOM-Regulated-Ecosystem-Model/REcoM/blob/0d729bbf4e008979d30649174babaf4c3d78085e/recom_sinking.F90#L115-L130

The submodule pin is the part I would not leave as it is:

| ref | `__usetp` | Benthos writes |
|---|---|---|
| `master` `526df4c7` | absent | both, unconditional |
| `e7286ed9`, pinned by fesom `main` | absent | both, unconditional |
| `yings_branch` `0d729bb` | present and defined | mutually exclusive |

`e7286ed9` sits on `yings_branch` rather than `master`, and no commit in `master`'s history of `recom_sinking.F90` has ever carried the guards. The submodule pins a feature branch today, and a later bump onto a `master` commit reintroduces the double count. Merging `yings_branch` into REcoM `master` matters more than anything on the fesom side.

I will close this PR in favour of bumping the submodule to your commit, since fixing the redundant write beats compensating for it downstream. The two cannot both land: once only `Benthos_tr` is written, subtracting the group's own slice as this branch does would drop those tracers entirely and under-count.

`test_pi_recom` truth values move either way and still need regenerating.

PR #1025, opened 2026-08-27
Density-MOC diagnostic fixes and neverworld2 toy configuration

patrickscholz:
## Summary

This PR contains two groups of changes: fixes to the density-MOC (`ldiag_dMOC`) diagnostics that are relevant for **all** configurations, and the accumulated development of the `neverworld2` idealized channel configuration used to find them.

### Density-MOC diagnostic fixes (relevant beyond the toy setup)

- **Include boundary edges in the density-class divergence assembly** (`gen_modules_diag.F90`). The edge loop building `std_dens_DIV`/`std_dens_DIVbolus` skipped boundary (coastal) edges, while the dynamics SSH assembly (`compute_ssh_rhs_ale`) includes them: a boundary edge contributes its single-element cross-segment flux, which closes the coastal nodes' scalar control volumes. Skipping them dropped the alongshore boundary-current transport (O(10 Sv)) from every coastal node's binned divergence, so the density-space overturning streamfunction failed to close toward the surface/light density classes at every latitude with a land boundary — while closing exactly at land-free re-entrant channel latitudes. Verified by reconstructing the per-node full-depth divergence from annual-mean u,v with both conventions: internal-edges-only reproduces the diagnostic field (corr 1.000000) and its ~16 Sv accumulated residual; including boundary edges collapses it to ~0.001 Sv. A rerun with the fix produces closed dGMOC streamlines. Whole-domain per-class sums were always zero by edge antisymmetry, which is why global conservation checks never caught this — the error only appears in zonally/latitudinally binned quantities. Every configuration with coastlines is affected.
- **Write the `ndens` output axis as double with proper units** (`io_meandata.F90`). The coordinate variable was declared `nf90_int`, silently truncating the real sigma2 bin values (e.g. 30.12) to whole numbers and collapsing fine bins into duplicate integer labels; units were the generic `m` instead of `kg/m^3`.
- **Optional `density_dmoc` min/max in the per-step info printout** (`write_step_info.F90`), guarded by `ldiag_dMOC`, reported as sigma2 — useful for choosing `std_dens` bin ranges that actually cover the simulated density span.

### EOS fix

- **`sw_alpha_beta` respects `state_equation` for linear-EOS setups** (`oce_ale_pressure_bv.F90`): with the linear EOS, alpha/beta are now consistent with the actual equation of state instead of the full TEOS-style polynomial, fixing Redi/GM neutral slopes for such configurations.

### neverworld2 toy configuration (`toy_channel_neverworld2.F90`)

- All parameters exposed via a new optional `&oce_neverworld2` namelist group (read tolerantly, so existing `namelist.oce` files keep working; a `rewind` guards the read against the preceding optional `&oce_perturb` scan ending at EOF).
- Selectable SST-restoring targets including a DINO-style profile (Kamm et al. 2025) with optional north/south cold patches and a seasonal cycle; optional DINO-style salinity restoring.
- Optional cabbeling and thermobaricity terms for the simplified equation of state.
- Latitude/longitude-dependent forcing derived from `geo_coord_nod2D` for rotated-grid invariance (wind stress deliberately injected along the model's native axes — documented in-code, since true-geographic vector rotation is ill-posed on the artificial 60-degree-periodic domain).

`main` is merged in (`177f5bf3`), so the branch is up to date with current upstream.


patrickscholz:
the truth values of neverworld needs to be updated !

JanStreffing (approved):

dsidoren (approved):

JanStreffing:
Could you show the magnitude of impact? Should we mention this in our AWI-ESM3-veg-HR CMIP7 paper as erratum?

patrickscholz:
@JanStreffing & @dsidoren : the  dmoc bug Dima has to test for the realistic core2 configuration (if it affects his natur paper), we hope there the impact is not that dramatic since it only affects coastal edges where in the real world the waters are very shallow hopefully not contributing much to the watermass transformation. However in the idealised neverworld configuration the effect was dramatic since it prevented the closing of the streamlines on the surface .

1) neverworld with dmoc bug: 
dmoc_dmoc_dgmoc_y1977-1977

2) neverworld without dmoc bug: 
dmoc_dmoc_dgmoc_y1971-1971

JanStreffing:
Ty! We will include that after checking our output. @nwieters, Can you put this on the list?

PR #1028, opened 2026-08-28
cmake: opt-in runtime and build-time diagnostic flags

suvarchal:
Adds two opt-in diagnostic levels for development builds. Both default to `off`,
and with both off the compile flags are byte-identical to before — verified by
diffing the generated flag lines.

```
FESOM_RUNTIME_CHECKS = off | checks | paranoid     # what a run does
    checks    -fcheck=bounds -fbacktrace           # ~3% overhead at -O3
    paranoid  + -fcheck=all, FP traps, -finit-real=snan -finit-integer=-9999

FESOM_BUILD_CHECKS   = off | warn | strict         # whether the build succeeds
    warn      -Wall -Wextra
    strict    also stops adding -fallow-argument-mismatch, so gfortran >=10
              reports the mismatches as errors
```

Two options rather than one, because they act on different things: asking for
runtime diagnostics should not be able to break your build.

### Intended use

1. **Local checks during development** — run the normal CTest suite against a
   checked build.
2. **A path to removing the bad flags from production.** `-finit-local-zero` and
   `-fallow-argument-mismatch` are in the production flags today and each hides a
   class of bug: the first makes uninitialised reads return `0.0` and therefore
   undetectable by construction, the second downgrades real interface errors.
   These levels let us find and fix what they hide, and drop them once the code
   is clean.
3. **Eventually CI**, so new contributions cannot introduce these silently.
   Production CI builds and runs will not catch them.

Not proposed as a CI gate yet — it is a target to work toward, not a switch to
flip. Running the local suite with `FESOM_RUNTIME_CHECKS=checks` currently fails
3 of 9 tests on pre-existing issues (a CVMix KPP shape mismatch and a
`myList_elem2D` rank remapping in the partitioner), and `FESOM_BUILD_CHECKS=strict`
does not build. Those are the debts to pay off first.

### Notes

Defined at top level rather than in `src/`: `mesh_part` is a separate CMake
project and would not otherwise see the flags — and the partitioner is where the
rank-remapping violation shows up. Applied to all four Fortran targets.

Flags are appended to each target's existing options so they win on `-O` and
`-finit-*`. Passing them via `CMAKE_Fortran_FLAGS` does not work: that variable is
prepended, so an `-O0` given that way is silently overridden by the `-O2`/`-O3`
the targets set themselves.

Default build and full local suite unchanged: 9/9.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,66 @@ if(BUILD_YAC)
     include(cmake/BuildYAC.cmake)
 endif()
 
+# Compile-time diagnostics. "strict" stops suppressing argument mismatches, which
+# gfortran >=10 then reports as errors.
+set(FESOM_BUILD_CHECKS "off" CACHE STRING "Compile-time diagnostics: off | warn | strict")
+set_property(CACHE FESOM_BUILD_CHECKS PROPERTY STRINGS off warn strict)
+
+set(FESOM_BUILD_CHECK_FLAGS "")
+set(FESOM_ALLOW_ARG_MISMATCH ON)
+if(NOT FESOM_BUILD_CHECKS STREQUAL "off")
+   if(NOT FESOM_BUILD_CHECKS MATCHES "^(warn|strict)$")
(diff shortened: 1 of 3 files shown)


JanStreffing (approved):
Nice

PR #1029, opened 2026-08-28
Single-precision-safe OIFS/OASIS coupling + XIOS output

JanStreffing:
Makes `USE_SINGLE_PRECISION` (WP=real32) work in coupled OIFS-via-OASIS builds.

**cpl_driver.F90**
- OASIS exchange buffer `exfld` typed `WP_full` (always double), matching the `oasis_Double` transient and mirroring OpenIFS' dedicated double coupling kind JPRO.
- Every FESOM-internal coupling MPI call (grid gather, same-rank bcast) uses `MPI_WP` instead of hardcoded `MPI_DOUBLE_PRECISION`, so the MPI type tracks WP.

**gen_forcing_couple.F90**
- `integrate_2D` reductions use `MPI_WP`. The raw FESOM<->atmosphere net-flux exchange in `net_rec_from_atm` is left double (skipped for `__oifs`) with an explanatory note.

**io_tracks.F90 / io_xios.F90**
- XIOS domain/axis coordinate and geometry paths made double (`real(kind=8)` / matching `mod_tracks_geometry` TG_WP), so SP + XIOS compiles.

**CMakeLists.txt**
- SP guard relaxed to allow the OIFS-via-OASIS combination; other coupled modes (ECHAM/OASIS, YAC, IFS-interface) stay guarded until converted.

OASIS accepts single-precision buffers (its put/get and grid-write interfaces are generic over real4/real8 and promote to double internally), so nothing is corrupted when WP=real32; the genuine breakage was FESOM's own hardcoded double MPI calls on `real(kind=WP)` buffers.

Draft: builds and initializes coupled; runtime coupled validation on albedo (AWI-ESM3 TCO95/CORE3, single precision) in progress.


JanStreffing:
I did some test runs. It works in AWI-ESM3. Yet to test speedup.

JanStreffing:
I did a short test with core3 coupled to tco95.

Cores | nodes/core | regime | FESOM SP speedup
-- | -- | -- | --
256 | 861 | compute-bound-ish | 1.32×
PR#940: 64 (CORE2) | ~1980 | fully compute-bound | 1.69×


GET-wait (steady) | SP @256 | DP @256
-- | -- | --
OpenIFS waits on FESOM | 51.0 s | 129.4 s
FESOM waits on OpenIFS | 14.6 s | 8.7 s

Works. Not optimized at all. But i see substantial speedup in this short test already.

src/io_tracks.F90
suvarchal, line 1: may be if we really need kind=8 we should use WP_full  instead because kind=8 may not be portable, nec compilers use kind=1,2,3. we can leave it or if you have patience fix it.

suvarchal (approved):

PR #1033, opened 2026-09-02
Fix ghost records in mean output when the time step changes (#123)

### LINKED ISSUE #123: Ghost values in the output when timestep is changes
When we write the output, the timestamp is something like `output_period - time_step`. So for monthly mean output we get the following:
```
'1978-01-31T23:50:00.000000000', 
'1978-02-28T23:50:00.000000000',
'1978-03-31T23:50:00.000000000', 
'1978-04-30T23:50:00.000000000',
'1978-05-31T23:50:00.000000000'
```

If at some point we change the timestep, and don't delete the output file (for very large simulations this is not really possible, since files are created on yearly basis), we might end up with the following records:

```
'1978-01-31T23:50:00.000000000',
'1978-01-31T23:55:00.000000000',
'1978-02-28T23:55:00.000000000', 
'1978-03-31T23:55:00.000000000',
'1978-04-30T23:55:00.000000000',
'1978-05-31T23:55:00.000000000'
```

So, there is an extra record from the previous run with larger time step. 

I am not sure how to deal with it not to make things too complicated, but it becoming a problem in the data analysis, so some additional filtering of the output is required.

JanStreffing:
Fixes #123.

## The problem

Mean-output records were stamped with `timeold`, i.e. one time step *before* the period boundary the output events fire on (`gen_events.F90` triggers on `timenew == 86400.`). A monthly record therefore carried the stamp `end_of_month - dt`. Re-running a period with a different time step produced a new, slightly shifted stamp, and since the restart scan only overwrites records with stamps >= the current time, the old record survived as a duplicate "ghost" — exactly the behaviour reported in #123.

## The fix

1. **`io_meandata.F90`**: stamp records with `timenew`/`daynew` — the exact averaging-period end, independent of `dt`. Re-running a period now lands on the identical timestamp and overwrites the record in place.
2. **`io_meandata.F90`**: the restart scan that positions `rec_count` in an existing file called `nf90_get_var` without `start=(/k/)`, so every iteration read the *first* record's time instead of record `k`'s. In practice the scan could only ever append. Fixed.
3. **`io_netcdf_file_module.F90`**: unrelated one-liner — gfortran 16.2 ICEs (all optimization levels) on the intrinsic polymorphic assignment in `add_global_att_*`; switched to sourced allocation, which is the same semantics and the pattern the file already uses in `add_var_att_*`. Needed to build on current Homebrew gcc at all.

## Regression test

`integration_pi_output_restamp` (pi mesh, 2 ranks): runs one day at `step_per_day=96`, then rewrites `fesom.clock` and re-runs the same day at `step_per_day=144` without deleting the results, asserting the record is overwritten rather than appended.

Verified locally (macOS, gfortran 16.2 + OpenMPI 5):

| build | after run A (dt=900s) | after re-running same day (dt=600s) |
|---|---|---|
| current main | `time = 85500` | `time = 85500, 85800` — ghost record |
| this branch | `time = 86400` | `time = 86400` — overwritten in place |

The test **fails on current main** and passes with the fix. `integration_pi_mpi2` also passes on this branch.

## Notes for reviewers

- Output timestamps change: a daily record is now stamped `86400` (period end, i.e. midnight of the next day in CF terms) instead of `86400 - dt`. Downstream analysis that special-cased the `-dt` offset will see the shift; that is the point of the fix, but worth flagging before 2.7.
- Files written by *previous* code versions keep their old shifted stamps, so extending such a file across a time-step change can still leave one ghost the first time. Files written by this version are stable.
- The `entry%currentDate/currentTime` stream metadata (MULTIO) still uses `timeold`; left untouched deliberately to keep this PR minimal.

### DIFF
--- a/cmake/FesomTesting.cmake
+++ b/cmake/FesomTesting.cmake
@@ -505,6 +505,103 @@ function(add_fesom_test_with_options TEST_NAME MESH_NAME STEP_PER_DAY RUN_LENGTH
     message(STATUS "Added FESOM test: ${TEST_NAME} with mesh: ${MESH_NAME}, cavity: ${USE_CAVITY}")
 endfunction()
 
+# Regression test for github issue #123 (ghost records on time-step change):
+# run one day with step_per_day=96, then rewrite fesom.clock and re-run the SAME
+# day with step_per_day=144 without deleting the results. Records are stamped
+# with the exact averaging-period end, so the second run must land on the same
+# timestamp and overwrite the record: exactly 1 record afterwards. Code that
+# stamps period_end - dt appends a second, shifted record instead (the ghost),
+# which shows up here as "current mean I/O counter = 2".
+function(add_fesom_output_restamp_test TEST_NAME)
+    set(options "")
+    set(oneValueArgs NP TIMEOUT)
+    set(multiValueArgs "")
+    cmake_parse_arguments(FESOM_TEST "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+    if(NOT DEFINED FESOM_TEST_NP)
+        set(FESOM_TEST_NP 2)
+    endif()
+    if(NOT DEFINED FESOM_TEST_TIMEOUT)
+        set(FESOM_TEST_TIMEOUT 300)
+    endif()
+
+    set(TEST_RUN_DIR "${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}")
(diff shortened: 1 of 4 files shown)


JanStreffing:
Closing in favour of a design discussion first: the timestamp change is a far-reaching output-convention change and deserves its own issue (CF time bounds, midpoint stamps, legacy-file handling) rather than riding along in a bug-fix PR. The branch stays available.

Issue #763, opened 2025-09-04
Monthly mean output written with end-of-month timestamp instead of mid-month


PeterRony:
In FESOM2.6-REcoM runs, monthly mean fields (e.g.` temperature`,  `CO2 flux`) are written with an end of month timestamp. Would it perhaps make more sense for monthly means to have a mid-month timestamp instead?

koldunovn:
Ideally it would but I don't think we are going to change it now - will breack too many things (particularelly in post processing).

JudithHauck:
as it stands it creates confusion and issues for any new users of the model, it would be cool if you could consider changing it.

koldunovn:
@pgierz @JanStreffing Will ti break ESM Tools? Also not clear what to put to the output that is requested at, say, 4 time steps.

JanStreffing:
> Also not. I would put it behind a namelist switch in the IO control, though. Then legacy tools can use the old timestamp and we can slowly move to new ones.

Ideally default to on, until someone comes and complains. Which is hopefully never.

Issue #431, opened 2023-03-01
Shortening timestep causes duplicate outputs


n-beech:
Rerunning a simulation with a shortened timestep creates a duplicate output of the first timestep after the restart. 

I think this is because the i/o counter is defined by finding the last record in the output file with a timestamp earlier than the current time and adding 1. If you need to rerun part of a simulation with a shorter timestep than was previously used (for example, after a blowup), the first timestep after the restart will not be overwritten because it will have a slightly earlier timestamp than the first output of the restarted run (old timestep - new timestep). You then have two near-duplicate records.

From io_meandata.F90:
```
               !_______________________________________________________________
                ! if the time rtime at the rec_count is larger than ctime we
                ! look for the closest record with the timestamp less than ctime
                do k=entry%rec_count, 1, -1
                    !PS if (partit%flag_debug)  print *, achar(27)//'[33m'//' -I/O-> call assert_nf B'//achar(27)//'[0m'//',  k=',k, ', rootpart=', entry%root_rank
                    ! determine rtime from exiting file
                    call assert_nf( nf_get_vara_double(entry%ncid, entry%tID, k, 1, rtime, 1), __LINE__)
                    if (ctime > rtime) then
                        entry%rec_count=k+1
                        exit ! a proper rec_count detected, exit the loop
                    end if
                    if (k==1) then
                        write(*,*) 'I/O '//trim(entry%name)//' WARNING: the existing output file will be overwritten'//'; ', entry%rec_count, ' records in the file;'
                        entry%rec_count=1
                        exit ! no appropriate rec_count detected
                    end if
                end do
```

Could the i/o counter be initialized by finding the number of records in the output file with timestamps earlier than the restart time and updated by 1 after each output?

JanStreffing:
Duplicate of #123

PR #1036, opened 2026-09-03
CF interval-aware output time axis: time_bounds, midpoint stamps, honest accumulation intervals

### LINKED ISSUE #123: Ghost values in the output when timestep is changes

### LINKED ISSUE #1034: Output time axis: interval-aware timestamps (time_bnds), and a broken restart record scan
### Summary

Three related defects in the native mean-output time handling force a decision we deferred in #763. Rather than patching them one at a time, this issue proposes redesigning the output time axis once, properly.

### Defect 1: timestamps depend on the model time step (#123, #431)

Records are stamped from `timeold`, one step *before* the period boundary the output events fire on (`gen_events.F90` fires on `timenew == 86400.`). A monthly mean is stamped `end_of_month − dt`. Re-running any period with a different dt (e.g. after a blowup, as in #431) produces a shifted stamp, the overwrite logic doesn't recognize the record, and the old one survives as a ghost. Reported in 2021; still reproducible on `main` (pi mesh, 2 ranks: run 1 day at dt=900 s, rerun the same day at dt=600 s → `time = 85500, 85800`, two records for one day).

### Defect 2: the restart record scan is broken since the nf90 conversion (regression, Jul 2025)

The scan that positions `rec_count` in an existing file (`io_meandata.F90`, ~line 3002) calls `nf90_get_var(entry%ncid, entry%tID, rtime)` without `start=(/k/)`, so every loop iteration reads record **1** instead of record k. The pre-conversion code (`nf_get_vara_double(..., k, ...)`, quoted in #431) was correct; commit a267b529 lost the record index. Since then the scan can only append after all existing records or reset to record 1 — a plain same-dt rerun of a mid-year segment now appends duplicates *and* leaves stale later records in place. This part is an unambiguous bug fix regardless of any design decision and is PR'd separately.

### Defect 3: the files don't say what interval a record describes

A "January mean" from a run started on Jan 15 is indistinguishable from a full-month mean. There is no `time_bnds`, no `cell_methods`. Every downstream consumer guesses the interval from the stamp, which is exactly why #763 concluded we can't touch the stamp: the guess is load-bearing.

### Why revisit #763's won't-fix

#763 was closed because changing stamps breaks tooling built on the current convention. But fixing the ghost bug requires a dt-*independent* stamp — the current convention cannot be preserved exactly anyway. If we must move the stamp, we should move it once, to the convention tools understand natively, and add the metadata (`time_bnds`) that makes future consumers independent of stamp conventions altogether. `cdo shifttime` can't repair archives with mixed dt (the offset varies per segment) and can't fix boundary classification for tools that group by calendar period.

### Proposed design

1. **Track the accumulation interval per output entry**: record the model time when accumulation (re)starts (init and after each write). Bounds are `[t_accum_start, t_now]` — exactly what was averaged, automatically honest for partial intervals after cold starts/restarts. The MULTIO path's `previousDate/previousTime` already embodies this concept and gets unified with it (fixing the same `timeold` off-by-dt there).
2. **Write CF metadata**: `time_bnds(time, bnds)`, `time:bounds`, `cell_methods = "time: mean"` on averaged fields. The `time` value itself: interval midpoint (CF recommendation; a January mean sits mid-January and every calendar-grouping tool classifies it correctly by construction).
3. **Fix the restart scan** (with `start=(/k/)`) and change its rule to "overwrite from the first record whose stamp lies after the new interval's *start*". This rule is convention-independent — legacy `end−dt` stamps, end stamps, and midpoints all lie inside their interval — so mixed old/new archives rerun correctly.
4. **Legacy files stay legacy**: when appending to a file without `time_bnds`, keep writing old-style stamps into that file; only new files get the new convention. No file ever mixes conventions, and yearly files mean everyone migrates naturally at the next year boundary. A standalone migration script for old archives (midpoint stamps + synthesized bounds) can live in `tools/`.
5. **Out of scope**: XIOS (manages its own CF calendar), restart/blowup files (instantaneous state).

### Open questions

- Stamp value: midpoint (CF-native, recommended) vs. exact period end (smaller visible change, but naive calendar grouping shifts records into the following period)?
- Namelist switch for legacy stamping during a transition (as @pgierz suggested in #763), or one behaviour + release notes? A switch eases pipeline migration but doubles the tested code paths.
- Retrofit `time_bnds` into legacy files on append, or keep them untouched (proposed above)?

### Existing groundwork

PR #1033 (closed in favour of this discussion) contains a working end-stamp variant of defect 1+2 fixes plus a CTest regression test (`integration_pi_output_restamp`: one day at `step_per_day=96`, rewrite `fesom.clock`, rerun at 144, assert one record) that fails on `main` and passes with a fix — reusable for whichever design lands. Verified on macOS (gfortran 16.2/OpenMPI 5) and reproducible with the standard pi-mesh test setup.

Refs: #123, #431, #763, a267b529

JanStreffing:
Implements the design agreed in #1034. Fixes #123 (and the mechanism behind #431); supersedes the discussion outcome of #763.

## What changes

Mean-output records now say what interval they describe, instead of encoding it implicitly in a time-step-dependent stamp:

```
double time(time) ;
    time:bounds = "time_bounds" ;
    time:calendar = "standard" / "365_day" ;   (from include_fleapyear)
    time:time_origin = "1948-01-01 00:00:00" ;
double time_bounds(time, axis_nbounds) ;       // [interval_start, interval_end]
sst:cell_methods = "time: mean" ;
```

The `time` value is the interval **midpoint**. Names and layout are exactly what XIOS already writes in coupled AWI-CM3 runs, so downstream tooling sees one convention across backends.

### How it works

1. **Tracked accumulation intervals.** Every stream records when its accumulation started (at stream creation and after each write). Bounds are `[t_accum_start, t_now]`: what was actually averaged. A cold start on Jan 15 produces a January record with bounds `[Jan 15, Feb 1]` instead of silently pretending to be a full-month mean. XIOS writes nominal bounds regardless, so this is one point where the native backend is now more precise.
2. **Convention-independent overwrite rule.** When appending to an existing file, the record scan overwrites every record whose stamp lies after the start of the current interval (previously: a stamp-value comparison, which is what let ghost records survive a time-step change). Any stamping convention places a record's stamp inside its own interval, so this dedups correctly even against files written by older FESOM versions with `end - dt` stamps.
3. **Legacy files stay bounds-less.** Appending to a file without `time_bounds` warns once and appends new-style stamps without bounds: one write path for everybody, no retrofitted guesses. Since files are yearly, everyone migrates naturally at the next year boundary. (Rare case by agreement in #1034; normally output is not continued across model versions.)
4. The 0D scalar streams get the identical treatment.

## What this means for downstream tooling

This is the output-convention change discussed at length in #763 and #1034:

- A daily mean is stamped `43200` (noon) instead of `86400 - dt`; a January monthly mean sits mid-January instead of `Jan-31 23:5x`. Tools that group by calendar period classify records correctly by construction. Tools that special-cased the old `-dt` offset need updating.
- Anything already consuming XIOS-written FESOM output needs no changes at all. 

## Verification (pi mesh, 2 ranks; macOS gfortran 16.2 / OpenMPI 5)

| scenario | result |
|---|---|
| `integration_pi_mpi2` | passes |
| daily record | `time = 43200`, `time_bounds = [0, 86400]` |
| re-run same day, dt 900 s to 600 s (the #123 ghost case) | one record, overwritten in place |
| 2-day run, warm-restart re-run of day 2 (the #431 case) | record 2 overwritten; bounds `[0,86400], [86400,172800]` |
| append to hand-built legacy file (stamp `85500`, no bounds) | old record kept, new record appended without bounds, warning printed, no crash |
| year rollover Dec 31 1948 to Jan 1 1949 | 1948 file: `[31449600, 31536000]`; 1949 file: `[0, 86400]`, the interval start correctly re-based into the new yearly file's units |
| same ghost case on an `ASYNCHRONOUS_IO_THREADS=ON` build | identical result; the bounds travel through the same snapshot mechanism as ctime_copy |
| `integration_pi_output_restamp` (regression test, added here) | passes; fails on `main` |

## Backend coverage

- Serial gather path and asynchronous I/O threads: tested above.
- Parallel writer (`parallel_write=.true.`): wired in by construction. The `time_bounds` definition sits in the common `create_new_file` path, the bounds write sits inside the same collective guard as the time write (every writer issues it), and the record scan already runs on the whole writer set since the parallel-writer work in #969. Not runtime-verifiable on my laptop: `nf90_create_par` needs netCDF built against parallel HDF5, and the Homebrew build is serial-only (unmodified `main` aborts identically there). One pi run with `parallel_write=.true.` on albedo would close this gap.
- XIOS is untouched; it already writes this convention itself.
- MULTIO is untouched. The design's `previousDate/currentTime` unification needs a `__MULTIO` build to verify, which I cannot do locally; those fields still carry the old `timeold`-based values. Flagged as a follow-up for whoever runs MULTIO.

## Caveats

- Rebased on current `main`, which already contains the record-scan index fix (#1035).
- Includes a gfortran >= 16 workaround in `io_netcdf_file_module.F90` (sourced allocation instead of an intrinsic polymorphic assignment that ICEs at every optimization level). Same semantics, same idiom the file already uses in `add_var_att_*`; needed to build on current Homebrew gcc at all. Can be split out if preferred.
- **0D streams are compile-tested only**: nothing on `main` registers one (`def_stream0D` callers live on RECOM/CMOR branches), so that path had no runtime exercise here.
- Restart/blowup files are untouched (out of scope per #1034).

### DIFF
--- a/cmake/FesomTesting.cmake
+++ b/cmake/FesomTesting.cmake
@@ -628,6 +628,103 @@ function(add_fesom_test_with_options TEST_NAME MESH_NAME STEP_PER_DAY RUN_LENGTH
     message(STATUS "Added FESOM test: ${TEST_NAME} with mesh: ${MESH_NAME}, cavity: ${USE_CAVITY}")
 endfunction()
 
+# Regression test for ghost records on time-step change:
+# run one day with step_per_day=96, then rewrite fesom.clock and re-run the SAME
+# day with step_per_day=144 without deleting the results. Records are stamped
+# with the exact averaging-period end, so the second run must land on the same
+# timestamp and overwrite the record: exactly 1 record afterwards. Code that
+# stamps period_end - dt appends a second, shifted record instead (the ghost),
+# which shows up here as "current mean I/O counter = 2".
+function(add_fesom_output_restamp_test TEST_NAME)
+    set(options "")
+    set(oneValueArgs NP TIMEOUT)
+    set(multiValueArgs "")
+    cmake_parse_arguments(FESOM_TEST "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+    if(NOT DEFINED FESOM_TEST_NP)
+        set(FESOM_TEST_NP 2)
+    endif()
+    if(NOT DEFINED FESOM_TEST_TIMEOUT)
+        set(FESOM_TEST_TIMEOUT 300)
+    endif()
+
+    set(TEST_RUN_DIR "${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}")
+    set(TEST_DATA_DIR "${CMAKE_SOURCE_DIR}/tests/data")
+    set(RESULT_DIR "${TEST_RUN_DIR}/results")
+
+    generate_fesom_clock("${RESULT_DIR}")
+    configure_fesom_namelists("${TEST_RUN_DIR}" "${TEST_DATA_DIR}" "${RESULT_DIR}")
+
+    set(TEST_SCRIPT "${TEST_RUN_DIR}/run_test.cmake")
+    file(GENERATE OUTPUT ${TEST_SCRIPT} CONTENT "
+        file(MAKE_DIRECTORY \"${TEST_RUN_DIR}\")
+        file(MAKE_DIRECTORY \"${RESULT_DIR}\")
+
+        # see add_fesom_test_with_options for why these are safe unconditionally
+        set(ENV{OMPI_ALLOW_RUN_AS_ROOT} \"1\")
+        set(ENV{OMPI_ALLOW_RUN_AS_ROOT_CONFIRM} \"1\")
+        set(ENV{OMPI_MCA_rmaps_base_oversubscribe} \"1\")
+        set(ENV{PRTE_MCA_rmaps_default_mapping_policy} \":oversubscribe\")
+
+        # start from a clean slate: the point of the test is the SECOND run below
+        file(REMOVE_RECURSE \"${RESULT_DIR}\")
+        file(MAKE_DIRECTORY \"${RESULT_DIR}\")
+        file(WRITE \"${RESULT_DIR}/fesom.clock\" \"0 1 1948\\n0 1 1948\\n\")
+
+        # run A: one day at step_per_day=96 (dt=900s), as configured
+        execute_process(
+            COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${FESOM_TEST_NP} ${CMAKE_BINARY_DIR}/bin/fesom.x
+            WORKING_DIRECTORY \"${TEST_RUN_DIR}\"
+            RESULT_VARIABLE run_a_result
+            OUTPUT_VARIABLE run_a_output
+            ERROR_VARIABLE run_a_error
+            TIMEOUT ${FESOM_TEST_TIMEOUT}
+        )
+        file(WRITE \"${TEST_RUN_DIR}/test_output_run_a.log\" \"\${run_a_output}\")
+        file(WRITE \"${TEST_RUN_DIR}/test_error_run_a.log\" \"\${run_a_error}\")
+        if(NOT run_a_result EQUAL 0 OR NOT run_a_output MATCHES \"fesom should stop with exit status = 0\")
+            message(FATAL_ERROR \"Test ${TEST_NAME} FAILED: first run did not finish cleanly (exit \${run_a_result})\")
+        endif()
+
+        # re-run the SAME day with a different time step, keeping the output file
+        file(WRITE \"${RESULT_DIR}/fesom.clock\" \"0 1 1948\\n0 1 1948\\n\")
+        file(READ \"${TEST_RUN_DIR}/namelist.config\" _nml)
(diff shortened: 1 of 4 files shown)


suvarchal (approved):
hard for me to verify, seems reasonable. and i am relying on the test that seems to capture the details.

patrickscholz (approved):

PR #927, opened 2026-06-03
workbench_idemix2: IDEMIX2 - Internal Wave Energy Parameterisation for FESOM2

patrickscholz:
## IDEMIX2: Internal Wave Energy Parameterisation for FESOM2

IDEMIX2 (Internal wave Dissipation, Energy and MIXing, version 2) is a prognostic parameterisation of internal gravity wave energy and its contribution to diapycnal mixing, based on Olbers & Eden (2013) and Eden et al. (2014). It is implemented here as an extension to the existing IDEMIX1/TKE mixing scheme in FESOM2.

### Concept

Unlike IDEMIX1, which tracks a single vertically integrated internal wave energy field, IDEMIX2 resolves the **horizontal propagation direction** of internal waves through a spectral angular dimension (`nfbin` directional bins, default 50 interior + 2 ghost bins covering 0–2π). This captures anisotropic wave fields near generation sites such as rough topography and coastlines, and allows wave energy to be tracked as it propagates across ocean basins.

### Wave energy compartments

Two spectral wave energy fields are evolved independently:

```
  Surface wind forcing                 M2 tidal generation
  (NIW, near-inertial)                 (rough topography)
          │                                    │
          ▼                                    ▼
  ┌───────────────┐                  ┌─────────────────┐
  │  E_niw(φ, x)  │                  │   E_M2(φ, x)    │
  │  Near-inertial│                  │   M2 tidal IW   │
  │  waves        │                  │   (~12.4 h)     │
  └───────┬───────┘                  └────────┬────────┘
          │  ω ≈ f (lat-dependent)            │  ω = ω_M2 (fixed)
          └──────────────┬────────────────────┘
                         │  wave-wave interaction
                         ▼
               ┌──────────────────┐
               │   E_iw(z, x)     │   background IW continuum
               │   (vertically    │   (GM spectrum)
               │    resolved)     │
               └────────┬─────────┘
                        │  dissipation → diapycnal mixing
                        ▼
               ┌──────────────────┐
               │   TKE scheme     │   iw_diss added as
               │   Kv, Av         │   energy source term
               └──────────────────┘
```

### Timestep structure

Each call to `calc_cvmix_idemix2` follows this sequence:

```
  ┌─────────────────────────────────────────────────────────────────┐
  │ 1. Node parameters                                              │
  │    cn (baroclinic wave speed), α_c, v0, c0, τ_M2, τ_niw,        │
  │    vertical structure functions  [loop: node, OMP parallel]     │
[...]

PR #1037, opened 2026-09-03
Documentation catch-up for the 2.8 release

JanStreffing:
Documentation catch-up for everything user visible that landed since release 2.7, plus the material that older releases never picked up.

Stacked on #1036: the base of this PR is `feat/output-time-bounds`, because the last commit documents the CF time axis introduced there. GitHub will retarget this to `main` once #1036 merges. The first commit is independent of that work.

## Why

The namelist reference pages have been kept current, but the build options, the coupling interfaces and the developer tooling were never documented. Checking the last four release intervals, the same pattern repeats: parameters get back-filled into the configuration pages, everything else is dropped. This PR closes that gap and fixes what had become wrong.

## New pages

- `getting_started/single_precision`: building and running in single precision, what deliberately stays in double, what to expect from the results.
- `recom_in_fesom`: enabling and configuring REcoM, modelled on the Icepack page.
- `fesom_as_library`: the IFS interface and the library build used by IFS-FESOM. It states explicitly that AWI-CM3 does not use this path but couples OpenIFS and FESOM through OASIS, in the text and in the figure, since that is a recurring source of confusion.
- `developer_documentation/gpu_and_dwarfs`: state of the OpenACC port and the compute dwarfs.
- `developer_documentation/profiler`: `FESOM_PROFILING` and how to read `fesom.stats`.

## Extended pages

- `output_configuration`: the parallel writer and its restart I/O, the XIOS backend including track and transect output, the CMOR diagnostics, and the time axis.
- `getting_started`: the OpenMP build, the double gyre and neverworld2 toy setups, a table of cmake options, the executable command line options and the ecbundle build path. It also no longer tells the reader to compile the Parms solver library, which was removed when the split-explicit sea surface height was introduced.
- `ocean_configuration`: the CVMix schemes, freshwater hosing, the sea surface height solver controls.
- `general_configuration`: a section on the iceberg module covering the physics, the seeding files, the grounding modes and restarts. The parameter list is corrected and completed.
- `meshes`: the standalone `fesom_meshdiag` and `fesom_meshpart` tools.
- Smaller entries for the `&meltpond` group and the liquid and solid runoff split.

## Figures

Three schematics, shipped as PNG for the html build and PDF as the vector source, matching the existing figures:

- `fig_parallel_io`: the gather-to-one-writer path against the collective parallel writer.
- `fig_recom`: the REcoM3 compartments and the fluxes between them.
- `fig_ifs_fesom_callseq`: how IFS drives FESOM as a library.

That takes the manual from three figures to six. All three are anchored, captioned and referenced with `:numref:` like the existing ones.

## Scope

Roughly 800 added lines against a manual of about 2800, so it grows by a bit under a third. Bugfixes, CI work, HPC environment ports and internal refactorings are deliberately not documented, in line with what the manual already does.

Parameter names and defaults were checked against the tree rather than taken from release notes. The prose was audited afterwards for claims that were inferred rather than verified, and the corrections are in the last commit. `sphinx-build` is clean: the warnings that remain are pre-existing and none of them are in new or changed content.

Not covered, because the code is not merged yet: IDEMIX2 (#927), partial wind coupling (#989), the DARS mesh namelists (#1004) and the CO2 coupling (#806). Each is a short section to add on top of what is here once those land.


JanStreffing:
This needs a good read to ensure that the ai did not invent reasoning.

JanStreffing:
@ackerlar one question on the iceberg section, since you know that code best.

I documented `lverbose_icb=.true.` as simply printing per-iceberg diagnostics. Is it in practice the thing you reach for first when trajectories look wrong, or is there a better first step (the grounding flags, the melt rates, something else)? If there is a usual debugging path I would rather write that down than leave the switch as a bare fact.

For context on the rest of that section: the wave radiation force, the surface slope term, the four melt terms and the seeding file format are all taken from `icb_dyn.F90`, `icb_thermo.F90` and `icb_step.F90` directly. One claim I had written from assumption, that iceberg fluxes are handed to the coupler as surface fields, was wrong and is removed: `icb_coupling.F90` puts them into the ocean forcing arrays with the heat flux spread over the iceberg's depth range. Corrections are in the last commit.

patrickscholz (approved):

ackerlar:
@JanStreffing the verbose flag helps to check whether the "cell saturation" mechanism is working properly, and hence may be helpful for weird looking trajectories or temperature blow-ups. But as a first step, I would usually look at the heat fluxes (3D) and the grounded and frozen-in flags

PR #1038, opened 2026-09-04
Mixing diagnostics

dsidoren:
Diagnostic of numerically induced diapycnal mixing.
Enabled by ldiag_diapmix (default off, implies ldiag_dMOC), evaluated every diap_call_freq units. diag_densMOC is also optimized: the per-element binning loop and its five outputs are removed, and the linear scans over the density classes become binary searches.


config/namelist.io
```diff
@@ -25,6 +25,13 @@ ldiag_Ri          = .false.  ! enables Richardson number diagnostics ('shear', '
 ldiag_turbflux    = .false.  ! enables turbulent flux diagnostics ('KvdTdz', 'KvdSdz')
 ldiag_salt3D      = .false.  ! enables 3D salinity diagnostics
 ldiag_dMOC        = .false.  ! enables 'dMOC' output (density MOC diagnostics)
+ldiag_diapmix     = .false.  ! adds 'w_diap_diff', 'dV_diap_dT' to the 'dMOC' output: diapycnal velocity
+                             ! and density class uplift induced by the parameterized vertical mixing,
+                             ! from which the numerically induced diapycnal velocity and spurious MOC
+                             ! are formed in postprocessing. Requires ldiag_dMOC (ldiag_diapmix 
+                             ! is automatically switched off with a warning if ldiag_dMOC unset)
+diap_call_freq    = 1        ! evaluate ldiag_diapmix every N units of diap_call_freq_unit ...
```
JanStreffing, line 33: Should we also set a default of 1 day for the main ldiag_dmoc?
Should these run on the same timestep, or does it make sense to control them separately

dsidoren: No, it’s unrelated. `diap_call_freq` indicates how often we call this diagnostic; in between calls, we accumulate the values. Calling it at every timestep might be too expensive.

JanStreffing (approved):

PR #976, opened 2026-08-13
code with recom as submodule and tracer parallelisation

a270105:
tested for:
fesom standalone
fesom-recom standalone (with recom code in the submodule branch 'yings_branch', with and without tracer parallelisation)
awiesm2

PR #1042, opened 2026-09-07
Recom dev

ogurses:


JanStreffing:
Could you describe what this is @ogurses, It also has a lot of merge conflicts? Comes from on old branch?

JanStreffing:
Trial-merged this against current `main` (a62f1806). Most of it is already in, which is what generates the conflicts.

**Already merged elsewhere:**

| here | already on `main` via |
|---|---|
| `80bca998` update tracer paralellization code | #976, as `9a1ddd27`. 452 of its 468 substantial lines are byte-identical; the rest differ only because #976 later used `MPI_WP` instead of `MPI_DOUBLE_PRECISION` |
| `030d227d` `__usetp` missing code pieces | #806, 100% |
| `31707e2a` `__usetp` in CMakeLists | #976 |
| `ad7884be` `sbc_ini_recom` split, `sbc_do_recom` in all cases | #976 |
| `dac46094` `CASE (1023:1037)`, tracer 1037 in `namelist.tra` | #806 |
| `d5118f4e`, `1ad91b80`, `19e1afbb` `recom()` argument list | #976 |
| all `src/recom` bumps | `main` is at `28760bf6`, which has this PR's `4f24178a` as an ancestor |

`80bca998` alone accounts for 9 of the 15 conflicting paths: #976 is not an ancestor of this branch, so git sees two independent rewrites of the same code and conflicts on every hunk. #806 created no new conflicts, it only enlarged four existing ones.

**Still missing from `main`:**

1. `eba8578e` ALBEDO_INTELMPI_WORKAROUNDS. #1013 put `export ENABLE_ALBEDO_INTELMPI_WORKAROUNDS=''` in `env/albedo/shell` with no CMake side reading it, so the variable is dead on `main` today. Is the `files(200)` to `(300)` bump in `io_restart_file_group.F90` part of this, or separate?
2. `a22e88eb` MEDUSA sediment and loopback refactor, 99 lines net removal, plus the "not used!" message still printing on both branches at `gen_surface_forcing.F90:2356`.
3. `88d2011b` REcoM CASE-block documentation.
4. `9a545609` DIC_PI from namelist instead of hardcoded `filelist(5)`.
5. The `if (nzmin > nzmax) cycle` guards from `ad7884be`, and the named `DICremin` stream from `dac46094`.
6. `oce_adv_tra_fct.F90:150-151`, element-based to node-based FCT bounds. This changes results for every cavity run, not only REcoM, so it wants its own PR and a cavity test rather than sitting inside a REcoM forcing commit.

**Suggestion:** close this and open small PRs off current `main` for those six. In most of the 15 conflicting files the correct resolution is "take `main` entirely", so merging here is a manual reconstruction with real risk and little to show for it. Item 1 should go in almost conflict-free on its own.

PR #1050, opened 2026-09-08
Select the coupling interface at build time and the partner at run time

k202077:
Reworks how the coupling configuration is expressed: *which* interface is
compiled becomes one cmake variable, *which* partner is talked to becomes a
runtime namelist choice, and the ice tracer layout follows from that choice
instead of from the preprocessor. The last three commits fix YAC detection
and linking.

## Compile-time: one cmake variable

```
cmake -DFESOM_COUPLING=<standalone|direct|oasis28|oasis50|yac>
```

`cmake/FesomCoupling.cmake` is the single source of truth. Exactly one
interface macro is defined structurally, and an unknown value is a
`FATAL_ERROR`. Shared code tests derived group macros rather than
enumerating interfaces: `__cpl_enabled` (an external component is present),
`__cpl_coupler` (a coupler library owns MPI init, `MPI_COMM_FESOM` and
teardown), `__cpl_oasis` (either OASIS generation).

The old booleans (`FESOM_COUPLED`, `OIFS_COUPLED`, `USE_YAC`,
`ENABLE_IFS_INTERFACE`, `BUILD_FESOM_AS_LIBRARY`) still map onto
`FESOM_COUPLING` with a `message(DEPRECATION)` for one release.
`ENABLE_IFS_INTERFACE` stays derived from `FESOM_COUPLING=direct` because it
drives the `src/ifs_interface/*` source list and the package export.
`__oasis` is still emitted for the `src/recom` submodule.

## Run time: config/namelist.cpl

New `src/cpl_config.F90` reads `&coupling_partner` (exactly one
atmosphere), `&coupling_oasis` and `&coupling_yac`. The partner selection
drives the `cpl_send`/`cpl_recv` field tables and their counts, the
ECHAM-only flux-correction machinery (`force_flux_consv`,
`net_rec_from_atm`, `exchange_roots`), and the ice tracer layout.

`read_cpl_namelist()` runs before `MPI_Init` - under `oasis50`, OASIS calls
`MPI_Init` inside `cpl_oasis3mct_init` and already needs `cpl_comp_name`.
Hence plain Fortran I/O, `error stop` rather than `par_ex`, and no
printing; `check_cpl_config()` reports and validates after `setup_model`,
and folds in the deprecated `compute_oasis_corners` from `&run_config`.

## Ice tracer layout

`set_ice_tracer_layout` in `MOD_ICE.F90` is now the only place that decides
it. Slots 1-3 are always area, ice thickness and snow thickness; the
IFS-family partners add the prognostic surface temperature, and
`use_icebergs` adds a concentration/thickness pair. Each index is 0 when
its tracer is absent, so a consumer tests its own index instead of
subscripting relative to `size(ice%data)`, which silently aliased
`m_ice`/`m_snow` when the count did not reserve the iceberg slots.
`ASYNC_ICEBERGS` is removed: it only ever sized this array, and did so
independently of the `use_icebergs` that gates the consumers. `READ_T_ICE`
now rejects a bin restart whose tracer count disagrees with the allocated
`ice%data` instead of reading out of bounds.

This changes the binary restart layout for iceberg-free runs (5 tracers to
3), which is the mismatch `READ_T_ICE` reports.

## YAC detection and linking

- **Require YAC >= 3.2.0.** `cpl_yac_driver.F90` uses the Fortran module
  `yac`, which YAC provides from 3.2.0 on; before that the interface module
  was `mo_yac_finterface`. Nothing said so, so an older YAC configured
  cleanly and then failed to compile the driver with a missing `yac.mod`.
  3.2.0 is also the release that added `yac-mci.pc` and deprecated
  `yac.pc`, so `yac-mci` is checked first now. Also fixes the fallback
  test, which read `yac_FOUND` where `pkg_check_modules` sets
  `PC_yac_FOUND`, so it always ran and discarded a successful `yac.pc` hit
  on an installation without `yac-mci.pc`.
- **`yac_fenddef`.** The generic interface has no variant taking an error
  code, so `yac_fenddef(ierr)` selected `yac_fenddef_instance` with an
  unassigned instance id.
- **RPATH.** Linking `fesom.x` against a YAC installation with shared
  libraries failed with `libyac_mci.so, needed by libfesom.so, not found`.
  YAC arrives as raw pkg-config `-L/-l` flags, so
  `CMAKE_INSTALL_RPATH_USE_LINK_PATH` has no library path to work from. The
  pkg-config library directories are now appended to the RPATH of
  `libfesom.so`, the object carrying the `DT_NEEDED` entries; the
  executables need nothing themselves.

## Notes for review

- **Coupled builds register no integration tests.**
  `tests/integration/CMakeLists.txt` skips them when
  `FESOM_COUPLING_COUPLER` is set, because such a `fesom.x` aborts at
  startup without a `namcouple`/`coupling.yaml`. `FESOM_COUPLING=direct` is
  not skipped - it links no coupler and still yields a runnable standalone
  binary. The `coupled*` presets are build-only verification and use
  `noTestsAction: ignore`.
- **Known asymmetry, unchanged here:** the YAC path calls neither
  `net_rec_from_atm` nor `force_flux_consv`, and `a2o_fcorr_stat` is
  allocated only in the OASIS branch although `cpl_yac_driver` declares it.
- **Left compile-time deliberately**, both marked with a `TODO`: the
  bin-restart stream of `T_ICE_ATMCOUPL`, and the `ice_temp`/`alb` restart
  and output field lists. Widening them changes file layouts.

## Verification

Because no local build reaches the coupler-linked code, the macro-only
commits were checked by diffing `gfortran -cpp -E -P` output per
configuration before and after (byte-identical modulo comments), and
`oasis28`/`oasis50` were compiled and linked against a stub `mod_oasis`/
`mod_prism` with the `FindOASIS.cmake` cache variables pre-seeded.

The YAC commits were built against a real YAC 3.21.0.0 shared install:
configure reports `Found YAC 3.21.0.0`, `fesom.x` links, and `ldd` resolves
all five `libyac_*.so` without `LD_LIBRARY_PATH`. A too-old or absent YAC
now fails at configure time with the version it found. `BUILD_YAC=ON` and
the standalone build are unaffected.

CI runs the build presets; a reviewer with a real partner component is
still the only end-to-end check of the exchanged field sets.


sveta-loza:
Hi Moritz,

following up on what we discussed earlier: I have taken your recommendations on the TerraDT sea-ice split (now put on top of this PR). In TerraDT the FESOM2 sea ice runs as a separate YAC component ("FESIM", built from the FESOM2 source tree with its own driver) next to the ocean and ICON-A, and the runtime partner selection introduced here turned out to be exactly the mechanism it needed. The result is on my fork:

- https://github.com/sveta-loza/fesom2/tree/fesom-ocean-yac-decoupled-pr1050 (ocean, 4 commits on the PR head)
- https://github.com/sveta-loza/fesom2/tree/fesim-yac-decoupled-pr1050 (sea-ice executable, 3 commits)

Verified so far on CORE2 at DKRZ Levante: FESOM+FESIM standalone (forcing files), a restart of that pair, and ICON-A+FESOM+FESIM three-way, all with the partner chosen at run time in namelist.cpl. No compile-time switch selects the atmosphere any more; one ocean binary and one sea-ice binary serve every case.

What I added to the runtime configuration, and would like your opinion on:

1. is_coupled_to_fesim in &coupling_partner -- the "future second component
   role" your commit message mentions. With it the ocean skips its built-in ice
   step and exchanges ocean state and ice state with FESIM over YAC; the ocean
   forwards the atmosphere's fluxes to the ice (ICON or IFS) or, with no
   atmosphere, the raw atmospheric state from its forcing files. check_cpl_config
   therefore requires at most one atmosphere and at least one partner of any
   role, and a small helper cpl_has_atmosphere() decides whether the forcing
   files are read under a coupler.

2. cpl_component_is_sea_ice -- set by the FESIM driver before the namelist is
   read. The sea-ice executable reads the same namelist.cpl as the ocean; under
   this flag the atmosphere flags describe the ocean's atmosphere (what is
   forwarded), and check_cpl_config skips the per-interface checks for them.
   Would you prefer this as a namelist entry, or as a property of a separate
   cmake target, rather than a driver-set variable?

3. The sea-ice thermodynamics is dispatched at run time in the same spirit:
   thermodynamics() calls thermodynamics_cpl (ice_thermo_cpl.F90) when an
   atmosphere provides the fluxes and thermodynamics_bulk (ice_thermo_oce.F90)
   otherwise, both compiled in a coupled build. The IFS area-weighting exception
   in the coupled scheme follows is_coupled_to_ifs instead of __cpl_direct.

4. The one-interface rule versus IFS: our third configuration has IFS driving
   the ocean in-process (direct) while the sea ice is a YAC component, i.e.
   direct + yac in one ocean binary. The code for it is ported and compiles, but
   FESOM_COUPLING cannot express the combination. Is there a plan for IFS to
   reach FESOM through YAC, or would you consider allowing direct together with
   yac?

Two smaller things from the port that may be of interest for the YAC path in
general: (a) the components now hand YAC both the start and the end datetime
(start = clock, end = start + nsteps*dt); a start_date in coupling.yaml, which a
run without a time master needed before, breaks every restart with an
"inconsistent start datetime" abort. (b) The field time step is registered in
milliseconds and yac_fenddef() is called without argument, as in the PR.

The whole cpl_config change is about 60 lines (cpl_config.F90, namelist.cpl,
docs/coupling_configuration.rst). I would be glad to open it as a PR on top of
#1050, or to fold it in whatever way suits you.

Best,
Svetlana

k202077:
Hi [Sveta](@sveta-loza),

First of all, I would propose to wait with the final decisions on FESIM integration until this PR has been merged. This avoids later conflicts in case it still requires some adjustments in order to be accepted.

## FESOM/FESIM:

Do they share the same infrastructure?

How are they run?
1. SPMD: single executable, which is split at runtime into FESOM and FESIM ranks? (same as ICON with its atmosphere, ocean, and wave component)
2. MPMD: FESOM and FESIM share share most of the infrastructure but are built into individual executables.
(In my opinion the second option handles easier; role discovery at startup and placement on hardware by scheduler, especially in het-job cases)

## Your questions:

on 1.
- `is_coupled_to_fesim` that was exactly the intended purpose of `&coupling_partner`
- `cpl_has_atmosphere` is required by FESIM to determine at runtime whether it is coupled to an atmosphere? If built without coupler support, this routine probably would default to `.FALSE.`. I am not entirely sure whether this is not a property of the basic FESIM component configuration, however it is easy to derive it from the coupling configuration. An alternative name would be: `is_coupled_to_atmosphere`, which would return something like `is_coupled_to_ifs() .OR. is_coupled_to_oifs() .OR. is_coupled_to_icon_a()`.

on 2.
My answer would depend on your answer on my questions about FESOM/FESIM relationship (see above). I would suggest to think very carefully about the SPMD vs. MPMD question, as it carries quite some consequences.

on 3.
`thermodynamics_cpl`: the `cpl` looks like the `cpl_` prefix; `thermodynamics_coupled` is slightly longer, but more explicit in its meaning.
Using both `__cpl_direct` and `is_coupled_to_ifs` mixes two topics (coupling interface selection and coupling partner configuration). Could this be simplified? Make sure that all possibilities (interface and partner) are properly covered.

on 4.
Currently, the design only accepts a single coupling interface (direct, oasis2.8, oasis5.0, or yac). For now I would like to avoid adding support for more complex setups.
Within TerraDT an IFS-NEMO configuration coupled via OASIS (probably 5.0 or newer) is currently in development at BSC. It is planned to extend the new coupling infrastructure to also support YAC. Therefore, the best option for coupling IFS/FESOM/FESIM is probably to wait until this work has finished and then implement the coupling via OASIS and/or YAC.

## Additional comment

FESOM probably has its own configuration for its start and end date. To avoid duplicated definition and inconistencies, it is therefore probably better to leave out the start and end date from the coupling configuration file and provided it via the API directly.

The abort probably stems from an inconsistent use of the `lag` option. Documentation on time handling in YAC was recently extended (see [Time handling](https://yac.gitlab-pages.dkrz.de/YAC-dev/d4/d48/time_handling.html)). In case of open questions or persistent issues, contact me directly.

Why do you register field timestep in milliseconds?

Cheers,
Moritz

PS: My answers may read like AI instructions, which is probably what you were looking for. ;-)

sveta-loza:
Hi Moritz,

many thanks for the detailed reply. We will follow your suggestion and wait with the final decisions on the FESIM integration until this PR is merged; until then, we will not open anything on top of it.

A brief word on why I wrote here in the first place. Running the FESOM2 sea ice as a separate component (FESIM) alongside the ocean, with the two coupled back through YAC, is one of the TerraDT tasks, and a draft of how to implement this was discussed with you some time ago. The work itself was done before this PR existed and currently lives on my fork:

- https://github.com/sveta-loza/fesom2/tree/fesom-ocean-yac-decoupled (ocean side)
- https://github.com/sveta-loza/fesom2/tree/fesim-yac-decoupled (sea-ice executable)

The tags `*-verified-2026-09-11` mark the state that was tested: standalone FESOM+FESIM, ICON-A+FESOM+FESIM, restarts, and two high-resolution meshes.

When your PR appeared, we wanted to consolidate this work and check how the decoupled ice fits with the runtime partner selection introduced there; the `*-pr1050` branches from my previous comment are the result. It would be great if the FESOM/FESIM decoupling and its YAC coupling could at some point be reviewed by the YAC team (by you, Nils, or Benjamin), since the whole design relies on YAC.

We use MPMD, i.e. your option 2: FESOM and FESIM are two executables started with a single `srun --multi-prog`, so they share MPI_COMM_WORLD (with ICON-atmosphere as the third program in the coupled case), and each obtains its component communicator from `yac_fget_comp_comm`. Both are built from the FESOM2 source tree: FESIM uses the same code base, with its own driver (`fesim_module.F90` / `fesim_main.F90`) and the corresponding YAC coupling interfaces. The two components can use independent time steps and decompositions.

Your remarks on points 1, 3 and 4 are noted :). Thanks!

The start and end dates are clear now. Thanks! They are already provided through the API only (`yac_fdef_datetime(start, end)`, with the start taken from the FESOM clock and the end calculated as `start + nsteps*dt`), and `coupling.yaml` no longer contains any dates — that was the fix. The abort mentioned before came from the earlier state, not from the lag.

Thanks also for the pointer to the extended time-handling documentation. Both sides use lag 1 with the same time step, and the exchange counts match.

Regarding milliseconds: we simply followed `cpl_yac_driver.F90` on this branch (`WRITE(dt_str,'(I8.8)') INT(dt*1000)` together with `YAC_TIME_UNIT_MILLISECOND`), so that both components register the field time step in the same way as the ocean does. We can of course use a different format, as we did previously.

I will get in touch on Mattermost for further discussion.

Cheers,
Sveta
P.S.  :) Thanks a lot for your great instructions!

k202077:
Hi Sveta,

Originally, I would have assumed that this MR would cause quite some issues with the FESIM work. But I glad to hear that they work nicely together.

Just contact me, if you need a detailed review of an FESIM related PR.

Regarding `MPI_COMM_WORLD`: I have plans for a dedicated PR on the handling (splitting) of `MPI_COMM_WORLD`.

Keep in mind that once this PR has been merged, I will continue with the rework of the coupling infrastructure. In theory this is supposed to make extending of coupling capabilities easier. However, it will still have an effect on your developments, if they are not already merged into `main`.

Regarding time units: do whatever fits best for you (YAC will convert internally to ISO anyway)!

## On your original 2. question:

Since FESOM and FESIM are run in MPMD mode, `cpl_component_is_sea_ice` is not really a property of the coupling infrastructure, but more of the basic runtime configuration. I would try to keep out the difference between FESOM and FESIM from the coupling infrastructure. This is an additional complexity that can be avoided.

The partner selection section in the namelist differ between FESOM and FESIM (`is_coupled_to_fesim` vs. `is_coupled_to_fesom`). I see two options, either provide dedicated namelist-code for each component or add an additional `can_couple_to_*`, that list supported partners of the respective component and aborts on mismatch (see [here](https://gitlab.dkrz.de/icon/icon-model/-/blob/release-2026.04-public/src/namelists/mo_coupling_nml.f90?ref_type=heads#L159-185) for the solution in ICON).

Cheers,
Moritz

k202077:
@suvarchal this PR seems to fail due to an issue with the added AI attribution guard, but it looks like an issue in the guard itself (I have removed all AI attribution from the commit messages)

suvarchal:
> @suvarchal this PR seems to fail due to an issue with the added AI attribution guard, but it looks like an issue in the guard itself (I have removed all AI attribution from the commit messages)

we fixed it. hopefully should work smoothly here on.

k202077:
> > @suvarchal this PR seems to fail due to an issue with the added AI attribution guard, but it looks like an issue in the guard itself (I have removed all AI attribution from the commit messages)
> 
> we fixed it. hopefully should work smoothly here on.

How do I trigger a recheck? :thinking:

suvarchal:
Are my skills even relavent, 😅. What would I do without Claude 🤔

k202077:
@JanStreffing or @suvarchal: do you have a timeline for the review and merge of this PR? [Sevta](@sveta-loza) would be interested in it. Additionally, I could plan my next FESOM refactoring steps.

suvarchal:
Sorry, very much interested. Latest today or probably tomorow.

k202077:
> Sorry, very much interested. Latest today or probably tomorow.

Thank you very much!

The original AI attribution in the commit message probably already hinted at the fact, that this is mostly vibe coded. I carefully reviewed all the code and documentation (with limited knowledge about FESOM internals). However, I have basically no expertise in respect to CMake and fully depended on Claude in that regard.

I will probably not be available for the next two weeks.

PR #1053, opened 2026-09-09
gen_ic3d: note why the IC read needs a contiguous buffer

koldunovn:
Comment only, no behaviour change.

Starting from EN4 initial conditions used to segfault on rank 0 with the Intel build while working fine with GNU. Cause: `getcoeffld` passed the strided section `ncdata(2:nc_Nlon-1,:,:)` straight to the netCDF read. With an implicit interface the compiler has to build a copy-in/copy-out temporary of the whole field, and ifort puts that temporary on the stack.

- EN4, 1440x720x42 doubles = 332 MiB -> overflows the 200 MiB stack our run scripts set (`ulimit -s 204800`), SIGSEGV
- PHC, 360x180x33 = 16 MiB -> fits, which is why only some inputs ever failed
- gfortran heap-allocates the temporary, hence the compiler split

Reproducer under the levante Intel flags: passing the strided section to an external routine segfaults at 200 MiB stack; the same source under gfortran runs. Nothing to fix in the code -- 5a91ce60 already replaced this with the contiguous rank-1 `ncdata_inner` buffer when adding single-precision support, and that version is safe. This just records why the buffer has to stay, so it does not get simplified back.

### DIFF
--- a/docs/getting_started/getting_started.rst
+++ b/docs/getting_started/getting_started.rst
@@ -295,6 +295,13 @@ In the job file, the changes are done based on the HPC you are using. For ``leva
 
 - ``#SBATCH -A <account>``: define your project account.
 
+.. note::
+   The job scripts also set the stack size with ``ulimit -s`` before launching the model.
+   Keep that line and keep the limit generous (``unlimited`` where the site allows it).
+   Some compilers put large array temporaries on the stack, and reading a high-resolution
+   initial-condition or forcing file can need a few hundred MiB of it. Too small a limit
+   shows up as a segmentation fault during initialisation, not as an out-of-memory error.
+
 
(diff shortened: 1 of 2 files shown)


JanStreffing:
Perhaps a note in the doc about ulimit size?

koldunovn:
Added in 8bfc37e4 — a note next to the job script settings and a troubleshooting entry. Framed as the general stack-size requirement rather than a gen_ic3d bug, since 5a91ce60 already fixed that path.

JanStreffing (approved):

sebastianbeyer:
I don't see why this array should be stack allocated at all... Shouldn't we rather try to have it be consistently put to heap?

PR #1054, opened 2026-09-10
Keep the FCT low-order tendency in flux form in single precision

JanStreffing:
In single-precision builds the FCT path does not conserve tracer content. `oce_tra_adv_flux2dtracer` recovers the low-order tendency as `LO*hnode_new - ttf*hnode` from the low-order solution `LO`, which is stored as a tracer value. Where the per-step change is below half an ulp of the tracer, `LO` rounds back to `ttf` and that cell's low-order flux divergence is lost. Both the horizontal and the vertical low-order parts are booked into `dttf_v`, so the vertical advection no longer telescopes.

Measured in coupled AWI-ESM3 (CORE3, `linfs`), 5-day runs from the same restart, global ocean heat change minus surface heat input:

| build | mean |
|---|---|
| SP | -0.50 W/m2 |
| DP | -0.01 W/m2 |
| SP with this change | +0.01 W/m2 |

Over 40 years the SP deficit stayed at -0.52 W/m2 against -0.03 in DP.

In single-precision builds the tendency is now taken from the flux divergence directly (`lo_tend`). `LO` is still used for the limiter bounds. The flux form is switched off when the implicit vertical-advection split is on, because `adv_tra_vert_impl` updates `LO` itself. Double precision takes the unchanged code path.

Tested with `linfs` only. A `zstar` run showed the same signature (-0.32 W/m2 in SP, -0.001 in DP) but was not repeated with the change.

Follow-up to #940.

### DIFF
--- a/src/oce_adv_tra_driver.F90
+++ b/src/oce_adv_tra_driver.F90
@@ -1,3 +1,13 @@
+! Single precision cannot resolve the per-step change of the FCT low-order solution
+! LO, so LO*hnode_new - ttf*hnode loses it. The low-order tendency is kept in flux
+! form here instead.
+module adv_lo_tend_mod
+  use o_PARAM, only: WP
+  implicit none
+  real(kind=WP), allocatable, save :: lo_tend(:,:)
+  logical,                    save :: lo_flux_form = .false.
+end module adv_lo_tend_mod
+
 module oce_adv_tra_driver_interfaces
   interface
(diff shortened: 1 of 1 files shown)


JanStreffing:
If wanted, we could always do it in flux form. That would mean less preprocessor flags. But I would also mean changing DP results. We have already begun our final eval (300/500y done) so I suggest to do that after 2.8.

patrickscholz (approved):

patrickscholz:
in FESOM3 we want to do everything in flux form

PR #1057, opened 2026-09-10
Add Fox-Kemper mixed-layer eddy restratification (default off)

JanStreffing:
This adds `use_mle` to `&oce_dyn`, the mixed-layer eddy streamfunction of Fox-Kemper et al. (2008, 2011). It is added to `fer_gamma` before `fer_gamma2vel`, so the bolus velocity, the vertical velocity, tracer advection and the GM diagnostics all pick it up unchanged. It works with or without `Fer_GM`. It is off by default. With `use_mle = .false.` the only change is that the GM work arrays are allocated when either scheme is on.

- **Frontal width.** `Lf = max(N*H/f, mle_Lf_min)`, where `f` is regularised towards the equator with `mle_tau`. The resolution factor `ds/Lf` uses `mesh_resolution` per node and is capped by `mle_resscale_max`.
- **Mixed-layer depth.** `H = min(MLD2, mle_hmax)`. It can optionally be passed through a one-sided running mean (`mle_mld_decay_time`), which is not carried in the restart.
- **Level range.** The streamfunction is limited to `ulevels_nod2D_max..nlevels_nod2D_min`, as in `fer_solve_Gamma`, so the column bolus transport integrates to zero.
- **Velocity bound.** Each column is scaled so that the induced bolus velocity stays below `mle_ustar_max`.
- **Output.** `mle_psi` is a new output field. It is registered through `def_stream`, so it is not written when output goes through XIOS.

**Testing** used AWI-ESM3 (OpenIFS 48r1, CORE3 mesh, linfs, single precision): one year each from the same restart, with `use_mle` on and off. The mixed-layer depth uses the WOA18 density criterion.

- The mixed layer shoals in every latitude band on the annual mean, by 1.9 to 17.5 m. It shoals most in winter: 62 m at 45–60N in March, and 48 m at 30–45S in October.
- Two further one-year runs completed without instability: one without the `mle_hmax` cap, and one with the frontal width as a field.

**This is a draft for three reasons:**

- The added bolus streamfunction at the mixed-layer base, diagnosed from monthly-mean bolus output, is 0.00–0.03 m2/s. An offline estimate predicted 0.9–2.0 m2/s at the outcrops. I do not understand this discrepancy yet.
- The test runs predate #1054.
- The commit was tested on an older base. It cherry-picks cleanly onto current main but has not been compiled there.

### DIFF
--- a/config/namelist.oce
+++ b/config/namelist.oce
@@ -47,6 +47,16 @@ scaling_GMzexp     = .true.  ! do downscaling of GM & Redi over depth by exp(-z/
 GMzexp_zref        = 500.0   ! reference for vertical exp GM scaling 
 GMzexp_smin        = 0.6     ! minimum scaling factor
 
+! --- Mixed Layer Eddy Restratification (Fox-Kemper et al. 2008, 2011) ---
+use_mle            = .false. ! enable MLE restratification (GM is tapered out of the mixed layer by ODM95)
+mle_Ce             = 0.06    ! efficiency coefficient
+mle_Lf_min         = 1000.0  ! floor on the frontal width (m), Lf = max(N*H/f, Lf_min)
+mle_tau            = 86400.0 ! frontal lifetime (s), regularises f towards the equator
+mle_hmax           = 500.0   ! max. mixed layer depth used (m), Psi scales with H^2
+mle_resscale_max   = 20.0    ! max. resolution factor ds/Lf
+mle_ustar_max      = 0.05    ! max. induced bolus velocity (m/s), column scaled to satisfy it
+mle_mld_decay_time = 0.0     ! running-mean decay time for H (s), 0 = use instantaneous MLD
+
 scaling_GINsea     = .false. ! Increase K_GM resolution scaling only in the GIN sea by factor 
 GINsea_fac         = 2.0     ! GIN sea K_GM upscaling factor 
 
--- a/src/io_meandata.F90
+++ b/src/io_meandata.F90
@@ -597,6 +597,7 @@ subroutine ini_mean_io(ice, dynamics, tracers, partit, mesh)
           "benC                ", "benCalc             ", "benN                ", &
           "benSi               ", "beta                ", "bolus_u             ", &
           "bolus_v             ", "bolus_w             ", "calcdiss            ", &
+          "mle_psi             ", &
           "calcif              ", "cfl_z               ", "Chldegc             ", &
           "Chldegd             ", "Chldegn             ", "CO2_aq              ", "CO2f                ", "CO3                 ", &
           "curl_surf           ", "dens_flux           ", "d_eta               ", &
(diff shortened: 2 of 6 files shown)


JanStreffing:
I would like to get this merged as default = off for 2.8.0 already.

patrickscholz:
This stuff needs to be tested well, how it interferes with our GM implementation. before we can make ON as default.

JanStreffing:
Yes, I will post here, what I can already show from coupled simulations. But it needs standalone eval too.

JanStreffing:
Here is what the coupled AWI-ESM3 runs show so far. A standalone FESOM evaluation is still needed.

**Setup.** AWI-ESM3 (OpenIFS 48r1, CORE3 mesh with cavities, linfs, single precision including the #1054 fix), pre-industrial forcing. Two runs from the same restart differ only in `use_mle`, over 10 years. The MLE code is exactly this PR's two commits. The runs used it on our working branch, not on current main. MLE settings are `mle_Ce 0.06`, `mle_tau 1 d`, `mle_ustar_max 0.05`, `mle_Lf_min 1000`, `mle_mld_decay_time 0`, and `mle_hmax 1e5`, so the depth cap is off.

MLE on minus off

Panels (a) and (b) show MLE on minus off, with 95 % error bars from the year-to-year scatter. Heat uptake is in W/m2 of Earth area. Panel (c) shows the deepest-month mixed-layer depth bias against WOA18, using the WOA density criterion, for years 6 to 10.

**Mixed layer.** MLE shoals the too-deep winter mixed layer where it is too deep:

- **30-45N:** from +50 to +28 m.
- **45-30S:** from +35 to +2 m.
- **45-60N:** from -46 to -13 m.

It over-shoals the subantarctic, where 60-45S moves from -40 to -103 m. It does not change the Antarctic deep convection.

**Heat uptake.** Total ocean uptake falls by 0.20 ± 0.12 W/m2. Almost all of that is in 100-700 m (-0.24 ± 0.39). At 700-2000 m the change is -0.05 ± 0.08, and below 2000 m it is 0.00 ± 0.03.

**Net TOA.** It falls by 0.21 ± 0.18 W/m2 over the 10 years. All of these runs are still adjusting over that period, so a longer pair is needed before this counts as a lasting effect.

**AMOC** (maximum below 500 m). It does not change: +0.1 ± 0.5 Sv at 26.5N and +0.3 ± 0.5 Sv at 40-60N.

**Sea ice.** SH February extent drops from 4.65 to 3.28 million km2, against 3.0 observed. NH March extent rises by 0.7 million km2.

**Stability.** The runs had no instability.

**Open.** The mixed-layer eddy streamfunction diagnosed at the mixed-layer base was only 0.00-0.03 m2/s in the one-year test described in the PR text. The offline estimate was 0.9-2.0 m2/s. That has not yet been re-diagnosed on these runs, and the mixed layer shoals regardless.

These results support merging with `use_mle` off by default. They do not support switching it on yet.

patrickscholz (approved):

PR #1058, opened 2026-09-11
Describe all remaining namelists, identical to the defaults where values match

JanStreffing:
Follow-up to #1004: every remaining namelist in `config/` gets the treatment the default namelists already have, a header, section banners and a description with units for every parameter. No setting changes anywhere.

## Diffs against the default show only real differences

People diff these files against the defaults, so a variant now copies the default's line byte for byte wherever its value is the same. Where a value differs only the value text changes and the description stays. Keys a variant does not set are removed rather than added (adding them would silently switch them from the Fortran built-in to the default file's value), and keys only a variant has keep their own lines. A multi-line value such as `io_list` or a tracer table is taken from the variant as it was.

Notes that only a variant carries are kept: the CORE2 `ForcingDataPath` line, the tracer IDs, DIC and gfortran notes in `bin_2p3z2d/namelist.tra`, and the commented-out alternatives in `bin_4p3z2d/namelist.recom`. Dropped are stale commented alternatives (old `/work/ollie` paths), old file headers, and notes the defaults already carry.

## Commits

1. **Variants of the described defaults (48 files):** the `bin_*` configurations, the toy setups, the forcing and io variants, `ice_ERA5`, `oce.core2` and `tra.recom.2p3z2d`. This commit also fixes four wrong descriptions in the defaults (`nm_nc_freq`, `nm_snow_var`, the wind and stress variable names, `K_GM_min`). Those edits are byte-identical to the ones in #1004, so whichever PR merges second sees no conflict there.
2. **icepack:** `namelist.icepack` is itself the default and only its array sizes were described. The rest comes from the Icepack documentation and from the checks in `icedrv_set.F90` that tie options together. Its nine variants are rebuilt from it.
3. **REcoM:** there was no REcoM default, so `namelist.recom.2p3z2d` becomes the reference and the four `bin_*` files are rebuilt from it. Units are corrected where the code disagrees with the old comments: the ballasting sinking speeds, depth scaling and maximum sinking velocity are in m/day, since `recom_sinking.F90` divides by `SecondsPerDay` afterwards.

## For the REcoM people

27 parameters are read from the namelist but never used by the code, and are marked "not used by the current code": the 19 `s*Nsq` grazing terms, `cPhyN`, `cHetN`, `cZoo2N`, `P_cm_c`, `UseFeDust`, `UseDustClim`, `UseDustClimAlbani` and `HetRespFlux_plus`. `REcoMDataPath` carries the same marker. Please check these, since they may be meant to do something.

Separately, and not changed here, the `bgc_num` shipped in `bin_2p3z2d`, `bin_3p3z2d` and `bin_4p3z2d` does not match the value the startup check expects from `enable_3zoo2det` and `enable_coccos`, so those setups print a warning.

## Verification

- All 65 changed namelist files compare equal to `main` for all 3998 settings, both with f90nml and with a separate comment-stripping comparison.
- No inline comment was added on a row of a continued value list, since gfortran aborts on that.
- The `sed` patterns the REcoM CI job applies to `bin_2p3z2d` still match.
- The checks are static; the CI runs here are the first time the model reads the rebuilt files.

Not touched: `test/input/recom/*.working` (test data) and `fpost2/namelist.interp` (a post-processing tool).

### DIFF
--- a/config/bin_2p1z1d/namelist.config
+++ b/config/bin_2p1z1d/namelist.config
@@ -1,76 +1,133 @@
-! This is the namelist file for model general configuration
+! ============================================================================
+! ============ Namelist file for FESOM2 general configuration ================
+! ============================================================================
+! This file contains the main configuration parameters for FESOM2, including:
+! - Model identification and run settings
+! - Time stepping and simulation duration
+! - Initial time/date settings
+! - File paths for mesh, forcing, and output
+! - Restart and logging configuration
+! - Vertical coordinate system (ALE)
+! - Grid geometry and rotation
+! - Calendar settings
+! - Model components (ice, cavities, etc.)
+! - Parallel decomposition
+! - Iceberg settings
+! ============================================================================
 
+! ============================================================================
+! RUN IDENTIFICATION
+! ============================================================================
 &modelname
-runid='fesom'
+runid = 'fesom'        ! run identifier (used in output filenames)
 /
 
+! ============================================================================
+! TIME STEPPING AND RUN LENGTH
+! ============================================================================
 &timestep
-step_per_day=32 !96 !96 !72 !72 !45 !72 !96
-run_length= 1 !62 !62 !62 !28
-run_length_unit='y' ! y, m, d, s
+step_per_day      = 32         ! number of time steps per day (determines dt = 86400/step_per_day seconds)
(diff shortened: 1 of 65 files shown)


JanStreffing:
I think this should not go into 2.8 anymore. Should be checked. I target 2.8.1 for now.

PR #1059, opened 2026-09-11
Make CORE3 with ice-shelf cavities the default mesh

### LINKED ISSUE #703: CORE3 mesh development
-     We want a mesh with cavities and turn them on by default
-     Certain marginal seas should be added (Black sea, Persian Gulf)
-     Some passages should be revised (mainly in Indonesia)

Paper describing CORE2 mesh: https://gmd.copernicus.org/articles/7/663/2014/gmd-7-663-2014.pdf

JanStreffing:
Makes CORE3 with Antarctic ice-shelf cavities the default mesh, as discussed in #703. Open questions that this PR does not answer are collected in #703.

## Default namelist

`config/namelist.config` changes exactly three settings:

| setting | before | after |
|---|---|---|
| `MeshPath` | `.../MESHES_FESOM2.1/core2/` | `.../MESHES_FESOM2.1/core3/` |
| `use_cavity` | `.false.` | `.true.` |
| `step_per_day` | 36 (2400 s) | 48 (1800 s) |

1800 s is the step the CORE3 runs use after spin-up, so the default is meant for runs that branch off from a spun-up state. A cold start at 1800 s has not been shown to work, and FESOM has no timestep ramp; see #703.

`use_cavity` stays `.false.` as the code default in `gen_modules_config.F90`, since that only applies when a namelist omits the key.

## CI and setups

Every mkrun setup that relied on the namelist default now sets `use_cavity: False` explicitly, otherwise it would look for cavity files its mesh does not have: `core2`, `farc`, `pi`, `souf`, `test_core2`, `test_neverworld2`, `test_pi`, `test_pi_floatice`, `test_pi_icepack`, `test_pi_linfs`, `test_pi_partial`, `test_pi_visc7`, `test_pi_zstar`, `test_souf`. Setups that relied on the default timestep now also set `step_per_day: 36`, the old default, so their `fcheck` reference values stay valid: `test_neverworld2`, `test_pi`, `test_pi_cavity`, `test_pi_floatice`, `test_pi_icebergs`, `test_pi_icepack`, `test_pi_linfs`, `test_pi_partial`, `test_pi_visc7`, `test_pi_zstar`, `test_souf`. Each file differs from `main` only by these two keys. The CTest harness, `mesh_partition.cmake`, the `bin_*` REcoM configs and the toy configs already set `use_cavity` explicitly and need no change.

New:

- `setups/core3/setup.yml` for mkrun, cold-starting at 300 s as in the CORE3 spin-up runs, and a `core3` entry for Levante in `setups/paths.yml`.
- A `core3` entry in `tests/mesh_registry.json`. Its `config` block comes before `mesh_stats` on purpose: the harness reads the registry with `"<mesh>"[^}]*"<key>"` patterns, which cannot cross a nested object, so `use_cavity: true` would otherwise not be found and fall back to `.false.`.
- A CORE3 remote pipeline in `tests/remote/`: download, partition for 16 ranks, mesh diagnostics. Remote tests are skipped in CI.

The docs state the new defaults and that meshes without cavity files need `use_cavity=.false.`.

## Before merging

- ~~`core3.tar.gz` has to be uploaded next to `core2.tar.gz` in the `FESOM_MESHES2.1` container.~~ Done, see the comment below.
- The pool path is assumed to be `/pool/data/AWICM/FESOM2/MESHES_FESOM2.1/core3/`. `setups/paths.yml` has no albedo section, so only Levante is added.

## Verification

- `namelist.config` compared setting by setting: only the three settings above differ.
- All setup files parse, and apart from the new `core3` setup they differ from `main` only by the added `use_cavity` and `step_per_day` keys.
- CMake configures and lists the three CORE3 remote tests. The registry patterns resolve `use_cavity=true` and `step_per_day=48` for `core3`, and `false` for `dars` and `pi_remote`. The existing `core2` entry has its `mesh_stats` first, so none of its `config` values are reachable either; it falls back to the harness defaults, which happen to be right for CORE2.

### DIFF
--- a/cmake/FesomTesting.cmake
+++ b/cmake/FesomTesting.cmake
@@ -769,7 +769,7 @@ endfunction()
 
 # Function to add a FESOM meshdiag test with custom options
 function(add_fesom_meshdiag_test_with_options TEST_NAME MESH_NAME RUNID)
-    set(oneValueArgs NP TIMEOUT PREFIX)
+    set(oneValueArgs NP TIMEOUT PREFIX USE_CAVITY)
     set(multiValueArgs COMMAND_ARGS)
     cmake_parse_arguments(FESOM_TEST "" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
     
@@ -785,6 +785,22 @@ function(add_fesom_meshdiag_test_with_options TEST_NAME MESH_NAME RUNID)
     if(NOT DEFINED FESOM_TEST_TIMEOUT)
         set(FESOM_TEST_TIMEOUT 120)  # 2 minutes default (much faster than full FESOM)
     endif()
+    # A mesh with cavity files must be read with use_cavity=.true., otherwise
+    # the model stops. Take it from the mesh registry unless given explicitly.
+    if(NOT DEFINED FESOM_TEST_USE_CAVITY)
+        set(FESOM_TEST_USE_CAVITY ".false.")
+        set(_registry "${FESOM_TESTING_ROOT}/tests/mesh_registry.json")
+        if(EXISTS "${_registry}")
+            file(READ "${_registry}" _registry_content)
+            string(REGEX MATCH "\"${MESH_NAME}\"[^}]*\"use_cavity\"[ ]*:[ ]*([^,}]+)" _cavity_match "${_registry_content}")
+            if(_cavity_match)
+                string(STRIP "${CMAKE_MATCH_1}" _cavity)
+                if(_cavity STREQUAL "true")
+                    set(FESOM_TEST_USE_CAVITY ".true.")
+                endif()
+            endif()
+        endif()
+    endif()
     
     # Create test run directory
     set(TEST_RUN_DIR "${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}")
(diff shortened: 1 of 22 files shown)


JanStreffing:
`core3.tar.gz` is uploaded to `FESOM_MESHES2.1`, next to `core2.tar.gz`, so the registry URL in this PR resolves:

https://swift.dkrz.de/v1/dkrz_035d8f6ff058403bb42f8302e6badfbc/FESOM_MESHES2.1/core3.tar.gz

Checked from outside DKRZ: the download is public, the archive unpacks to a single `core3/` directory with 220509 nodes and 426403 elements, and it contains `nod2d.out`, `elem2d.out`, `aux3d.out`, `cavity_depth@node.out`, the edge and level files and `cavity_elvls.out`/`cavity_nlvls.out`, without partitions. It is 176 MB, against 6.8 MB for CORE2, because it also carries `mesh.nc`, `fesom.mesh.diag.nc` and a Python mesh pickle.

JanStreffing:
Ran the new CORE3 remote pipeline locally against the uploaded archive. The first run showed that the meshdiag test helper passed `use_cavity=.false.` for every mesh, so the CORE3 meshdiag test stopped at the cavity-files check. b990468d takes `use_cavity` from the registry, as `mesh_partition.cmake` already does.

With that:

| test | result |
|---|---|
| `remote_download_mesh_core3` | passed, 84 s for the 176 MB archive |
| `remote_partition_mesh_core3_16` | passed, partitioned with `use_cavity=.true.` |
| `remote_generate_meshdiag_core3_mpi16` | passed |
| `remote_*_core2`, `integration_meshdiag_pi_mpi2/8` | passed, still with `use_cavity=.false.` |

Remote tests are not part of PR CI, so this was the only way to exercise them.

